PR #10, opened 2014-11-06
Added ActiveRecord middleware

elsurudo:
To clear active connections after worker is processed. Courtesy of Mike Perham :)


waynerobinson:
Does this mean we can just use ActiveRecord as normal in Shoryuken, or do we need to use `#with_connection` all over the place?

elsurudo:
Yeah, you can just use it as normal.

Issue #13, opened 2014-11-07
cross-pollination


leemhenson:
Hey there. 

We've built a similar gem https://github.com/musicglue/envoy/tree/simplify to meet some needs in our organisation. 

Differences:
- It pushes to sns, not sqs. This is to support fanout pub-sub.
- It has rake tasks to build the sns and sqs infrastructure for you, according to config you define in your initializer: 

``` ruby
  config.add_dead_letter_queue 'my_dlq'

  config.add_queue('order_processing') do |queue, subscriptions|
    subscriptions.add 'order_placed', 'OrderWorker'
    subscriptions.add 'order_refunded', 'RefundWorker'
  end
```
- It generates IAM policies according to the config you have defined.
- It allows you to host custom celluloid actors in-process so you can have long-running jobs:

``` ruby
  config.celluloid.actors += [DailySalesAggregationWorker]
```
- It uses the latest aws-sdk-core-ruby gem.
- Middlewares for Airbrake, transactional workers + idempotent workers (persisting an application-defined message id in the database when it's processed on a queue and ignoring reprocessing of that message on that queue again).

Drawbacks:
- It doesn't do any kind of load-balancing or queue prioritisation.
- Signal handling is a little bit flakey.
- Requires raw messages in sqs.
- Only me hacking on it. :bus: 
- Very patchy test coverage.

Would you be interested in taking any of these ideas in shoryuken?

PR #24, opened 2014-12-10
Use heartbeat timer to keep message invisible in slow workers 

leemhenson:
(this PR depends on #18 and includes all of it's commits)

As I mentioned in #13 there is general need for queue consumers to handle the issue of long running workers exceeding the message visibility timeout on the queue they are consuming messages from. This PR uses a simple timer to extend the timeout every 5 seconds by default.

I've used Celluloid 0.16 because it allows the user of Timers 4.0.1, rather than Timers 1.1 which is used by Celluloid 0.15.2. There is a memory leak fixed in Timers 4 which we'd want. See: https://github.com/celluloid/timers/blob/master/CHANGES.md

I haven't yet put this code through it's paces, but I thought I'd at least get something down 'on paper' for discussion. The specs pass, at least. :meat_on_bone:

### DIFF
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -5,19 +5,30 @@ module Shoryuken
     include Celluloid
     include Util
 
-    def initialize(manager)
+    DEFAULT_HEARTBEAT = 5
+
+    def initialize(manager, heartbeat = DEFAULT_HEARTBEAT)
       @manager = manager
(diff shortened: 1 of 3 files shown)


elsurudo:
Looks good to me too. This is great, as it solves the edge condition with one of my jobs that doesn't have an "upper bound" visibility timeout I can reliably guess :+1:

PR #26, opened 2014-12-30
Worker registry

leemhenson:
(this PR depends on #24 and includes all of it's commits)

In order to make Shoryuken support custom message routing better I've pushed some code into a new class which I called the WorkerRegistry, which takes over from the WorkerLoader. WorkerRegistry defines an interface that subclasses must implement, and there is a DefaultWorkerRegistry which provides the normal Shoryuken behaviour. 

Publishing to sns topics is supported too.


leemhenson:
I've removed all the aws-sdk-core stuff from this work, so this should be easier to review and pull in now. Once it's in, I'll pull request in the aws-sdk-core commit too.

phstc:
Super awesome! Added to master :beers:

Issue #4, opened 2014-10-08
activejob support


juggy:
Is that in the plans?

elsurudo:
This is working for me in dev, in case someone needs it before it's merged: https://github.com/elsurudo/shoryuken/tree/active-job

phstc:
Added [ActiveJob Support](https://github.com/phstc/shoryuken#activejob-support) :smile:

renanmzmendes:
For the main ruby process to be able to save the job to the queue, Shoryuken must know about the main AWS configuration. It would be nice to have a generator that would create a loader of `shoryuken.yml` on the initializers folder.

Maybe refactor a `Shoryuken::CLI` method so it can also be used in the initializer?

Mine was something like this:

```ruby
def parse_config(config_file)
  if File.exist?(config_file)
    YAML.load(ERB.new(IO.read(config_file)).result)
  else
    raise ArgumentError, "Config file #{config_file} does not exist"
  end
end

config = parse_config([Dir.pwd, 'config/shoryuken.yml'].join('/')).deep_symbolize_keys
Shoryuken.options.merge!(config)

AWS.config Shoryuken.options[:aws]
```

Cheers

phstc:
you would to extract https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L46 and call it a in rails initializer?

renanmzmendes:
Actually, one would actually need to load `shoryuken.yml` so that `Shoryuken.options[:aws]` wasn't empty when `initialize_aws` was called.

P.S.: my previous post was based on the 0.0.5 gem version

phstc:
I agree!

Can you update to the master version? If so, maybe something like that would work:

```ruby
options = Shoryuken.load_config_file 'shoryuken.yml'
Shoryuken.initialize_aws(options)
```

The we would need to update [here](https://github.com/phstc/shoryuken/blob/36150d3ffc98c92bddb59e4bd0a7c8c0499699e1/lib/shoryuken/environment_loader.rb#L20) and [here](https://github.com/phstc/shoryuken/blob/36150d3ffc98c92bddb59e4bd0a7c8c0499699e1/lib/shoryuken/environment_loader.rb#L46) to use these new methods.

wdyt?

claudia-beas:
Hi,
so how exactly would I use Shoryuken with ActiveJob? Because I need to set  queuing backend, but I don't know what I'm supposed to write for Shoryuken.
[...]

PR #28, opened 2015-01-02
ActiveJob support

elsurudo:
This is working for me in development currently. This is similar to the code @phstc posted here: https://github.com/phstc/shoryuken/issues/4#issuecomment-62056681, with a few fixes and improvements.

Notes:
- Queue pausing/sleeping is disabled when ActiveJob is present. This is needed because the workers are dynamically-defined, so the queues would be paused indefinitely and jobs wouldn't be processed.
- Logging middleware altered to use the ActiveJob subclass name (instead of wrapper class name) for better logging.
- It would be good to add tests, but I'm not sure where to start with this yet. All current tests pass.
- Altered Celluloid dependency to track the minor version, in case of breaking changes.

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -173,7 +173,7 @@ module Shoryuken
       # get/remove the first queue in the list
       queue = @queues.shift
 
-      unless defined?(::ActiveJob) || Shoryuken.workers.include?(queue)
+      unless Shoryuken.workers.include? queue
         # when no worker registered pause the queue to avoid endless recursion
 
         logger.debug "Pausing '#{queue}' for #{Shoryuken.options[:delay].to_f} seconds, because no workers registered"
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -172,15 +172,16 @@ module Shoryuken
       # get/remove the first queue in the list
       queue = @queues.shift
 
+      unless defined?(::ActiveJob)
+        unless Shoryuken.workers.include? queue
+          # when no worker registered pause the queue to avoid endless recursion
 
-      unless Shoryuken.workers.include? queue
-        # when no worker registered pause the queue to avoid endless recursion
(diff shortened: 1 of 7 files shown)


lib/shoryuken/manager.rb
```diff
@@ -173,11 +173,10 @@ def next_queue
       # get/remove the first queue in the list
       queue = @queues.shift
 
-
-      unless Shoryuken.workers.include? queue
+      unless defined?(::ActiveJob) || Shoryuken.workers.include?(queue)
```
phstc: @elsurudo 

> Queue pausing/sleeping is disabled when ActiveJob is present. This is needed because the workers are dynamically-defined, so the queues would be paused indefinitely and jobs wouldn't be processed.

After we dynamically add a worker using `register_worker!`, `Shoryuken.workers.include? queue` will return `true`, then it should stop constantly pausing. The cons would be the huge number of debug message until the worker is registered.

Just wondering because if there's no worker defined yet, the processor will fail, because of no workers.

https://github.com/elsurudo/shoryuken/blob/active-job/lib/shoryuken/processor.rb#L13

wdyt?

shoryuken.gemspec
```diff
@@ -23,5 +23,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "pry-byebug"
 
   spec.add_dependency "aws-sdk-v1"
-  spec.add_dependency "celluloid", ">= 0.16.0"
+  spec.add_dependency "celluloid", "~> 0.16.0"
```
phstc: :heart:

lib/shoryuken/manager.rb
elsurudo: I had this in there originally because I found that only jobs on the default queue were being processed. However, I just tried getting rid of this now and all queues are processing, so it must have been unrelated.

Good catch! Will update.

phstc:
Super awesome again @elsurudo! Added to master :beers:

elsurudo:
Excellent!

PR #31, opened 2015-01-06
ActiveJob: skip pausing if workers aren't registered

elsurudo:
Turns out this is required after all, and I had it in there for a reason :)

So the issue is that ActiveJob workers get registered at runtime *only on the server process*. As far as the CLI processor is concerned, it doesn't know that there are workers registered to run the jobs, and thus the queues are paused and never resumed.

Because we resolve the worker classes at runtime, the CLI can still find and execute them.

Let me know if you can think of a better way to handle this, but it seems to be necessary to never pause the queues for ActiveJob purposes, as things are now.

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -173,7 +173,7 @@ module Shoryuken
       # get/remove the first queue in the list
       queue = @queues.shift
 
-      unless Shoryuken.workers.include? queue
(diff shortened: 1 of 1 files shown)


phstc:
Oh, got it. Nice find. Sorry for the trouble :(

Added to master! Thanks :dancer:

PR #32, opened 2015-01-06
fix activejob implementation

Maysora:
fixes undefined local variable body in `worker_name` method and delay seconds in float causing malformed input error from sqs


phstc:
Awesome! Added to master :beers:

PR #33, opened 2015-01-07
Initialize logger before loading Rails

elsurudo:
This is necessary in order to replace the standard Rails logger with the Shoryuken logger (need to initialize the logger before loading Rails, so that it is available to Rails initializers).

Why would we want replace the Rails logger? So that what is normally logged in Rails (via the usual `logger.info` and `Rails.logger.info` shows up in the Shoryuken logfile.


phstc:
Yeah! Thanks :beers:

PR #35, opened 2015-01-08
Shoryuken.on_start callback added

leemhenson:
I wanted to be able to execute some code only when shoryuken had booted celluloid, so I added a hook in:

```ruby
Shoryuken.on_start do
  # do some more stuff to celluloid, like setting up new SupervisionGroups
end
```

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -79,6 +79,12 @@ module Shoryuken
       @@default_worker_options = options
     end
 
+    def on_start(&block)
+      @start_callback = block
+    end
+
+    attr_reader :start_callback
+
     private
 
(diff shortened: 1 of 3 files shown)


phstc:
Hey @leemhenson just out of curiosity, do you have an example on what you would like to add in the startup? How the `SupervisionGroup` would work?

phstc:
Nice. Thanks for the explanation!

Do you have a use case specifically for Shoryuken? 

Just wondering if this change is too specific.

leemhenson:
Well, without it I don't see a clean way of running a block of code when:

- the current process is a Shoryuken process
- celluloid has been started

phstc:

PR #36, opened 2015-01-08
process_died should use ERROR log level, or not log if exit was clean

leemhenson:


phstc:
Awesome, added to master! :beers:

PR #37, opened 2015-01-08
message_attributes is nil on comcast/cmb

leemhenson:
I develop against a local instance of https://github.com/Comcast/cmb which doesn't currently support message attributes. It returns ```nil``` for ```Aws::SQS::Message#message_attributes``` which breaks ```worker_name``` which assumes that message_attributes will always be a hash. A very edge case I suppose, but it does no harm to fix it.


phstc:
:+1: it does no harm to fix it.

PR #39, opened 2015-01-12
use aws-sdk v2

leemhenson:


elsurudo:
This doesn't work with certain options. For example, I get the following error on launch with my config:

```
invalid configuration option `:receive_message'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/configuration.rb:160:in `rescue in block in apply_options'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/configuration.rb:156:in `block in apply_options'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/configuration.rb:155:in `each'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/configuration.rb:155:in `apply_options'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/configuration.rb:143:in `build!'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/base.rb:69:in `build_config'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/base.rb:20:in `initialize'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/aws-sdk-core/client_stubs.rb:19:in `initialize'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/base.rb:106:in `new'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/client.rb:20:in `sqs'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/client.rb:24:in `sqs_resource'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/client.rb:8:in `queues'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/cli.rb:260:in `block in validate!'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/cli.rb:257:in `each'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/cli.rb:257:in `validate!'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/lib/shoryuken/cli.rb:31:in `run'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/shoryuken-90ba8754ea49/bin/shoryuken:6:in `<top (required)>'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bin/shoryuken:23:in `load'
/Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bin/shoryuken:23:in `<main>'
```

If there are changes to options etc., they should be documented in the README.

Relevant section in my config:
```
receive_message: # See http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/SQS/Queue.html#receive_message-instance_method
    wait_time_seconds: <%= ENV['SHORYUKEN_WAIT_TIME'] || 0 %>,  # The number of seconds to wait for new messages when polling.
    attributes:
      - receive_count
      - sent_at
```

lib/shoryuken/cli.rb
```diff
@@ -269,7 +267,16 @@ def validate!
     def initialize_aws
       # aws-sdk tries to load the credentials from the ENV variables: AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
       # when not explicit supplied
-      AWS.config Shoryuken.options[:aws] if Shoryuken.options[:aws]
+      return unless Shoryuken.options[:aws]
+
+      non_standard_keys = %i(
+        account_id
+        sns_endpoint
+        sqs_endpoint)
```
phstc: @elsurudo could you try to add `receive_message ` as a `non_standard_keys`?

leemhenson:
done.

elsurudo:
Original error message done, here is what I get now:
```
13:39:37 worker.1 | 2015-01-26T12:39:37Z 66258 TID-ovb9vbtcw ERROR: Error fetching message: unexpected value at params[:attributes]
13:39:37 worker.1 | 2015-01-26T12:39:37Z 66258 TID-ovb9vbtcw ERROR: /Users/elsurudo/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.20/lib/seahorse/client/param_validator.rb:24:in `validate!'

```

leemhenson:
You'll need to modify your config to fit with the v2 gem's api:

http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Queue.html#receive_messages-instance_method

Try:

```
receive_message: # See http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/SQS/Queue.html#receive_message-instance_method
    wait_time_seconds: <%= ENV['SHORYUKEN_WAIT_TIME'] || 0 %>,  # The number of seconds to wait for new messages when polling.
    attribute_names:
      - ApproximateReceiveCount
      - SentTimestamp
```

elsurudo:
Thanks, that did it – the README should be updated to reflect the new configuration style.

Also, this seems to go a long towards solving https://github.com/phstc/shoryuken/issues/44, which I just opened today.

Basically, idle network usage has been reduced by 85%! I wonder if this is solely due to the new AWS library. Did you notice anything like this?

leemhenson:
Readme updated. 

@elsurudo I haven't been running the v1 version so I can't comment on it's performance vs v2.

elsurudo:
@leemhenson are you using ActiveJob, btw? If not, then @phstc please wait for me to test this together with Rails 4.2 + ActiveJob to ensure that nothing was missed.

leemhenson:
No I'm not.

phstc:
Thanks @elsurudo! Let me know, so I can merge this PR. 

I believe it will be safer to create a version 1.0 as it breaks the compatibility with the current version.

elsurudo:
@phstc will do, I hope to get it done later this week.

Rakefile
```diff
@@ -11,7 +11,7 @@ task :console do
   if File.exist? config_file
     config = YAML.load File.read(config_file)
 
-    AWS.config(config['aws'])
+    Aws.config = config['aws']
```
elsurudo: This isn't particularly to do with this PR, but it's convenient to bring it up here for discussion. What if I need to use some ruby code in my AWS config? For example, right now I need to use `logger: Logger.new($stdout)`, and I can't just put that in my YAML config file.

@phstc thoughts?

leemhenson: Well, changing the Shoryuken logger isn't aws config so you wouldn't put it in there anyway. You can call ```Shoryuken::Logging.logger = Logger.new(STDOUT)``` at any time to change the logger, but it should be logging to STDOUT anyway: see https://github.com/musicglue/shoryuken/blob/aws-sdk-core/lib/shoryuken/logging.rb#L28

I'm running Shoryuken with Rails, so I just dump any extra configuration code in ```config/initializers/shoryuken.rb```. That takes care of making sure it's executed for rake tasks, console sessions, etc. Is your situation different?

On a related note, there's a good argument for extracting all the config file reading stuff (e.g. https://github.com/musicglue/shoryuken/blob/aws-sdk-core/lib/shoryuken/cli.rb#L232) from ```Shoryuken::CLI``` and putting it somewhere else. Perhaps then it could be automatically loaded the first time you do anything with a Shoryuken object. You probably wouldn't need a ```:console``` task then. As it stands, this console rake task won't work if you have any non-Aws compatible keys in your yaml because it doesn't know to filter them out.

phstc: I'm not running Shoryuken with Rails, so maybe I'm missing something. But I agree that initializers seem to be the way to go. Does that work for you @elsurudo?

@leemhenson your note is because of https://github.com/phstc/shoryuken/blob/master/Rakefile#L14? So now it's not going to work without removing the standard keys: https://github.com/musicglue/shoryuken/blob/aws-sdk-core/lib/shoryuken/cli.rb#L275-L279? Do you guys use this task?

leemhenson: I don't use that task. I just ```rails console``` usually, but I think we definitely need a way to make Shoryuken work out of the box there without requiring the user to know a particular magic incantation to get it work.

phstc: @leemhenson what do you do with shoryuken using `rails console` just `Worker.perform_async`?

leemhenson: That, or ```Shoryuken::Client.sns.publish(...)```.

phstc: hm

So something like that:

```ruby
class Client
  def sqs
    @sqs ||= Shoryuken.aws::SQS.new
  end
end

module Shoryuken
  def self.aws
    unless @@aws_configured
      # initialize AWS
      # https://github.com/musicglue/shoryuken/blob/aws-sdk-core/lib/shoryuken/cli.rb#L275-L279
      @@aws_configured = true
    end

    AWS
  end
end
```

would work for you?

leemhenson: Yeah, I expect so.

elsurudo:
> Well, changing the Shoryuken logger isn't aws config so you wouldn't put it in there anyway. You can call Shoryuken::Logging.logger = Logger.new(STDOUT) at any time to change the logger, but it should be logging to STDOUT anyway: see https://github.com/musicglue/shoryuken/blob/aws-sdk-core/lib/shoryuken/logging.rb#L28

Actually @leemhenson, I was talking about changing the Aws logger for debugging purposes (not the Shoryuken logger, which I agree with you about). I'll see how this goes once I move my app over to this branch. Could you rebase/merge so I can try with the latest?

leemhenson:
Ah I see what you're saying. Um, I haven't done anything with the Aws logger so I can't comment. If you can't replace it with something like ```Aws.logger = logger``` then I guess we'd need to offer some sort of hook so that we could yield to block of the developer's choosing in the ```def self.aws``` that @phstc has pasted above. 

```ruby
Shoryuken.on_aws_config do |aws|
  aws[:logger] = my_logger
end
```

or something like that?

leemhenson:
So what needs to happen to get this one over the line?

Also, it's kind of annoying that AWS have removed the SQS resource classes from the recent aws-sdk gem versions. We have to either:

- pin to aws-sdk v2.0.21
- replace all usages of the resource classes with direct Aws::SQS::Client manipulation
- add our own versions of the resource classes in-house

I'm thinking option 3, because it keeps code changes to a minimum and we can use the code from the resource classes in v2.0.21.

phstc:
Hey @leemhenson 

I tested your PR and it's working, except for one test failing when testing with `SPEC_ALL=true rspec`.

I didn't accept it yet, because I believe the `aws-sdk-core` isn't stable enough as `aws-sdk` is. For example they released two versions in the same day and most of the versions are less than a week apart.

https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md#2022-2015-02-05

> Upgrading - Aws::SQS::Resources - Removed resource interfaces for Aws::SQS. These are unstable, so they have been removed for the public release of aws-sdk-resources.

So I don't see the advantage of using it yet. 

> add our own versions of the resource classes in-house

I agree with you about the option 3, I think it's the best. But I don't think we should implement that, as it seems to be a responsibility of the aws-sdk-core (or aws-sdk-resources). Part of my application tests are relying  one [AWS::ReceivedMessage](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/SQS/ReceivedMessage.html) and I believe for other people too. So we will need to update it for our specific version of message, then update it back to the Amazon Version when it's released. 

I believe if we release shoryuken with `aws-sdk-core` we will release an unstable version of the project, that's currently stable.

wdyt? 

Being super honest - you and @elsurudo are helping a lot with shoryuken. If both of you are on to merge it and help with the further maintenance because of the releases of `aws-sdk-core`. I will not be against of merging it.

leemhenson:
>
> I tested your PR and it's working, except for one test failing when
> testing with SPEC_ALL=true rspec.
>

Ah ok, I'll check that out tomorrow.


> I didn't accept it yet, because I believe the aws-sdk-core isn't stable
> enough as aws-sdk is. For example they released two versions in the same
> day and most of the versions are less than a week apart.
>
>
> https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md#2022-2015-02-05
>
> Upgrading - Aws::SQS::Resources - Removed resource interfaces for
> Aws::SQS. These are unstable, so they have been removed for the public
> release of aws-sdk-resources.
>
> So I don't see the advantage of using it yet.
>
I think it's having so many releases because it covers the whole of Aws's
offerings. Changes to the parts Shoryuken uses are pretty irregular.
Although the big one recently is that removal of the resource classes. I
know what you mean about not being much advantage to changing, but at some
point either:

- it's going to support features v1 doesn't
- shoryuken users are more likely to have v2 in their gemfiles than v1

I can't say when that tipping point will be.

>  add our own versions of the resource classes in-house
>
> I agree with you about the option 3, I think it's the best. But I don't
> think we should implement that, as it seems to be a responsibility of the
> aws-sdk-core (or aws-sdk-resources). Part of my application tests are
> relying one AWS::ReceivedMessage
> <http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/SQS/ReceivedMessage.html>
> and I believe for other people too. So we will need to update it for our
> specific version of message, then update it back to the Amazon Version when
> it's released.
>
The namespace in v2 differs in case Aws vs AWS, so potentially we could
ship an api-equivalent proxy under AWS::ReceivedMessage constant. Then your
code wouldn't need to change right away.


>  I believe if we release shoryuken with aws-sdk-core we will release an
> unstable version of the project, that's currently stable.
>
SQS itself isn't changing so we're not changing core functionality, and if
we fix the aforementioned issues I don't see stability being affected. What
do you think @elsurudo?

skippy:
hey folks,

this is a pretty major change, and I've been holding off on sending PRs until this one is merged in.  some of the tweaks I've looked at making change quite a bit between v1 and v2, such as helpers for batch sends.

But I'm patient :) just throwing my 2cents in.

leemhenson:
https://forums.aws.amazon.com/thread.jspa?threadID=172341&tstart=0

leemhenson:
pushed a couple of extra commits here so you can do:

* in rails-app/config/application.rb:

```ruby
console do
  Shoryuken::EnvironmentLoader.load_for_rails_console
end
```

to get all your config parsed and aws clients set up etc.

* in rails-app/config/initializers/shoryuken.rb:

```ruby
Shoryuken.on_aws_initialization do |aws_options|
  aws_options[:logger] = my_logger
end
```

@phstc those spec failures you mentioned are present on https://github.com/phstc/shoryuken/tree/master anyway (see https://github.com/phstc/shoryuken/pull/51). Looks like specs that are trying to talk to a real Aws endpoint?

phstc:
There are some "integration" tests hitting AWS SQS: https://github.com/phstc/shoryuken/blob/master/spec/spec_helper.rb#L9-17

I need to add a `shoryuken.yml` and AWS env variables in travis. 

I will create new creds for that :smile:

phstc:
> We can’t automatically merge this pull request.

@leemhenson Could you merge master into your branch? Then I will merge it into master :beers: 

> pin to aws-sdk v2.0.21

Maybe for now we could pint to v2.0.21 as you said until @trevorrowe releases a new version with the resources back (less code to maintain). wdyt? 

https://forums.aws.amazon.com/thread.jspa?threadID=172341&tstart=0

leemhenson:
Ok I've rebased onto master. I've split the branch into two:

- aws-sdk-core is pinned to aws-sdk 2.0.21
- aws-sdk-core-without-resources is not pinned and uses shoryuken-owned queue+message classes

I'll keep the second branch around in case I need a version of aws-sdk > 2.0.21 at some point. But hopefully it won't be long until aws-sdk gets it's sqs resources back.

phstc:
Awesome work @leemhenson! Thanks a lot :beers: 

Let's wait a little before releasing a new version, which needs to be v1.0.0, because of the breaking changes. wdyt?

PR #40, opened 2015-01-12
process_died should not log if no reason provided

leemhenson:
Looks like the previous commit where I made this change was accidental reverted?


phstc:
Have you ever received a reason different than `nil`? For me it's always `nil`. Maybe we could drop this line...  but on the other If we do that we will have no log saying that the processor died.

What do you think?

leemhenson:
I see it for any uncaught error in the worker.

phstc:
Added to master :smile:

Issue #41, opened 2015-01-12
hard shutdown is broken


leemhenson:
If you have a long-running worker processing a job and you SIGINT in the middle of it, you will receive an error from this line:

https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/manager.rb#L212

because there is no such method ```actual_work_thread```. The process is then zombied and has to be SIGTERMed. It looks like this was code taken from an old version of Sidekiq, which now does something different by keeping a hash of processor object ids to threads. See:

https://github.com/mperham/sidekiq/commit/06acbd4f60d9e10c4e4b3105c6f6beb85541693c

leemhenson:
When I try adding the Sidekiq-style thread tracking, the ```Manager``` never appears to handle the ```stop``` call. It just hangs, which is weird because I've verified it is a different thread entirely from the long-running processor thread. 

Weird!

phstc:
Closed by #42 :smile:

PR #42, opened 2015-01-13
modern hard shutdown sequence from Sidekiq

leemhenson:
For issue #41.

This is taking the Sidekiq approach. It seems to work ok for me.


phstc:
Nice, looks good!

So, basically `processor.bare_object.actual_work_thread` is not working?

leemhenson:
Yeah it looks like that was the old way of doing it in Sidekiq, which they then replaced with the thread tracking hash.

Issue #44, opened 2015-01-26
High Bandwidth Usage


elsurudo:
We've been experiencing high bandwidth usage in production/staging with shoryuken.

My settings are as follows:
- wait_time_seconds: 5
- concurrency: 10
- delay: 5

I have 7 queues, and according to the above settings (confirmed with verbose logging), the queues are being checked every 5 seconds.

In staging (an EC2 instance), I'm seeing upstream usage averaging about 70kb/s with shoryuken, with negligible downstream. I used a 5 min. average to calculate this. Note that during this period, no jobs were executed, as confirmed by the log.

In development (local machine), a 5 min average shows 8kb/s being used by shoryuken, which definitely isn't as bad as on staging, but still higher than I would expect, considering just SQS communication every 5 secs (and why upstream, as opposed to downstream?).

I tried upgrading to the newest version of "aws-sdk-v1", but it had no effect. I also narrowed down the usage to SQS, as the IPs that the data heads to reverse-map to SQS hosts.

Any ideas on what could be causing this strange behavior? 70kb/s average on EC2 is definitely not ideal!

PR #45, opened 2015-01-28
Bandwidth fix

elsurudo:
Fix for https://github.com/phstc/shoryuken/issues/44

What was happening was that `shoryuken_class` was being added to the `message_attribute_names` of `options` every time Fetcher's `receive_message` was called, leading to larger-and-larger request sizes.

Now, we simply `dup` the options every time to start with a clean slate.

BTW, I cherry picked a commit that already is merged into master here: https://github.com/phstc/shoryuken/pull/33/files (I am working from our "stable" based on 0.0.4, since we are using this in production). As this commit is already there, the only file of interest here is `fetcher.rb`.

I thought Github was smart enough to spot the similarities, but guess not.


phstc:

PR #47, opened 2015-01-29
Call `configure_server` only on server

elsurudo:
`Shryuken.configure_server` block should only be executed on server. The issue I was having is that the content of this block was also being executed by my Rails initializer, which meant if I did anything funny like redirecting Rails logger output to Shoryuken in there (this is good for the Shoryuken CLI), my Rails app would also do the same thing (not good).

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -57,7 +57,7 @@ module Shoryuken
     #   end
     # end
     def configure_server
-      yield self
+      yield self if server?
     end
 
     def server_middleware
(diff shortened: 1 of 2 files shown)


phstc:
How the `configure_server` was being called in your initializer?

elsurudo:
Closing for now, as I may have discovered an issue with this.

elsurudo:
Actually all is good. Someone on the team didn't run `bundle`, so they were still getting the old functionality.

elsurudo:
@phstc in the usual way: `Shoryuken.configure_server do |config|` Is there any other way? :)

The issue with that is the initializer is run by the web app, AND the CLI, whereas we only want the code inside `configure_server` to run on the worker.

With the changes here, we get the desired functionality (similar to Sidekiq).

phstc:
Fair enough! 

It does make sense. Added to master :beers:

PR #50, opened 2015-02-25
Specify what queue does not exist

Holek:
Before the message was ambiguous, and didn't specify it was AWS queue that wasn't there.


phstc:

PR #51, opened 2015-02-25
run all specs under travis

leemhenson:
shows that specs are broken on master


phstc:

PR #53, opened 2015-02-26
travis should target ruby 2.2.0 too

leemhenson:


leemhenson:
The travis failures are because it doesn't have access to the aws env vars. I expect you've set them up like this: 

http://docs.travis-ci.com/user/environment-variables/#Using-Settings 

right? If so, it will pass when merged.

phstc:
@leemhenson I set these variables yesterday 
screen shot 2015-02-26 at 11 07 20 am

And the shoryuken.yml should use them: https://github.com/phstc/shoryuken/blob/master/spec/shoryuken.yml
https://github.com/phstc/shoryuken/blob/master/spec/spec_helper.rb#L9

It works in development: `AWS_ACCESS_KEY_ID=... AWS_SECRET_ACCESS_KEY=... SPEC_ALL=true rspec` weird.

phstc:
The credentials seem to be working, but now there are other failures: https://travis-ci.org/phstc/shoryuken/jobs/52279068

Issue #52, opened 2015-02-26
auto_visibility_timeout uses queue in options instead of receiving queue


kookster:
In `Processor.auto_visibility_timeout`, there is a `queue` passed in as the first arg, but this is not used for the `visibility_timeout_heartbeat`:
```ruby
    def auto_visibility_timeout(queue, sqs_msg, worker_class)
      if worker_class.auto_visibility_timeout?
        timer = every(worker_class.visibility_timeout_heartbeat) do
```
Instead, when it gets the value from `Worker.extended_visibility_timeout`, it uses the `queue` from the shoryuken options to get the `visibility_timeout`:

```ruby
      def extended_visibility_timeout
        Shoryuken::Client.visibility_timeout(get_shoryuken_options['queue'])
      end
```

However, the actual receiving queue, passed in to `auto_visibility_timeout`, may well not match the queue from `get_shoryuken_options`.

phstc:
@kookster hm... in which cases the queue will not match?

kookster:
I have different priority tasks, but they all processed by the same task worker.

I drop message on one of 4 queues: p1, p2, p3, and p3, but they are all processed using the same TaskWorker.  Any given message being processed by that one TaskWorker may be any one of those 4 queues.

The correct current queue is passed in to the `auto_visibility_timeout` method, but then isn't used when determining the `visibility_timeout` in the heartbeat logic.

kookster:
I should mention I have written the code to use the queue passed in to `auto_visibility_timeout` to determine the timeout, and I'm happy to provide it as a PR.

phstc:
gotcha! 

Could you send a PR with the changes you made?  :heart:

kookster:
awesome, yes, working on it now - one quick question - the timeout right now is hard coded to always extend by 5 seconds.

Some of my jobs (transcoding and manipulating > hour long audio files) can take a long up to an hour, so always extending by 5 seconds at a time seems like it could result in many extension messages.

What do you think about making the extension a function of the original timeout?
e.g. if the original timeout is ~ 2 minutes, then 5 seconds makes sense, if it is 1 hour, then maybe extend by 1 minute at a time?

phstc:
The timeout isn't hardcoded to 5 seconds, the extended timeout is the queue default visibility timeout and the heartbeat verifier is 5 seconds less than that, just to give a enough time to update the message visibility before SQS returns the message to the queue.

[...]

PR #54, opened 2015-02-26
use the queue passed to process for auto_visibility_timeout

kookster:
Fixes issue #52 

When processing an sqs message, the processor should look up the visibility timeout for the queue where the message came from, rather than rely on the default queue of the worker it fetches to process the message.


kookster:
I'll rebase against your master.

kookster:
never mind - looks like this commit:

https://github.com/phstc/shoryuken/commit/74732c695c2cb3ca83863199d076f32f87e27bb0

from:  https://github.com/phstc/shoryuken/pull/39

takes care of it already

phstc:
Thanks @kookster 

:star: @leemhenson :beers:

PR #56, opened 2015-02-26
no commas in between yaml properties

kookster:
Not a big deal, but bothers me a bit when I see it.


phstc:

PR #58, opened 2015-02-27
fix all the specs that I broke in a somewhat cavalier fashion

leemhenson:
er, yeah. :blush:

### DIFF
--- a/lib/shoryuken/client.rb
+++ b/lib/shoryuken/client.rb
@@ -29,7 +29,7 @@ module Shoryuken
       end
 
       attr_accessor :account_id
-      attr_writer :sns, :sqs, :sns_arn
+      attr_writer :sns, :sqs, :sqs_resource, :sns_arn
 
       private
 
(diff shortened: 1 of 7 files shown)


phstc:
hehe tks :beers: 

let's see if it will pass merged into master :pray:

leemhenson:
I think these tests are pretty valuable to be honest. I think we should try to get them running. I suppose we could do something like setting up each test to avoid crosstalk by uniquely identify it's own messages, but that would make writing tests pretty ugly.

leemhenson:
Annoying that this has failed:

https://travis-ci.org/phstc/shoryuken/builds/52414881

Can you try sleeping longer here? https://github.com/phstc/shoryuken/blob/master/spec/integration/launcher_spec.rb#L105

kookster:
hey @leemhenson @phstc this is great.

There are also incorrect `find_message` calls in `Rakefile` and `active_job_adapter.rb`
Do you want me to put in a different PR for fixing those?

PR #59, opened 2015-02-27
only purge once per minute in travis

leemhenson:


leemhenson:
hmm. this might not work either because of the build matrix. we might need ruby-versioned queue names.

phstc:
hm let me try a quick test having only one ruby version on travis first.

PR #62, opened 2015-03-04
update aws dependencies

magegu:
if you want to use the latest gem 'aws-sdk', '~> 2' and shoryuken it returned 
```
Bundler could not find compatible versions for gem "aws-sdk-resources":
  In Gemfile:
    aws-sdk (~> 2) ruby depends on
      aws-sdk-resources (= 2.0.22) ruby

    shoryuken (>= 0) ruby depends on
      aws-sdk-resources (2.0.21.pre)
```
so i updated the dependencies to the required version.

### DIFF
--- a/shoryuken.gemspec
+++ b/shoryuken.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "nokogiri"
   spec.add_development_dependency "dotenv"
 
-  spec.add_dependency "aws-sdk-core", "2.0.21"
-  spec.add_dependency "aws-sdk-resources", "2.0.21.pre"
+  spec.add_dependency "aws-sdk-core", "2.0.22"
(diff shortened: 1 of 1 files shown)


magegu:
ah sorry, i just realised that there is more work involved in updateding the dependencies than just the version numbers. I'll have a look at the changes and see what i can do to upgrade shoryuken.

phstc:
hey @magegu we forced the aws-sdk version because [aws removed the resource classes in the newer versions](https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md#2022-2015-02-05), we discussed that here: https://github.com/phstc/shoryuken/pull/39#issuecomment-75812657

They said that it's a priority to add it back https://forums.aws.amazon.com/thread.jspa?threadID=172341&tstart=0

Did they add it again?

PR #71, opened 2015-03-09
Adding "rails installer", which generates the initializer in order to co...

renanmzmendes:
...nfigure rails main process environment

I have used Thor for that. Maybe we should considering using it also when running shoryuken. In my opinion it is cleaner than using OptionParser

### DIFF
--- a/lib/shoryuken/extensions/active_job_adapter.rb
+++ b/lib/shoryuken/extensions/active_job_adapter.rb
@@ -36,10 +36,11 @@ module ActiveJob
         private
 
         def message(job, options = {})
-          {
-            message_body: job.serialize,
-            message_attributes: message_attributes
-          }.merge(options)
+          body = job.serialize
+          body = JSON.dump(body) if body.is_a?(Hash)
+
+          { message_body: body,
+            message_attributes: message_attributes }.merge(options)
         end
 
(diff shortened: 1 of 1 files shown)


phstc:
Not sure if we need to add a dependency on Thor just for `shoryuken install` and it seems to be a Rails thing `config/initliazers/shoryuken.rb`, could it be a Rails generator or a simple Rake task?

Could you split this PR in two? One with the `initialize_aws` extraction and another for the `shoryuken start`? Maybe we can just add an entry on wiki or README to setup the shoryuken.yml and initializer instead of having a CLI command.

lib/shoryuken.rb
```diff
@@ -95,6 +95,10 @@ def on_stop(&block)
       @stop_callback = block
     end
 
+    def config_rails_main_process
+      EnvironmentLoader.load_for_rails_main_process
+    end
```
phstc: why expliciting calling it `rails`? For example, I'm not using Shoryuken with Rails, I ran it without Rails.

phstc:
PS: Could have a look at the indentation? https://github.com/renanmzmendes/shoryuken/blob/master/lib/shoryuken/cli/install.rb

Issue #65, opened 2015-03-06
Retrying jobs


joekhoobyar:
I would like to work on this feature, do you have any feedback about how you would like it to work?

I figure that we would be using the `:approximate_receive_count` message attribute to drive the retrying of jobs, and then set the message's visibility timeout based on the number of retries.  Does this make sense?

Configuration for the retry could be in the shoryuken_options for the worker.  Maybe a Hash of Fixnum and/or Range keys that map to visibility timeout values?  Middleware would have to scan the Hash to find the first matching key and the assign the visibility timeout.  If the retry count is greater than all of the keys in the Hash (Range end or Fixnum), then delete the message?

Just food for thought, really I am just making a bunch of assumptions about how you would like it to work. :-)  I am open to anything.

joekhoobyar:
FYI - I got the idea from your comment here:

https://github.com/phstc/shoryuken/blob/v0.0.4/lib/shoryuken.rb#L92

phstc:
SQS has auto retry built in, and also dead letter queues :smile: 

That comment was to add exponential backoff, but no one asked for yet. 

Have a look at https://github.com/phstc/shoryuken/wiki/Retrying-a-message and [Auto retrying errors](https://github.com/phstc/shoryuken/wiki/Retrying-a-message).

joekhoobyar:
I do like the exponential backoff idea, if you look at my idea above it could easily be used to implement that via configuration.

phstc:
I like your idea, would it be something like that:

```ruby
class MyWorker
  include Shoryuken::Worker
  
  shoryuken_options retry_intervals: [10.minutes, 30.minutes, 1.hour, 6.hours, 12.hours]
end
```

The `retry_intervals` could be optional, having a reasonable default value.

I'm doing exponential backoff in my project, but I didn't extract it yet, would you like to send a PR with it as a middleware?

a snippet of what I'm doing in my project:

```ruby
def retry_intervals
  if development?
    [7.seconds, 9.seconds, 12.seconds]
  else
    [10.minutes, 30.minutes, 1.hour, 6.hours, 12.hours]
  end
end

def handler_failure(sqs_msg, e)
  attempts = sqs_msg.receive_count - 1

  interval = if attempts < retry_intervals.size
     retry_intervals[attempts]
   else
     retry_intervals.last
  end

  # timeout cannot be larger than 43200 (12.hours)
  if interval + sqs_msg.queue.visibility_timeout > 43200
[...]

PR #73, opened 2015-03-11
Worker retry intervals

joekhoobyar:
Here's a new pull request.  I have renamed AutoRetry to ExponentialBackoffRetry.

### DIFF
--- a/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
+++ b/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
@@ -5,22 +5,30 @@ module Shoryuken
         include Util
 
         def call(worker, queue, sqs_msg, body)
+          started_at = Time.now
           yield
         rescue Exception => e
           retry_intervals = Array(worker.class.get_shoryuken_options['retry_intervals'])
-          if retry_intervals.empty?
+           
+          if retry_intervals.empty? || ! handle_failure(sqs_msg, started_at, retry_intervals)
             # Re-raise the exception if the job is not going to be retried.
             # This allows custom middleware (like exception notifiers) to be aware of the unhandled failure.
             raise e
-          else
-            handle_failure sqs_msg, retry_intervals
           end
         end
         
       private
       
-        def handle_failure(sqs_msg, retry_intervals)
-          attempts = sqs_msg.receive_count - 1
+        def handle_failure(sqs_msg, started_at, retry_intervals)
+          attempts = sqs_msg.attributes['ApproximateReceiveCount']
+            
+          # We need these two attributes in order to calculate the next visibility timeout.
+          unless attempts
+            logger.debug { "Message #{sqs_msg.id} will not be retried, because ApproximateReceiveCount missing" }
+            return
+          end
+          
+          attempts = attempts.to_i - 1
         
           interval = if attempts < retry_intervals.size
               retry_intervals[attempts]
@@ -28,12 +36,15 @@ module Shoryuken
               retry_intervals.last
             end
         
-          # Visibility timeouts are limited to 12 hours and are not additive.
+          # Visibility timeouts are limited to a total 12 hours, starting from the receipt of the message.
+          # We calculate the maximum timeout by subtracting the amount of time since the receipt of the message.
+          #
           # From the docs:  "Amazon SQS restarts the timeout period using the new value."
           # http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/AboutVT.html#AboutVT-extending-message-visibility-timeout
-          interval = 43200 if interval > 43200
+          max_timeout = 43200 - (Time.now - started_at).ceil - 1
+          interval = max_timeout if interval > max_timeout
         
-          sqs_msg.visibility_timeout = interval
+          sqs_msg.change_visibility(visibility_timeout: interval.to_i)
         
           logger.info "Message #{sqs_msg.id} failed, will be retried in #{interval} seconds."
         end
--- a/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
+++ b/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
@@ -1,7 +1,7 @@
 module Shoryuken
   module Middleware
     module Server
-      class AutoRetry
+      class ExponentialBackoffRetry
         include Util
 
         def call(worker, queue, sqs_msg, body)
--- a/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
+++ b/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
@@ -24,7 +24,7 @@ module Shoryuken
             
           # We need these two attributes in order to calculate the next visibility timeout.
           unless attempts
-            logger.debug { "Message #{sqs_msg.id} will not be retried, because ApproximateReceiveCount missing" }
+            logger.debug { "Message #{sqs_msg.message_id} will not be retried, because ApproximateReceiveCount missing" }
             return
           end
           
@@ -46,7 +46,7 @@ module Shoryuken
         
           sqs_msg.change_visibility(visibility_timeout: interval.to_i)
         
-          logger.info "Message #{sqs_msg.id} failed, will be retried in #{interval} seconds."
+          logger.info "Message #{sqs_msg.message_id} failed, will be retried in #{interval} seconds."
         end
       end
     end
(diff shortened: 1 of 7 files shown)


joekhoobyar:
This references #65

phstc:
Hi @joekhoobyar 

Great work!

Did you have a chance to test it for real? `visibility_timeout` in the sdk is [change_visibility](http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html) and I couldn't find a new one for `receive_count`.

The maximum visibility timeout must be less than 43200 `12.hours`, i.e. 43199. And it's incremental, if the first time it was delayed for `1.hour`, the next delay should not exceed 11.hours otherwise it will raise an exception.  But I don't think we can do much with that, I believe people must be careful while configuring the `retry_intervals`.

```ruby
sqs_msg.change_visibility visibility_timeout: 43200
Aws::SQS::Errors::InvalidParameterValue: Value 43200 for parameter VisibilityTimeout is invalid. Reason: Total VisibilityTimeout for the message is beyond the limit [43200 seconds].
from /Users/pablo/.gem/ruby/2.0.0/gems/aws-sdk-core-2.0.21/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
```

joekhoobyar:
@phstc - wow, I didn't take the new SDK into consideration.  Let me make these changes.  Also, I see what you mean about the visibility timeout.  I think there is a timestamp for the first receipt date of a message.  We could use that, possibly.

joekhoobyar:
Alright, I have been testing this at the console and it looks like the visibility timeout starts from the _most recent_ message receipt.  So, I guess we don't need the first receipt date after all.

joekhoobyar:
@phstc - This is finally ready!  I have tested it for real, and gotten everything working properly. :+1:

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
```diff
@@ -0,0 +1,54 @@
+module Shoryuken
+  module Middleware
+    module Server
+      class ExponentialBackoffRetry
+        include Util
+
+        def call(worker, queue, sqs_msg, body)
+          started_at = Time.now
```
phstc: @joekhoobyar shouldn't the `started_at` be based on `CreatedTimestamp `?

http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html#receive_message-instance_method

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
```diff
@@ -0,0 +1,54 @@
...
+            # Re-raise the exception if the job is not going to be retried.
+            # This allows custom middleware (like exception notifiers) to be aware of the unhandled failure.
+            raise e
+          end
+        end
+        
+      private
+      
+        def handle_failure(sqs_msg, started_at, retry_intervals)
+          attempts = sqs_msg.attributes['ApproximateReceiveCount']
```
phstc: `sqs_msg.attributes` is returning `nil` for me, we probably need to 

https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/fetcher.rb#L18

```ruby
options[:attribute_names] = %w[ApproximateFirstReceiveTimestamp ApproximateReceiveCount]

# OR %w[All]
```

http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html#receive_message-instance_method

> :attribute_names => Array<String>
> A list of attributes that need to be returned along with each message.
> 
> The following lists the names and descriptions of the attributes that can be returned:
> 
> All - returns all values.
> ApproximateFirstReceiveTimestamp - returns the time when the message was first received from the queue (epoch time in milliseconds).
> ApproximateReceiveCount - returns the number of times a message has been received from the queue but not deleted.
> SenderId - returns the AWS account number (or the IP address, if anonymous access is allowed) of the sender.
> SentTimestamp - returns the time when the message was sent to the queue (epoch time in milliseconds).


```ruby
resp = sqs.receive_message(
  # required
  queue_url: "String",
  attribute_names: ["Policy|VisibilityTimeout|MaximumMessageSize|MessageRetentionPeriod|ApproximateNumberOfMessages|ApproximateNumberOfMessagesNotVisible|CreatedTimestamp|LastModifiedTimestamp|QueueArn|ApproximateNumberOfMessagesDelayed|DelaySeconds|ReceiveMessageWaitTimeSeconds|RedrivePolicy", '...'],
  message_attribute_names: ["MessageAttributeName", '...'],
  max_number_of_messages: 1,
  visibility_timeout: 1,
  wait_time_seconds: 1,
)
```

wdyt?

phstc: after adding `options[:attribute_names] = %[All]`:

```ruby
2.0.0 (#<StandardWorker:0x007fa7937ea5a8>)> sqs_msg.attributes['ApproximateReceiveCount']
=> "1"
2.0.0 (#<StandardWorker:0x007fa7937ea5a8>)> sqs_msg.attributes
=> {"SenderId"=>"AIDAJL6PC3ADCVDF74OWO", "ApproximateFirstReceiveTimestamp"=>"1426176742828", "ApproximateReceiveCount"=>"1", "SentTimestamp"=>"1426176742373"}
```

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
joekhoobyar: When I tested it at the console, it seemed that the visibility timeout actually starts at each receipt of the message.  I was able to set a visibility timeout of 43180 on a message that was first received 10 minutes ago, because the most recent receipt was 10 seconds ago.

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
joekhoobyar: How about `options[:attribute_names] ||= %[All]` so that users can still customize?  The middleware detects if the attribute is missing, so it should not cause a crash.

joekhoobyar:
An example of the visibility timeout working on messages first received an hour ago, but most recently received 60 seconds ago (or less):

```ruby
# Time.now is  2015-03-12 12:40:59 -0400
time1 = Time.at(msg1.attributes['ApproximateFirstReceiveTimestamp'].to_i / 1000)  
# time1 is 2015-03-12 11:46:55 -0400

# Message was LAST received around 60 seconds ago, so the following succeeds:
msg1.change_visibility(visibility_timeout: 43100)
```

joekhoobyar:
@phstc - See my latest comment (with example code).  It looks like we don't need `ApproximateFirstReceiveTimestamp`

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
phstc: Awesome! I didn't know that. Tks :beers:

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
phstc: hm actually I believe we don't need `||=` with `%[All]`,  because `All` will cause to return everything.

joekhoobyar: I'll switch it to `= %[All]`.  Just didn't want to overwrite the `:receive_message` key coming from `shoryuken.yml` underneath the `:aws` key.

joekhoobyar:
@phstc - I've made that change to `= %[All]`.  Think this is ready to be merged in.  wdyt?

phstc:
Great work, thanks! Added to master :beers:

steveklebanoff:
Nice feature!  I thought we may have to build this ourselves, but here it is.  :+1: 

Would be nice to see some documentation about this in the README or wiki.  I had to discover this feature through this PR.

phstc:
> Nice feature!

Great work from @joekhoobyar :beers: 

> Would be nice to see some documentation about this in the README or wiki

@steveklebanoff good idea. Just added an entry, could you review it? :smile: 

https://github.com/phstc/shoryuken/wiki/Worker-options#retry_intervals

steveklebanoff:
Entry in the wiki looks good, thank you for updating! :+1: 

Only other thought is that it would likely be good to reference this on the [Retrying a message](https://github.com/phstc/shoryuken/wiki/Retrying-a-message) page as well. Thanks for your hard work @joekhoobyar and @phstc

phstc:
@steveklebanoff good idea again! :beers: 

Done https://github.com/phstc/shoryuken/wiki/Retrying-a-message#exponential-backoff

PR #76, opened 2015-03-14
Don't fail on unsupported signals, e.g. on Windows.

janpapenbrock:
Shameless imitation of how its done in Sidekiq https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L55-L63

Basically it won't fail if a signal is not supported, but only report the issue.


phstc:
Awesome! Added to master :beers:

PR #79, opened 2015-03-17
ignore Eclipse project files

joekhoobyar:
Not a big deal, but it would be nice if these were ignored by default. :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ shoryuken.yml
 *.log
 .env
 rubocop.html
+.buildpath
+.settings
+.project


phstc:
Not sure if that should be in shoryuken or in your global gitignore:

https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore

I usually try to keep the .gitignore editor agnostic. 

wdyt?

phstc:
except for a few lines (project specific) the current gitignore is the same as https://github.com/github/gitignore/blob/master/Ruby.gitignore

joekhoobyar:
that's fine, no worries. :-)

Issue #83, opened 2015-03-27
Update AWS SDK version


sschepens:
I know that amazon has removed some implementations that were crucial to shoryuken.
But, is there no way of updating the amazon sdk version? right now i have request that's failing and works with latest sdk version.
And it's just a pain to use v1.
Amazon sdk's are open source, can't you just copy the code that's missing until amazon adds support?

Thanks,
Sebastian

phstc:
Hi @sschepens 

Shoryuken uses the aws-sdk v2: https://github.com/phstc/shoryuken/blob/master/shoryuken.gemspec#L27

sschepens:
@phstc I think i havent made myself clear, I know Shoryuken uses v2 but its locked to version 2.0.21 and.
The problem is that for example Autoscaling::Client's method #enter_standby is not working on 2.0.21 but it's working on latest 2.0.33.
Is there no chance of updating to latest version of v2?

PR #84, opened 2015-03-28
Add Queue and Message, Update aws-sdk

sschepens:
#83 
Added simple Queue and Message implementations which provide all the functionallity needed by Shoryuken itself, maybe needs some more methods, but i cant find documentation for old versions of v2.
This implementations allows for unlocking aws-sdk versions and use latest.

All tests pass and tested locally with a queue and worked just fine.

Cheers,
Sebastian


phstc:
Thanks @sschepens!  Let me check with @leemhenson, because he worked on something message/queue before.

Hey @leemhenson, is that similar to what you have for message & queue? Will these changes work for you as well?

leemhenson:
What's changed since https://github.com/phstc/shoryuken/pull/39 ? The consensus then was that we would pin to 2.0.21 of aws-sdk rather than reimplementing the missing resources.

phstc:
> The consensus then was that we would pin to 2.0.21 of aws-sdk

I agree with that. But it seems that aws-sdk is taking long to add these resources back, even being a [priority](https://forums.aws.amazon.com/thread.jspa?threadID=172341&tstart=0).

I'm not 100% sure on having these implemented in Shoryuken, but aws-sdk releases a version every week https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md (usually more than one per week)

What are your thoughts on that? Let's keep those like our previous consensus or add them? Will that be useful for you?

leemhenson:
I'm happy either way as long as it's not a API breaking change.

phstc:
Added to master :beers:

sschepens:
Thanks for merging this guys! as soon as aws-sdk adds it's own implementations we can revert this :smile:

PR #87, opened 2015-04-07
Support IAM role based access while providing non-credentials related AWS config

chrislwade:
The proposed change allows the configuration of non-credentials related AWS config options without forcing users to provide credentials, thereby enabling role-based access.

**Background**

The AWS SDK supports `Aws::InstanceProfileCredentials` if no other credentials are provided and the SDK is being used from an EC2 instance.  Unfortunately, providing `Aws.config[:credentials]` with an instance of `Aws::Credentials` with `nil` or empty credentials disables this functionality.

Admittedly, I have yet to find exactly where in the SDK this is taking place as `Aws:: CredentialProviderChain` does not appear to reference `Aws.config[:credentials]`.  The request signers also appear to use the result of `Aws::CredentialProviderChain#resolve` instead of referencing `Aws.config[:credentials]`.  I am probably just overlooking something obvious at this point though.

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -58,11 +58,13 @@ module Shoryuken
         shoryuken_keys.include?(k)
       end
 
+      # assume credentials based authentication
       credentials = Aws::Credentials.new(
         aws_options.delete(:access_key_id),
         aws_options.delete(:secret_access_key))
 
-      aws_options = aws_options.merge(credentials: credentials)
+      # but only if the configuration options have valid values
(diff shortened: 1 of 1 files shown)


chrislwade:
By the way, I wasn't really sure how to integrate specs for this into the specs currently in the project.  If you have suggestions, I'll happily write the specs to ensure this works in as many scenarios as possible.

phstc:
Awesome! I didn't know `credentials.set?`.

https://docs.aws.amazon.com/sdkforruby/api/Aws/Credentials.html#set%3F-instance_method

Added to master :beers:

PR #91, opened 2015-04-13
body_parser should call load method on class

sschepens:
Parse method is not standardized in ruby gems, but load is, for example multi_json and oj dont implement parse but do implement load.
Json gem implements load as well of course.


phstc:

I kept `parse` to avoid breaking changes: https://github.com/phstc/shoryuken/commit/7f97f9d803d507ce1e1af1b86af47daa60576bad

PR #96, opened 2015-04-24
reduce noise in logs by dropping some manager log severities to debug

leemhenson:
I think the current logging is a bit verbose for ```:info```. I imagine 99% of users will be using either the built-in timing middleware or one of their own, so they'll be seeing success/failure log per message anyway. We don't need to write out another 2 log lines per successful message.


phstc:
Good call! Added to master :beers:

PR #101, opened 2015-05-10
changes to support for ruby 1.9.3

geekguy:


shoryuken.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rspec'
-  spec.add_development_dependency 'pry-byebug'
+  spec.add_development_dependency 'pry-debugger'
```
phstc: @geekguy hm this one breaks the compatibility with Ruby 2.0. If we keep `pry-byebug` as `add_development_dependency` can you still run Shoryuken with 1.9.3?

lib/shoryuken/fetcher.rb
```diff
@@ -13,7 +13,7 @@ def receive_messages(queue, limit)
       # AWS limits the batch size by 10
       limit = limit > FETCH_LIMIT ? FETCH_LIMIT : limit
 
-      options = Shoryuken.options[:aws][:receive_message].to_h.dup
+      options = (Shoryuken.options[:aws][:receive_message] || {}).to_hash.dup
```
phstc: Not sure if `to_hash` is needed here, as it returns `self`.

geekguy: @phstc : Are we using ```to_h```  only to convert ```nil``` to empty hash?

phstc: @geekguy correct, just that.

geekguy: @phstc : I have removed `to_hash`.

phstc:
Alright @geekguy. Added to master :beers: 

The only change I  reverted was the `pry-debugger` one, as it breaks compatibility with Ruby 2.0.0+.

Issue #88, opened 2015-04-08
Memory Leak?


sschepens:
Memory seems to grow slowly but without limit when using shoryuken with rails environment, a single worker and waiting on an empty queue.

Any ideas what could be causing this behaviour?

phstc:
hey @sschepens which version of Shoryuken are you using? Old versions of Shoryuken had a problem with `message_attributes`, but @elsurudo fixed that https://github.com/phstc/shoryuken/commit/d0e69dc54eaebbd0f3ebe1622577857aa14a0b90

sschepens:
@phstc I'm using latest version

leemhenson:
ruby -v?

sschepens:
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux-gnu]

leemhenson:
What sort of growth are you talking about? Can you give numbers / new relic graphs etc?

sschepens:
i'll leave a process running for a few hours recording heap metrics and will come back with that info

phstc:
Any chance it's your worker itself causing the memory leak? Could you try to run Shoryuken with an empty `perform`?

Some questions for troubleshooting:

* How many Shoryuken processes do you have?
* What's the concurrency configured for each?
* What are your queues priorities?
* What's your host/container/whatever CPU (cores) and memory?

sschepens:
Only one process with 5 concurrency, only one queue with a priority of 2.
This behaviour happens on my laptop and in an EC2 instance, the process starts with around 150-160Mb of memory according to htop, and on an EC2 instance the process has been running for a day and it now consumes 700Mb. This instance hast 2000Mb of ram and 1 core.
On my laptop (16Gb ram and 4 cores) after an hour it now consumes 200Mb and continues to grow but very slowly. I have recorded heap metrics every two minutes on my laptop but cant seem to find something that ould indicate the growth of 40 mb.

I dont think running with an empty perform will do much, as i'm hanging on an empty queue.

leemhenson:
What happens if you stick in an occasional ```GC.start```?

sschepens:
That is exactly what i was doing now, and recording metrics before and after gc start

phstc:
@sschepens sorry I missed that part:

> waiting on an empty queue.

You are right, no point on testing an empty work.

I will try to run the same test here to see how it goes.

phstc:
[...]

PR #104, opened 2015-06-04
Removed use of defer

gshutler:
Following [sidekiq's lead][1] this is no longer necessary. It also means that `Shoryuken::Shutdown` will be raised on the worker thread which, in turn means it is possible to create middleware that respond to it.

---

What lead me here was looking for a way to react to `Shoryuken::Shutdown` within middleware. This is so that I can set a low visibility timeout which means the messages will become visible much quicker for reprocessing.

The use of `defer` means that the worker is executing in a second thread and so `Shoryuken::Shutdown` was being sent to its parent instead.

As an aside, using less threads will likely be better for performance too.

The only downside I can imagine for this is that any custom middleware with a blanket `rescue` will start catching these previously uncatchable errors. However, it's pretty easy to ignore them if this is a problem with:

```ruby
rescue Shoryuken::Shutdown
  raise
rescue => e
  # whatever
```

 [1]: https://github.com/mperham/sidekiq/issues/919

### DIFF
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -33,11 +33,14 @@ module Shoryuken
 
     private
 
-    def auto_visibility_timeout(queue, sqs_msg, worker_class)
-      if worker_class.auto_visibility_timeout?
+    class MessageVisibilityExtender
+      include Celluloid
+      include Util
+
+      def auto_extend(queue, sqs_msg, worker_class)
         queue_visibility_timeout = Shoryuken::Client.queues(queue).visibility_timeout
 
-        timer = every(queue_visibility_timeout - 5) do
+        every(queue_visibility_timeout - 5) do
           begin
             logger.debug "Extending message #{worker_name(worker_class, sqs_msg)}/#{queue}/#{sqs_msg.message_id} visibility timeout by #{queue_visibility_timeout}s."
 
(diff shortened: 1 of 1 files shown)


gshutler:
In case it's useful, here's a middleware that makes use of `Shoryuken::Shutdown`:

```ruby
class ShutdownRequeueMiddleware
  include Shoryuken::Util

  def call(worker, queue, sqs_msg, body)
    yield
  rescue Shoryuken::Shutdown
    set_visible(sqs_msg)
    raise
  end

  private

  def set_visible(sqs_msg)
    sqs_msg.change_visibility(visibility_timeout: 0)
  rescue
    # fail silently
  end
end
```

phstc:
@gshutler nice :beers: 

>  continuous-integration/travis-ci/pr — The Travis CI build failed

I need to fix Travis. This error "unable to sign request without credentials set" is because a missing configuration in Travis. But this one "expected: 1 time with arguments: (15)" is a real one , I can dig into that.

> [Use of defer was necessary when Celluloid message dispatch was done within fibers. This requires twice the number of threads. Remove it now that thread dispatch is the default.](https://github.com/mperham/sidekiq/issues/919)

@sschepens @esebastian maybe this change can help with the memory leak problem #88.

phstc:
Found the problem, `every` is no longer working without `defer`: https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/processor.rb#L42

I've also tried [timers](https://github.com/celluloid/timers) directly with no luck. The only way I could make it work was using a second actor. Not sure if I'm happy with that yet.

Anyway, I'm still debugging it :smile:

phstc:
Closing in favor of #105

gshutler:
Ah, sorry. I didn't notice that some of the specs weren't running with a plain `rspec` call.

phstc:
@gshutler awesome! great work. Added to master.

esebastian:
@phstc Great, I'll take it for a spin and keep you informed!

gshutler:
Excellent, thanks for your help on it @phstc (and Shoryuken as a whole)!

PR #105, opened 2015-06-04
Try to fix auto visibility problem after  "Removed use of defer"

phstc:
Related https://github.com/phstc/shoryuken/pull/104#issuecomment-109044536

### DIFF
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -19,12 +19,10 @@ module Shoryuken
       timer = auto_visibility_timeout(queue, sqs_msg, worker.class)
 
       begin
-        defer do
-          body = get_body(worker.class, sqs_msg)
+        body = get_body(worker.class, sqs_msg)
 
-          worker.class.server_middleware.invoke(worker, queue, sqs_msg, body) do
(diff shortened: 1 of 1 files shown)


gshutler:
Hey @phstc I don't know Celluloid well enough yet to know if using an actor is a good idea or not, I'll read up on it though. However, if it is the right approach I think the actor should probably be lazily initialized as creating it creates a thread and many consumers probably don't use the feature.

I've made the changes for that, along with some other little tweaks on my fork [git compare](https://github.com/phstc/shoryuken/compare/master...gshutler:no-defer) as it was a bit hard to explain in comments alone.

phstc:
@gshutler I like what you did! :smile: I re-opened #104, could you push these changes to `gshutler:no-defer` to auto update your PR?

phstc:
Thanks @leemhenson. I agree with you :smile: 

I was just worried on introducing a new actor `MessageVisiblity`, this part is not available in Sidekiq, it's specific to Shoryuken.

gshutler:
Looks like the changes are already there, I did them on the same branch.

phstc:
Closing in favor of #104

Issue #92, opened 2015-04-13
Using Aws::SQS::QueuePoller?


sschepens:
Instead of manually sleeping :delay seconds for every queue which doesn't have messages why not use Aws::SQS::QueuePoller which is part of aws-sdk? http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/QueuePoller.html
I think it's a great alternative and could save us from maintaining some code, like queue pausing.
I'm trying to port current code to using it, but i don't understand yet if there could be problems if for example Fetcher's fetch method blocked polling and dispatching messages until Aws::SQS::QueuePoller ended for some reason or an exception was raised.

Maybe the creators/maintainers could check this up.

Cheers!
Sebastian

phstc:
Hey @sschepens could you have a look at https://github.com/phstc/shoryuken/issues/30

I will try to document / sum up this in the wiki.

sschepens:
Thanks for the link, one question i have is, is loadblancing between queues actually necessary? If so, it could be enabled somewhere else, keep one open poller for every queue and maybe loadbalance on dispatch time or spawn more workers of those who have more priority.

sschepens:
I realize that it's not a big cost to do a lot o SQS requests, you're not taking account networking costs anyway, but it may consume significant network bandwidth to make requests all the time and may impact application performance.
Also, long_polling gets you the message as soon as there is one, with the current delay system we're just sleep flat seconds to retry later, increasing the time to receive a message.

phstc:
Not sure if I got your point. But basically the motivation on that was to have only one process `shoryuken` being able to consume message from N queues. And if you use a `fetcher` per queue (because of the long poll), you need to balance the processors across fetchers etc. In the Shoryuken conception I did some experiments and I chose the simpler implementation as the other didn't result in much benefit compared with its complexity.

Imagine you have 10 processors, and you do a long poll in 5 queues, and you receive 10 messages from each. How can you process those messages?

sschepens:
Yes, I realized that, that's why I went for the approach of having a separate fetcher for each queue and a configurable pool of processors per queue.

phstc:
If you have 10 queues, 25 processors each, and messages only in one queue. You will have 225 processors waiting, and only 25 working hard.

Or if you have 10 queues, 25 processors each, and all queues full of messages, will our container handle 250 processors up & running?

That's the why Shoryuken load balances the processors across queues.

[...]

PR #106, opened 2015-06-05
Extracted queue polling strategy

gshutler:
In order to be able to customize the order and manner in which Shoryuken polls SQS queues, the previous hard-coded strategy was extracted into a separate class.

This meant that some of the method names were altered to be more generic (for example rebalance_queue_weight! became messages_received), otherwise this was a direct extraction.

---

I don't think this is 100% ready for merging, I'm not entirely happy with the names and so forth but I think it's solid enough to start having a discussion around.

My motivation was to be able to have more strict prioritization semantics than provided by the default implementation. I also wanted the option to do an extended poll on the highest priority queue if all queues are empty rather than either continually polling them all or sleeping to not poll anything.

This is my custom strategy that achieves that:

```ruby
class StrictPriorityPolling
  include Shoryuken::Util

  def initialize(queues)
    @active_queues = unparse_queues(queues)
    @prioritized_queues = @active_queues.sort_by { |queue, priority| -priority }.map(&:first)
    @highest_priority = @prioritized_queues.first
    @queues_since_messages = 0
  end

  def active_queues
    @active_queues
  end

  def next_queue
    options = {}

    queue = @prioritized_queues[@queues_since_messages % @prioritized_queues.length]

    if @highest_priority == queue && @queues_since_messages > 0
      options[:wait_time_seconds] = 2
    end

    @queues_since_messages += 1

    Shoryuken::Polling::QueueConfiguration.new(queue, options)
  end

  def messages_present(queue)
    @queues_since_messages = 0
  end

  def pause(queue)
    # No-op
  end

  def restart(queue)
    # No-op
  end
end

  config.options[:polling_strategy] = StrictPriorityPolling
end
```

From this, a strategy could be developed that may satisfy #92.

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -19,6 +19,7 @@ require 'shoryuken/middleware/server/exponential_backoff_retry'
 require 'shoryuken/middleware/server/timing'
 require 'shoryuken/sns_arn'
 require 'shoryuken/topic'
+require 'shoryuken/polling'
 
 module Shoryuken
   DEFAULTS = {
@@ -26,7 +27,8 @@ module Shoryuken
     queues: [],
     aws: {},
     delay: 0,
-    timeout: 8
+    timeout: 8,
+    polling_strategy: Polling::WeightedRoundRobin,
   }
 
   @@queues = []
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -9,18 +9,6 @@ module Shoryuken
       @manager = manager
     end
 
-    def receive_messages(queue, limit)
-      # AWS limits the batch size by 10
-      limit = limit > FETCH_LIMIT ? FETCH_LIMIT : limit
-
-      options = (Shoryuken.options[:aws][:receive_message] || {}).dup
-      options[:max_number_of_messages] = limit
-      options[:message_attribute_names] = %w(All)
-      options[:attribute_names] = %w(All)
-
-      Shoryuken::Client.queues(queue).receive_messages options
-    end
-
     def fetch(queue, available_processors)
       watchdog('Fetcher#fetch died') do
         started_at = Time.now
@@ -28,23 +16,23 @@ module Shoryuken
         logger.debug { "Looking for new messages in '#{queue}'" }
 
         begin
-          batch = Shoryuken.worker_registry.batch_receive_messages?(queue)
+          batch = Shoryuken.worker_registry.batch_receive_messages?(queue.name)
           limit = batch ? FETCH_LIMIT : available_processors
 
           if (sqs_msgs = Array(receive_messages(queue, limit))).any?
             logger.info { "Found #{sqs_msgs.size} messages for '#{queue}'" }
 
             if batch
-              @manager.async.assign(queue, patch_sqs_msgs!(sqs_msgs))
+              @manager.async.assign(queue.name, patch_sqs_msgs!(sqs_msgs))
             else
-              sqs_msgs.each { |sqs_msg| @manager.async.assign(queue, sqs_msg) }
+              sqs_msgs.each { |sqs_msg| @manager.async.assign(queue.name, sqs_msg) }
             end
 
-            @manager.async.rebalance_queue_weight!(queue)
+            @manager.async.messages_present(queue)
           else
             logger.debug { "No message found for '#{queue}'" }
 
-            @manager.async.pause_queue!(queue)
+            @manager.async.queue_empty(queue)
           end
 
           @manager.async.dispatch
@@ -57,10 +45,23 @@ module Shoryuken
           @manager.async.dispatch
         end
       end
-
     end
 
     private
+
+    def receive_messages(queue, limit)
+      # AWS limits the batch size by 10
+      limit = limit > FETCH_LIMIT ? FETCH_LIMIT : limit
+
+      options = (Shoryuken.options[:aws][:receive_message] || {}).dup
+      options[:max_number_of_messages] = limit
+      options[:message_attribute_names] = %w(All)
+      options[:attribute_names] = %w(All)
+
+      options.merge!(queue.options)
+
+      Shoryuken::Client.queues(queue.name).receive_messages options
+    end
 
     def patch_sqs_msgs!(sqs_msgs)
       sqs_msgs.instance_eval do
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -1,4 +1,5 @@
 require 'shoryuken/processor'
+require 'shoryuken/polling'
 require 'shoryuken/fetcher'
 
 module Shoryuken
@@ -12,7 +13,7 @@ module Shoryuken
 
     def initialize(condvar)
       @count  = Shoryuken.options[:concurrency] || 25
-      @queues = Shoryuken.queues.dup.uniq
+      @polling_strategy = Shoryuken.options[:polling_strategy].new(Shoryuken.queues)
       @finished = condvar
 
       @done = false
@@ -99,22 +100,18 @@ module Shoryuken
       end
     end
 
-    def rebalance_queue_weight!(queue)
-      watchdog('Manager#rebalance_queue_weight! died') do
-        if (original = original_queue_weight(queue)) > (current = current_queue_weight(queue))
-          logger.info { "Increasing '#{queue}' weight to #{current + 1}, max: #{original}" }
-
-          @queues << queue
-        end
+    def messages_present(queue)
+      watchdog('Manager#messages_present died') do
+        @polling_strategy.messages_present(queue)
       end
     end
 
-    def pause_queue!(queue)
-      return if !@queues.include?(queue) || Shoryuken.options[:delay].to_f <= 0
+    def queue_empty(queue)
+      return if Shoryuken.options[:delay].to_f <= 0
 
       logger.debug { "Pausing '#{queue}' for #{Shoryuken.options[:delay].to_f} seconds, because it's empty" }
 
-      @queues.delete(queue)
+      @polling_strategy.pause(queue)
 
       after(Shoryuken.options[:delay].to_f) { async.restart_queue!(queue) }
     end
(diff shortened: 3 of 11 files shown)


gshutler:
I'll do my best to make houndci happy.

lib/shoryuken/manager.rb
```diff
@@ -99,22 +100,18 @@ def assign(queue, sqs_msg)
...
-        end
+    def messages_present(queue)
+      watchdog('Manager#messages_present died') do
+        @polling_strategy.messages_present(queue)
       end
     end
 
-    def pause_queue!(queue)
-      return if !@queues.include?(queue) || Shoryuken.options[:delay].to_f <= 0
+    def queue_empty(queue)
```
gshutler: I can't see how to fix this. The method itself seems pretty straightforward to me.

gshutler: Actually, it seems to be happy with the extraction of the `delay` method.

gshutler:
I think the latest RSpec has changed equality behaviour and so the tests are now failing. I'll get that fixed later today.

phstc:
Hi @gshutler how is that going so far? Are you using it in your setup?

gshutler:
It's been working well. Been running for at least a couple of weeks in production now without issue.

lib/shoryuken/fetcher.rb
```diff
@@ -57,11 +45,24 @@ def fetch(queue, available_processors)
           @manager.async.dispatch
         end
       end
-
     end
 
     private
 
+    def receive_messages(queue, limit)
```
Senjai: Reason for making this private? This is a breaking api change.

lib/shoryuken/polling.rb
```diff
@@ -0,0 +1,88 @@
+module Shoryuken
+  module Polling
+    QueueConfiguration = Struct.new(:name, :options) do
+      def hash
+        name.hash
+      end
+
+      def ==(other)
```
Senjai: This is complex and confusing. Don't compare QueueConfigurations to Strings. Use queue_configuration.name == instead.

lib/shoryuken/manager.rb
```diff
@@ -148,6 +145,10 @@ def real_thread(proxy_id, thr)
 
     private
 
+    def delay
+      Shoryuken.options[:delay].to_f
```
Senjai: Is it ever possible for this to call to_f on nil?

lib/shoryuken/manager.rb
```diff
@@ -12,7 +13,7 @@ class Manager
 
     def initialize(condvar)
       @count  = Shoryuken.options[:concurrency] || 25
-      @queues = Shoryuken.queues.dup.uniq
+      @polling_strategy = Shoryuken.options[:polling_strategy].new(Shoryuken.queues)
```
Senjai: We call this a polling strategy, but the class name that's actually used for the strategy isn't called a strategy at all. Nor is the namespace. This is confusing.

lib/shoryuken.rb
```diff
@@ -26,7 +27,8 @@ module Shoryuken
     queues: [],
     aws: {},
     delay: 0,
-    timeout: 8
+    timeout: 8,
+    polling_strategy: Polling::WeightedRoundRobin,
```
Senjai: :+1: All for extracting this kind of stuff out. It makes things more extensible. However we probably want a more concrete interface here. I would expect this to be a `PollingStrategy::WeightedRoundRobin`. I would also expect a `PollingStrategy::Base` that I could inherit from that enforces a certain interface if I wanted to write my own.

lib/shoryuken/manager.rb
gshutler: Doesn't matter if it is:

```ruby
irb(main):001:0> nil.to_f
=> 0.0
```

lib/shoryuken/manager.rb
gshutler: In my opinion calling it `WeightedRoundRobinStrategy` adds nothing over it being called `WeightedRoundRobin`.

lib/shoryuken/fetcher.rb
gshutler: Because it can be without breaking the implementation, reducing the surface area of the class.

lib/shoryuken/polling.rb
gshutler: Implementing equality in this way meant that the tests did not have to change making the intent of the change more obvious.

lib/shoryuken/fetcher.rb
```diff
@@ -9,42 +9,30 @@ def initialize(manager)
...
-
     def fetch(queue, available_processors)
       watchdog('Fetcher#fetch died') do
         started_at = Time.now
 
         logger.debug { "Looking for new messages in '#{queue}'" }
 
         begin
-          batch = Shoryuken.worker_registry.batch_receive_messages?(queue)
+          batch = Shoryuken.worker_registry.batch_receive_messages?(queue.name)
```
mariokostelac: Do we have to change this one?
It's changing the interface of the method.

lib/shoryuken/fetcher.rb
```diff
@@ -9,42 +9,30 @@ def initialize(manager)
...
             if batch
-              @manager.async.assign(queue, patch_sqs_msgs!(sqs_msgs))
+              @manager.async.assign(queue.name, patch_sqs_msgs!(sqs_msgs))
             else
-              sqs_msgs.each { |sqs_msg| @manager.async.assign(queue, sqs_msg) }
+              sqs_msgs.each { |sqs_msg| @manager.async.assign(queue.name, sqs_msg) }
             end
 
-            @manager.async.rebalance_queue_weight!(queue)
+            @manager.async.messages_present(queue)
```
mariokostelac: Queue being empty should not be a special case on interface level. If we go with this approach, I'd rather have `@manager.async.messages_found(q, number_of_message)`.

mariokostelac: Ok, I see the difference now.

lib/shoryuken/manager.rb
```diff
@@ -157,50 +158,33 @@ def build_processor
...
-      return nil if @queues.empty?
-
       # get/remove the first queue in the list
-      queue = @queues.shift
+      queue = @polling_strategy.next_queue
+
+      return nil unless queue
 
-      unless defined?(::ActiveJob) ||  !Shoryuken.worker_registry.workers(queue).empty?
+      if queue && (!defined?(::ActiveJob) && Shoryuken.worker_registry.workers(queue.name).empty?)
```
mariokostelac: if queue evaluates to something falsy we can't hit we've already returned

lib/shoryuken/manager.rb
mariokostelac: We can't break the config in minor versions. WeightedRoundRobin has to be set as default!

mariokostelac:
I was a little bit sceptic, but idea is not a bad one. The PR is too intrusive though. I'd really like to cut the scope of this one and make sure we are just isolating that component. Cutting the surface area of class interface and changing method interfaces should be part of some other PRs.

lib/shoryuken/fetcher.rb
gshutler: If I remember correctly, it's a long time ago now, this line was maintaining the interface. The refactor mean that `fetch` received a queue object (`QueueConfiguration` by the looks of it) rather than just a `String` and so for downstream calls I had to pass `queue.name` so that they received a `String` as before and therefore didn't need to change.

gshutler:
I'm don't think there's a way to introduce a strategy without it being intrusive. As mentioned in my last comment I tried to minimize the changes made to other methods that weren't directly related to making the polling strategy work.

I think the only thing that could be undone whilst retaining the functionality is making `receive_messages` public again.

lib/shoryuken/manager.rb
mariokostelac: All good, I see it now!

lib/shoryuken/polling.rb
mariokostelac: I think we should go without it. It's weird. `==` is very tricky, tricky == are even more trickier :).

mariokostelac:
@gshutler sorry, you're right. I've been looking at PR more closely and it looks good. What I'd like to do is to go maybe a little bit further.
It seems that pausing and restart queues is part of the strategy. There is actually no need to have all that complexity in manager. If all queues are paused, we can just return no messages and that's it. Strategies should actually be `fetchers` so we just call something like `fetch` and we get some work or we get no work.
Do you have time to work on it some more time or you want us to work on top of your commits?
It's tremendous start to move towards "write your own fetcher" thing and make it totally extensible.

gshutler:
@mariokostelac that sounds like it makes sense I did try that route at the time but it changed so much internally that I didn't think it would make a very polite PR!

I don't have the capacity to work on this at the moment, I ended up creating a custom message processor based on a forking model because of [the memory leaks that our application experienced](https://github.com/phstc/shoryuken/issues/88#issuecomment-118030740) that I was never able to get to the bottom of.

Feel free to extend on this work or create a new branch with some of the ideas from it. I do watch this repository so would be happy to cast my eye over the changes if you think that would be helpful.

mariokostelac:
Closed in favor of https://github.com/phstc/shoryuken/pull/236

phstc:
Hey @gshutler 

I know that it might be too late, but have you seen the FIFO queues?  It can guarantee the message processing order. 

https://github.com/phstc/shoryuken/wiki/FIFO-Queues

gshutler:
@phstc Yeah, they look pretty interesting, though I do think that generally if you rely on precise message ordering you're doing messaging wrong as (for example) at some point a message will fail and then should you stop processing all other messages?

We were more looking for "if both queues have messages, we should process the ones from the high queue first". This is because (to simplify) in general our "low" messages are perfectly fine to be processed any time in the next 60 minutes, but our "high" ones should be processed as soon as possible.

phstc:
@gshutler I agree with you, they fit for very specific cases, in the cases they do, it's very useful.

> We were more looking for "if both queues have messages, we should process the ones from the high queue first"

I see that, @atyndall  also needs that. I'm starting to work on the concurrent-ruby migration, and after that I plan to review #236 and #263, these PRs should support that. Meanwhile, the only way to do that, is having more than one `shoryuken` process.

PR #108, opened 2015-06-05
Remove constantize within worker_name

gshutler:
Constantize is non-trivial and as the result is only used to generate a string it seems redundant.


zakcrawford:
I concur.

phstc:
:+1: 

Added to master :beers:

PR #113, opened 2015-06-20
Use block-style logging in all places

gshutler:
Non-block style logging seems to be, at least partially, responsible for a form of memory leak depending on your configuration.

See https://github.com/phstc/shoryuken/issues/88#issuecomment-113598856 for details of how I came to decide to make this change. I know it doesn't really make sense but I've explained how to reproduce my results. Let me know if you need any more details.

Here's a chart of memory usage on one of our production servers showing a before and after:

screen_shot_2015-06-20_at_06_57_56

This might be the first of a few PRs as I don't think the problem is solved. My gut feel is that the baseline memory usage should still be higher than it is.

### DIFF
--- a/lib/shoryuken/cli.rb
+++ b/lib/shoryuken/cli.rb
@@ -41,7 +41,7 @@ module Shoryuken
       @launcher = Shoryuken::Launcher.new
 
       if callback = Shoryuken.start_callback
-        logger.info { "Calling Shoryuken.on_start block" }
+        logger.info { 'Calling Shoryuken.on_start block' }
         callback.call
       end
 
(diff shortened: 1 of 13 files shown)


gshutler:
Let me know if you want me to appease HoundCI. I don't think I can do anything about the "line is too long" ones, but I could sort out the "single-quoted strings" easily enough.

phstc:
Hey @gshutler 

If you can update the "single-quoted strings" that would be cool. Don't worry with the "line is too long".

Great find and research. I totally see (your graphs) the change, but it still "interesting" that a simple `logger.info 'message'` can cause it.

gshutler:
There's only "line is too long" warnings now.

phstc:
@gshutler Thanks! Added to master :beers:

PR #120, opened 2015-08-06
Fix bug in ActiveJob example in README.md

camhine:


phstc:

PR #122, opened 2015-08-14
Slightly better phrasing.

calasyr:


phstc:
Great! Added to master :beers:

calasyr:
Thanks.

Looking forward to making good use of this gem.  Thanks for building it.

PR #127, opened 2015-09-01
Pass the queue into the message object on initialize.

Senjai:
Instead of just passing in the queue url, pass in the queue instead. We
can take what information we need, or delegate certain behavior if we
choose to store the queue which would require no further interface
changes.

### DIFF
--- a/lib/shoryuken/message.rb
+++ b/lib/shoryuken/message.rb
@@ -1,11 +1,21 @@
 module Shoryuken
   class Message
-    attr_accessor :client, :queue_url, :data
+    attr_accessor :client, :queue_url, :queue_name, :data
 
-    def initialize(client, queue_url, data)
+    def initialize(client, queue, data)
       self.client = client
-      self.queue_url = queue_url
       self.data = data
+
+      if queue.is_a?(Shoryuken::Queue)
+        self.queue_url = queue.url
+        self.queue_name = queue.name
+      else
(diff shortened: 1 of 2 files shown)


phstc:
Hi @Senjai 

This is basically adding the `queue_name` to the `Message` object, right? Where/how do you use this info? What's the motivation/purpose on that?

Senjai:
We use several queues and a broker to manage them. We use a general failure
queue that is managed by another application. The queue name tells us where
the message was originally placed so we can debug it from there. Its purely
informal.

The assignment makes it easy, i can remove it, if we keep the queue as an
argument I can extend it in my wrapper. Thoughts?

On Sun, Sep 6, 2015, 9:10 AM Pablo Cantero <notifications@github.com> wrote:

> Hi @Senjai <https://github.com/Senjai>
>
> This is basically adding the queue_name to the Message object, right?
> Where/how do you use this info? What's the motivation/purpose on that?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phstc/shoryuken/pull/127#issuecomment-138096046>.
>
-- 
Richard Wilson

Software Developer
FreeRunning Technologies
M: (778) 677-5318
TF: (888) 913-2661

phstc:
@Senjai got it. Makes sense. Thanks, added to master :beers:

Issue #128, opened 2015-09-03
Publisher not using shoryuken.yml


nritholtz:
I am using a Rails app in conjunction with this gem. I am loading the worker with `shoryuken -R -C config/shoryuken.yml` and everything works as I put in the config.

However, when I am using this gem to publish a message to SQS within the app, it is not using the config, and is instead using the default [AWS SDK](http://docs.aws.amazon.com/sdkforruby/api/index.html) configuration setup (e.g. environment variables).

We would like to ensure that we are using the same config for both the worker and publisher. A "hack-ish" approach that we tested out was putting inside an initializer:
```ruby
path = "#{Rails.root}/config/shoryuken.yml"
Aws.config = YAML.load(ERB.new(IO.read(path)).result).deep_symbolize_keys[:aws].except(:receive_message)
```

Is there some way to include this in the gem itself, instead of implementing this brute hack to ensure consistency across the shoryuken components?

phstc:
Hi @nritholtz 

Currently the AWS initialization is coupled with other configurations in [initialize_aws](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L63-L65) method. But extracting it would be a great idea. PR?! :heart: :beers: 

What I would also suggest to you is to use ENV variables:

```
AWS_ACCESS_KEY_ID=...
AWS_SECRET_ACCESS_KEY=...
AWS_DEFAULT_REGION=us-east-1
AWS_REGION=us-east-1
````

Using ENV variables you can omit the `aws` config in the `shoryuken.yml`. In development you can use [dotenv](https://github.com/bkeepers/dotenv) to load these variables. wdyt?

phstc:
Closed by #130

PR #130, opened 2015-09-07
Allow configuration of SQS/SNS endpoints via environment variables

nritholtz:
As a response to #128, allow SQS/SNS endpoints to be configured via environment variables, such as in use for local integration testing (e.g. [fake_sqs](https://github.com/iain/fake_sqs) ).

As far as I can see for now, this would allow full configuration on the publisher side using environment variables without requiring the Shoryuken configuration file.


phstc:

PR #136, opened 2015-09-16
Re raise body parser exceptions.

### LINKED ISSUE #135: Data loss when exceptions are raised in Shoryuken while using auto delete.
https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/processor.rb#L91-L93 should re raise the error. If you have auto delete turned on, this action simply acts as if it was completed, causing the auto delete to delete the message which causes lost data as your worker was never able to process it. I believe this also can occur in the auto extend helper as well. 

Could we re raise the exception and submit a new release? I don't think we should be catching this.

Senjai:
Fixes #135.

When a error occurs when parsing the body, crash the worker. This falls in line with celluloids "let it fail" mentality, and also prevents middleware like auto delete from progressing normally when the body couldn't be parsed.

cc @phstc (could we make another rubygems release after this one?)

### DIFF
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -90,7 +90,7 @@ module Shoryuken
       end
     rescue => e
       logger.error { "Error parsing the message body: #{e.message}\nbody_parser: #{body_parser}\nsqs_msg.body: #{sqs_msg.body}" }
-      nil
+      raise
     end
   end
 end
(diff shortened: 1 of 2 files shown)


Senjai:
Hey @phstc, just wanted to poke you once more on this one, any comments?

phstc:
@Senjai Thanks. Added to master :beers: 

I've also published a new version v2.0.0.

Senjai:
Thanks! :beer:

PR #139, opened 2015-10-10
Fix Active Job queue prefixing in Rails apps

farski:
There was an issue where Active Job prefixes were not being applied if the Shoryuken environment was being loaded outside of the CLI with the `--rails` flag (i.e. in a Rails app itself). This change allows the prefixing to occur as long as `Rails` is defined.


Senjai:
:+1:

phstc:

PR #140, opened 2015-10-10
Allow passing array of messages to send_messages

farski:
I can't come up with a good reason why this would be the intended behavior, so I think it's a bug.

If a set of messages is passed to a queue like `queue.send_messages(batch)`, where `batch` is:

```
[
    {
        message_body: "foo",
        delay_seconds: 1,
        message_attributes: {...}
    }, {
        message_body: "foo",
        delay_seconds: 1,
        message_attributes: {...}
    }
]
```

the messages are being corrupted before they actually get sent by `send_message_batch` on the client.

When [`sanitize_messages!`](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/queue.rb#L45) tries to reformat the value from an `Array` to a `Hash` that is properly wrapped and ID'd, it doesn't take into account that the members of the array may be hashes. When hashes like in the example are used, the entire hash ends up getting serialized and shoved into `message_body`.

This change properly maintains the messages and adds an index, or behaves the old way if a string body is all that's getting passed in.

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -43,7 +43,7 @@ module Shoryuken
       options = case
                 when options.is_a?(Array)
                   { entries: options.map.with_index do |m, index|
-                    { id: index.to_s }.merge(m.is_a?(Hash) ? m : { message_body: m } )
+                    { id: index.to_s }.merge(m.is_a?(Hash) ? m : { message_body: m })
                   end }
                 when options.is_a?(Hash)
                   options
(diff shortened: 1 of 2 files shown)


phstc:
Thanks @farski. Before merging it, could you review houndci feedbacks?

farski:
@phstc Fixed a couple things. Line length in specs doesn't seem like a priority, but I can try to fix it if you'd prefer.

phstc:
Added to master :beers:

PR #141, opened 2015-10-11
Add lifecycle event registration support

farski:


phstc:
Looks good :beers: 

Could you review houndci feedbacks and maybe also write something to add [here](https://github.com/phstc/shoryuken/wiki/Lifecycle-Events)?

farski:
@phstc Not much I can do about the branch condition size for that method without introducing additional methods just for the sake of Hound.

phstc:
@farski all good. Added to master :beers:

PR #144, opened 2015-10-16
Fix warnings triggered by raise_error matcher

tawan:
Fix warnings that are triggered when the `raise_error` matcher
is used without providing a specific error class or message.

The warning quoted:
>WARNING: Using the `raise_error` matcher without providing a specific error or
>message risks false positives, since `raise_error` will match when Ruby raises
>a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the
>expectation to pass without even executing the method you are intending to
>call. Instead consider providing a specific error class or message. This
>message can be supressed by setting:
>`RSpec::Expectations.configuration.warn_about_potential_false_positives =
>false`.

### DIFF
--- a/spec/shoryuken/middleware/server/exponential_backoff_retry_spec.rb
+++ b/spec/shoryuken/middleware/server/exponential_backoff_retry_spec.rb
@@ -25,7 +25,9 @@ describe Shoryuken::Middleware::Server::ExponentialBackoffRetry do
     it 'does not retry the job by default' do
       expect(sqs_msg).not_to receive(:change_visibility)
 
-      expect { subject.call(TestWorker.new, queue, sqs_msg, sqs_msg.body) { raise } }.to raise_error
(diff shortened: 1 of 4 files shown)


spec/shoryuken/middleware/server/exponential_backoff_retry_spec.rb
```diff
@@ -33,7 +35,9 @@
 
       expect(sqs_msg).not_to receive(:change_visibility)
 
-      expect { subject.call(TestWorker.new, queue, sqs_msg, sqs_msg.body) { raise } }.to raise_error
+      expect {
```
tawan: @phstc Even though @houndci is complaining, I left it as a multi-line block, since you use curly brackets in other parts as well. I hope that's fine for you.

phstc:
Cool. Added to master :beers:

PR #147, opened 2015-10-22
Enable override the default queue with a :queue option

bobbrez:
While working on a data migation I came across the need to dynamically configure the queue that a job would go to (i.e. `high_priority` vs `low_priority` or queueing data from a system singleton into production into staging).

This PR allows a client to configure which queue to send to at the time of queuing the job

```
MyWorker.perform_async('Pablo', queue: 'new_queue')
```

### DIFF
--- a/lib/shoryuken/worker.rb
+++ b/lib/shoryuken/worker.rb
@@ -6,7 +6,6 @@ module Shoryuken
 
     module ClassMethods
       def perform_async(body, options = {})
-        options ||= {}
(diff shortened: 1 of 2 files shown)


phstc:
Nice @bobbrez! Could you also update https://github.com/phstc/shoryuken/wiki/Sending-a-message? :pray: :beers:

bobbrez:
@phstc Yup. Updated

https://github.com/phstc/shoryuken/wiki/Sending-a-message

Issue #150, opened 2015-10-29
Rescue Aws::SQS::Errors::NonExistentQueue


vemv:
Aws::SQS::Errors::NonExistentQueue (which btw looks like it's an automatically generated class - not defined in aws-sdk) errors could be handled, or at the very least re-thrown with a more useful message than the AWS default.

![screen shot 2015-10-29 at 09 02 13](https://cloud.githubusercontent.com/assets/1162994/10813027/081eb93a-7e1c-11e5-804d-a0412dd1b295.png)

I'd do

```ruby
begin
    self.url    = client.get_queue_url(queue_name: name).queue_url
rescue Aws::SQS::Errors::NonExistentQueue => e
  raise StandardError.new "Non existing queue: #{name}. Please create it in advance in SQS."
end
```

Cheers - Victor

vemv:
(I don't know if it'd be correct to create automatically a queue based on the name. Would you do that?)

phstc:
I wouldn't create a queue automatically, maybe the creds supplied don't have permissions for that or the queue was typoed, creating it automatically can cause other issues.

I try to keep Shoryuken as transparent as possible to SQS, so I usually don't mind re-raising AWS exceptions, but I will check #152 carefully.

vemv:
:+1:

phstc:
@vemv wdyt https://github.com/phstc/shoryuken/pull/152#issuecomment-172382789?

vemv:
Closing this issue since it got merged!

PR #152, opened 2015-11-02
Better logging of queue existence failure

fourseven:
Took #150's suggestion and turned it into a PR.

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -5,7 +5,11 @@ module Shoryuken
     def initialize(client, name)
       self.name   = name
       self.client = client
-      self.url    = client.get_queue_url(queue_name: name).queue_url
+      begin
+        self.url = client.get_queue_url(queue_name: name).queue_url
+      rescue Aws::SQS::Errors::NonExistentQueue => e
+        raise e.class, "#{e.message} (queue: #{name})", e.backtrace
+      end
     end
 
     def visibility_timeout
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -8,7 +8,8 @@ module Shoryuken
       begin
         self.url = client.get_queue_url(queue_name: name).queue_url
       rescue Aws::SQS::Errors::NonExistentQueue => e
-        raise e.class, "#{e.message} (queue: #{name})", e.backtrace
+        e.message << " (queue: #{name})"
+        raise e
       end
     end
 


lib/shoryuken/queue.rb
```diff
@@ -5,7 +5,11 @@ class Queue
     def initialize(client, name)
       self.name   = name
       self.client = client
-      self.url    = client.get_queue_url(queue_name: name).queue_url
+      begin
+        self.url = client.get_queue_url(queue_name: name).queue_url
+      rescue Aws::SQS::Errors::NonExistentQueue
+        raise StandardError.new "Non existing queue: #{name}. Please create it in advance in SQS."
```
Senjai: This is wrong. This error is raised when you don't have permission to access the queue either. The default Aws exception is fine, I dont know why we have to raise a nondescript StandardError instead of a self explanatory `NonExistentQueue`. 

If we do end up adding this, we would also need specs for the logic change.

fourseven: The default error makes no mention to the queue which is failing, which could be something unexpected if using ActiveJob or prefixing. This may not be the best approach but it's helpful to know where to start looking on Amazon when you get this error, hence the queue-name.

Senjai: Still, if this is an issue for all was SDK users could we fix the error
output there instead of in shoryuken?

On Sun, Nov 8, 2015, 12:37 PM Mathew Hartley <notifications@github.com>
wrote:

> In lib/shoryuken/queue.rb
> <https://github.com/phstc/shoryuken/pull/152#discussion_r44230086>:
>
> > @@ -5,7 +5,11 @@ class Queue
> >      def initialize(client, name)
> >        self.name   = name
> >        self.client = client
> > -      self.url    = client.get_queue_url(queue_name: name).queue_url
> > +      begin
> > +        self.url = client.get_queue_url(queue_name: name).queue_url
> > +      rescue Aws::SQS::Errors::NonExistentQueue
> > +        raise StandardError.new "Non existing queue: #{name}. Please create it in advance in SQS."
>
> The default error makes no mention to the queue which is failing, which
> could be something unexpected if using ActiveJob or prefixing. This may not
> be the best approach but it's helpful to know where to start looking on
> Amazon when you get this error, hence the queue-name.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phstc/shoryuken/pull/152/files#r44230086>.
>
-- 
Richard Wilson

Software Developer
FreeRunning Technologies
M: (778) 677-5318
TF: (888) 913-2661

fourseven: Yes, I can look into that instead.

vemv: @fourseven did you? It can be as simple as opening an issue in https://github.com/aws/aws-sdk-ruby.

phstc: I would keep it transparent to aws-sdk as possible. I agree `NonExistentQueue` is a generic issue from aws-sdk, it can be that the queue does not exist or that you don't have permissions to it.

> The default error makes no mention to the queue which is failing

Maybe we could `logger.error(...)` and `raise e` wdyt?

nickpearson: What about:

```ruby
rescue Aws::SQS::Errors::NonExistentQueue => e
  raise e.class, "#{e.message} (queue: #{name})", e.backtrace
```

This will keep the original error class and backtrace and will append the queue name to the error message.

A quick one-liner for testing in irb/pry:

```ruby
begin; raise ArgumentError, 'test'; rescue ArgumentError => e; raise e.class, "#{e.message} [modified]", e.backtrace; end
# ArgumentError: test [modified]
# from (pry):1:in `__pry__'
```

vemv: absolutely great @nickpearson! I'd go for that.

Senjai:
:+1:

nickpearson:
No problem. I still have the browser tab with the original discussion open and can post it here if anyone cares to keep it around.

lib/shoryuken/queue.rb
```diff
@@ -5,7 +5,11 @@ class Queue
     def initialize(client, name)
       self.name   = name
       self.client = client
-      self.url    = client.get_queue_url(queue_name: name).queue_url
+      begin
+        self.url = client.get_queue_url(queue_name: name).queue_url
+      rescue Aws::SQS::Errors::NonExistentQueue => e
+        raise e.class, "#{e.message} (queue: #{name})", e.backtrace
```
Onumis: Since automatically creating the queue can be too much intrusive, I'd say appending "Did you forget to create this queue?" might be a good aid for starters facing this error.

fourseven:
Turns out you can't re-raise an Aws exception, because they have a non-standard initializer (they require a context as well as the message). I'll need to investigate this further.

nickpearson:
What if you `<<` the queue name onto the error's `message`? Then just reraise the existing exception without having to construct a new one. As long as AWS doesn't `freeze` the message string, this should work.

One-liner you can test in irb/pry:

```ruby
begin; raise ArgumentError, 'test'; rescue ArgumentError => e; e.message << " [modified]"; raise e; end
# ArgumentError: test [modified]
# from (pry):1:in `__pry__'
```

phstc:
What do you guys think about:

```ruby
def initialize(client, name)
  # ...
  begin
    self.url = client.get_queue_url(queue_name: name).queue_url
  rescue Aws::SQS::Errors::NonExistentQueue => e
    raise Shoryuken::NonExistentQueue.new "The specified queue #{name} does not exist"
  end
end
```

I usually try to keep Shoryuken as transparent as possible to AWS SQS, but I understand that the default message for `NonExistentQueue` does not tell much.

```ruby
e.message << " (queue: #{name})"
```

^ is super clever :star:, but it feels a little fragile. What do you think about it?

vemv:
I agree that modifying an error message that we didn't write isn't super clean.

I like your solution - I keeps the original exception available (and untouched), while giving the info we want to read.

phstc:
Thank you all, merged into master!

https://github.com/phstc/shoryuken/commit/277638389e3e6d3bf88aa94c9a31f7080455488c#diff-0a96d9edb6958eeb395f1a9a8fe99c1dR11

It seems that:

```ruby
raise e, "The specified queue '#{name}' does not exist"
```

Did the trick re-raising the `NonExistentQueue` changing the message and keeping the backtrace :sparkles:

PR #153, opened 2015-11-06
Allow a worker to be used for multiple queues.

Senjai:
This allows workers to be used for multiple queues. There may be several queues from several applications, where the process logic is 100% the same. To handle this currently you need one class per queue which violates dry if we could simply have Shoryuken use the same class for each queue.

An example of where this is a requirement would be a failures interface for several applications based on SQS. Each application has its' own dead letter queue, but you can use a common interface to represent and store failures from each dead letter queue.

Additionally, if you work with queues outside of your codebase, the segregation may work for a third party, but may be indifferent in your application, so processing them as separate flows would not make sense.

This deprecates the queue parameter in favour of a queues parameter and allows for a soft cutover to the new logic.


Senjai:
Could we change the hound configuration to not care about meaningless things?

Senjai:
Failure unrelated.

Senjai:
@phstc Any comment on this one?

phstc:
@Senjai I still need to test it locally. But currently shoryuken supports [multiple queues for the same worker](https://github.com/phstc/shoryuken/wiki/Sending-a-message):

```ruby
MyWorker.perform_async('Pablo', queue: 'important') # will queue to the 'important' queue
```

How the shoryuken.yml would look like with your change? Your change is to define multiple queues in there? Could you send me a code snippet showing how to use your changes?

> Could we change the hound configuration to not care about meaningless things?

I prefer to keep the code base concise, keeping the same formatting for all code. The motivation on that is to avoid distractions while maintaining the code. Could you review those houndci feedbacks?

Senjai:
@phstc I don't use the workers like that. Our worker application runs standalone, and doesn't run with rails.

```ruby
class MyWorker
  include Shoryuken::Worker
  shoryuken_options queues: ['queue1', 'queue2']
end
```

This was previously not supported. Shoryuken.yml would not change as it defines what queues to listen to, not what workers to use. This simply registers the worker twice, once for each queue.

> Prefer single-quoted strings when you don't need string interpolation or special symbols.

This goes against the grain of most style guides, and the argument of single quotes being more performant is a moot one. 

From the github style guide:

> Use double-quoted strings. Interpolation and escaped characters will always work without a delimiter change, and ' is a lot more common than " in string literals.

In CanCanCan I don't really care which literal contributors use, I'd rather have the contributor focus on providing good solutions rather than worry about what quotes they use.

phstc:
@Senjai I see your use case, I will test it locally this week, tks :beers: 

> This goes against the grain of most style guides, and the argument of single quotes being more performant is a moot one.

My motivation on single vs double quotes is not about performance, it's more about being consistent, and to do that, we need to pick one style:

> https://github.com/bbatsov/ruby-style-guide
> (Option A) Prefer single-quoted strings when you don't need string interpolation or special symbols such as \t, \n, ', etc.
> (Option B) Prefer double-quotes unless your string literal contains " or escape characters you want to suppress.

I chose Option A.

> If you feel strongly about it, I can change it for this PR.

Could you? :heart: 

[Other projects do the same about code convention](http://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#write-your-code).

Senjai:
@phstc Bumped

phstc:
@Senjai I still think that multiple queues per worker is an edge case, for instance Sidekiq does not support it. So I wouldn't have `queues` instead of `queue` as a `shoryuken_options`.

While I'm still thinking about it, I've also created a new PR #164 which allows `queue` to be a `string`, `proc` or `array`: https://github.com/phstc/shoryuken/pull/164/files#diff-bd2ce13e624208d2b1e127de91c67eaeR71 - wdyt?

PR #174, opened 2016-01-20
Compatibility with Rails 5

avokhmin:
According to https://github.com/rails/rails/pull/19295

Fixed:

> NoMethodError: undefined method `enqueue' for #<ActiveJob::QueueAdapters::ShoryukenAdapter:...>

### DIFF
--- a/lib/shoryuken/extensions/active_job_adapter.rb
+++ b/lib/shoryuken/extensions/active_job_adapter.rb
@@ -15,45 +15,43 @@ module ActiveJob
     #
     #   Rails.application.config.active_job.queue_adapter = :shoryuken
     class ShoryukenAdapter
-      class << self
-        def enqueue(job) #:nodoc:
-          register_worker!(job)
+      def enqueue(job) #:nodoc:
+        register_worker!(job)
 
-          queue = Shoryuken::Client.queues(job.queue_name)
-          queue.send_message(message(job))
-        end
+        queue = Shoryuken::Client.queues(job.queue_name)
+        queue.send_message(message(job))
+      end
 
-        def enqueue_at(job, timestamp) #:nodoc:
-          register_worker!(job)
+      def enqueue_at(job, timestamp) #:nodoc:
+        register_worker!(job)
 
-          delay = (timestamp - Time.current.to_f).round
-          raise 'The maximum allowed delay is 15 minutes' if delay > 15.minutes
+        delay = (timestamp - Time.current.to_f).round
+        raise 'The maximum allowed delay is 15 minutes' if delay > 15.minutes
 
-          queue = Shoryuken::Client.queues(job.queue_name)
-          queue.send_message(message(job, delay_seconds: delay))
-        end
+        queue = Shoryuken::Client.queues(job.queue_name)
+        queue.send_message(message(job, delay_seconds: delay))
+      end
 
-        private
+      private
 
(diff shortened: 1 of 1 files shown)


avokhmin:
Mayby I missed something, but works for me.

phstc:
hi @avokhmin 

Because of the indentation changes, it's kind of hard to see the changes to make it compatible with Rails 5, could you update your PR to only include the changes Shoryuken needs for Rails 5, skipping the indentation changes?

Is your change also compatible with Rails 4?

avokhmin:
> Because of the indentation changes, it's kind of hard to see the changes to make it compatible with Rails 5, could you update your PR to only include the changes Shoryuken needs for Rails 5, skipping the indentation changes?

`#enqueue` and `#enqueue_at` should not be more as class methods, so, I only removed `class << self`.

> Is your change also compatible with Rails 4?

No( They break the back compatibility (see https://github.com/rails/rails/pull/19295/files#diff-fa86cbc86dd18818def18324430d4396R15 for example)

But we can add hook: call class methods from instance methods.

nickpearson:
Add `?w=1` to the URL when you're on the **Files** tab to see the changes with whitespace ignored:

https://github.com/phstc/shoryuken/pull/174/files?w=1

gshutler:
To expand on what @avokhmin said, something like this will probably be enough to provide cross-compatibility for Rails 4 and 5:

```ruby
class ShoryukenAdapter
  class << self
    def instance
      @instance ||= new
    end

    def enqueue(job)
      instance.enqueue(job)
    end

    def enqueue_at(job, timestamp)
      instance.enqueue(job, timestamp)
    end
  end
end
```

phstc:
> Add ?w=1 to the URL when you're on the Files tab to see the changes with whitespace ignored:

TIL ^  tks @nickpearson 

> To expand on what @avokhmin said, something like this will probably be enough to provide cross-compatibility for Rails 4 and 5:

@gshutler @avokhmin thank you both! Added to master :star: :beers:

avokhmin:
:+1:

avokhmin:
PS: maybe try to add the `shoryuken` adapter to the rails?

> https://github.com/rails/rails/search?utf8=✓&q=shoryuken
> We couldn’t find any code matching 'shoryuken'

phstc:
hi @avokhmin interesting. I've just created the issue on the Rails repo, let's see how it goes.

https://github.com/rails/rails/issues/23311

avokhmin:
> hi @avokhmin interesting. I've just created the issue on the Rails repo, let's see how it goes.

Perfect! Let's wait that they will say

Startouf:
Hey @phstc, I believe this broke the compatibility with Shoryuken::Later that was overriding the enqueue_at class instance to provide a means to start jobs after 15minutes

(there : https://github.com/joekhoobyar/shoryuken-later/blob/develop/lib/shoryuken/later/active_job_adapter.rb#L28)

How can I fix SHoryuken::Later to make it work with the new version of Shoryuken ?

phstc:
Hi @Startouf 

I think you only need to remove [this](https://github.com/joekhoobyar/shoryuken-later/blob/develop/lib/shoryuken/later/active_job_adapter.rb#L27), as `enqueued_at` is an instance method now. 

cc/ @joekhoobyar

PR #175, opened 2016-01-26
Fix undefined method error at EnvironmentLoader

h3poteto:
When I use `Shoryuken::EnvironmentLoader#load` in Rails3, I got an raised error below.
```
/vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/shoryuken-2.0.3/lib/shoryuken/environment_loader.rb:44:in `config_file_options': undefined method `deep_symbolize_keys' for #<Hash:0x007f657fb98048> (NoMethodError)
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/shoryuken-2.0.3/lib/shoryuken/environment_loader.rb:20:in `load'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/shoryuken-2.0.3/lib/shoryuken/environment_loader.rb:6:in `load'
	from /vagrant/crowdworks/config/initializers/shoryuken.rb:17:in `<top (required)>'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:245:in `load'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:245:in `block in load'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:236:in `load_dependency'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:245:in `load'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/engine.rb:593:in `block (2 levels) in <class:Engine>'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/engine.rb:592:in `each'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/engine.rb:592:in `block in <class:Engine>'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/initializable.rb:30:in `instance_exec'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/initializable.rb:30:in `run'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/initializable.rb:55:in `block in run_initializers'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/initializable.rb:54:in `each'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/initializable.rb:54:in `run_initializers'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/application.rb:136:in `initialize!'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/railtie/configurable.rb:30:in `method_missing'
	from /vagrant/crowdworks/config/environment.rb:5:in `<top (required)>'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:251:in `require'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:251:in `block in require'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:236:in `load_dependency'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/activesupport-3.2.22/lib/active_support/dependencies.rb:251:in `require'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/application.rb:103:in `require_environment!'
	from /vagrant/crowdworks/vendor/bundle/ruby/2.1.0/gems/railties-3.2.22/lib/rails/commands.rb:40:in `<top (required)>'
	from script/rails:6:in `require'
	from script/rails:6:in `<main>'
```
This is because ActiveSupport v3.2 does not define `Hash#deep_symbolize_keys`.
So, I've added  closely check hash methods in `core_ext.rb`.

### DIFF
--- a/lib/shoryuken/core_ext.rb
+++ b/lib/shoryuken/core_ext.rb
@@ -1,35 +1,47 @@
+module Shoryuken
+  module CoreExt
+    module StringifyKeys
+      def stringify_keys
+        keys.each do |key|
+          self[key.to_s] = delete(key)
+        end
+        self
+      end
+    end
+
+    module SymbolizeKeys
+      def symbolize_keys
+        keys.each do |key|
+          self[(key.to_sym rescue key) || key] = delete(key)
+        end
+        self
+      end
+    end
+
+    module DeepSymbolizeKeys
+      def deep_symbolize_keys
+        keys.each do |key|
+          value = delete(key)
+          self[(key.to_sym rescue key) || key] = value
+
+          value.deep_symbolize_keys if value.is_a? Hash
+        end
+        self
+      end
+    end
+  end
+end
+
 begin
   require 'active_support/core_ext/hash/keys'
   require 'active_support/core_ext/hash/deep_merge'
-  {}.stringify_keys
-  {}.symbolize_keys
-  {}.deep_symbolize_keys
-rescue LoadError, NoMethodError
-  class Hash
-    def stringify_keys
-      keys.each do |key|
-        self[key.to_s] = delete(key)
-      end
-      self
-    end if !{}.respond_to?(:stringify_keys)
+rescue LoadError
+end
 
-    def symbolize_keys
-      keys.each do |key|
-        self[(key.to_sym rescue key) || key] = delete(key)
-      end
-      self
-    end if !{}.respond_to?(:symbolize_keys)
-
-    def deep_symbolize_keys
-      keys.each do |key|
-        value = delete(key)
-        self[(key.to_sym rescue key) || key] = value
-
-        value.deep_symbolize_keys if value.is_a? Hash
-      end
-      self
-    end if !{}.respond_to?(:deep_symbolize_keys)
-  end
+class Hash
+  include Shoryuken::CoreExt::StringifyKeys unless method_defined?(:stringify_keys)
+  include Shoryuken::CoreExt::SymbolizeKeys unless method_defined?(:symbolize_keys)
+  include Shoryuken::CoreExt::DeepSymbolizeKeys unless method_defined?(:deep_symbolize_keys)
 end
 
 begin
--- a/lib/shoryuken/core_ext.rb
+++ b/lib/shoryuken/core_ext.rb
@@ -1,7 +1,10 @@
 begin
   require 'active_support/core_ext/hash/keys'
   require 'active_support/core_ext/hash/deep_merge'
-rescue LoadError
+  {}.stringify_keys
+  {}.symbolize_keys
+  {}.deep_symbolize_keys
+rescue LoadError, NoMethodError
   class Hash
     def stringify_keys
       keys.each do |key|


Senjai:
This is not a reasonable way to solve this problem. 

Calling methods on hashes till any specific one of them fails, so we redefine them all is not a good idea. Instead bump the version of active support, or add a check to the version of active support which has the problem and define it conditionally if the version does not bring in the correct method.

Please try again.

h3poteto:
It is good idea that I add a check to the version of active support, but active support is not written in gemspec. So, `require` will fail when someone use this outside Rails.
I guess that shoryuken is used outside Rails.

Therefore, I add `method_defined?` for each Hash methods after call active support.

Senjai:
@h3poteto 
```ruby
ActiveSupport::VERSION::MAJOR # 3
ActiveSupport::VERSION::MINOR # 2
# e.g. 3.2
```

h3poteto:
`ActiveSupport::VERSION::MAJOR ` will raise `NameError: uninitialized constant` outside Rails.
 
We should not use …

phstc:
@h3poteto @Senjai actually I think we don't need `deep_symbolize_keys`, we only use it [here](https://github.com/phstc/shoryuken/blob/98f10685e102876d22faca13398e97e13539fe8c/lib/shoryuken/environment_loader.rb#L44). And most YAMLs are written in this way:

```yaml
key: "value"
```

Not wrapping the keys with quotes:

```yaml
"key": "value"
```

So, maybe the fix would be to remove `deep_symbolize_keys` and review all references of `Shoryuken.options` to make sure it's using a symbol instead of a string as a key. wdyt?

h3poteto:
YAML parser did not construe key as symbol, if YAMLs are written in this way:
```yaml
key: "value"
```
and `YAML.load` return string as a key:
```ruby
YAML.load("aws:\n access_key_id: hoge\n")
=> {"aws"=>{"access_key_id"=>"hoge"}}
```
But, when YAMLs are written in this way:
```yaml
:key: "value"
```
`YAML.load` return symbol.
```ruby
YAML.load(":aws:\n :access_key_id: hoge\n")
=> {:aws=>{:access_key_id=>"hoge"}}
```

So, if everyone use symbol syntax for yaml like this:
```yaml
:key: "value"
```
we can remove `deep_symbolize_keys`. I think  that we should not remove `deep_symbolize_keys`, because maybe many developers did not use symbol syntax for yaml.

phstc:
@h3poteto interesting. Because of [this](https://github.com/mperham/sidekiq/blob/e664794b6a02ca9fec4892367c14a3519560da08/lib/sidekiq/cli.rb#L363) - I thought that symbol as the default, as Sidekiq uses `YAML.load` and retrieve the options passing symbols.

But based on this:

> `YAML.load("aws:\n access_key_id: hoge\n")`
> `=> {"aws"=>{"access_key_id"=>"hoge"}}`

I would remove `deep_symbolize_keys` and review all places that use `Shoryuken.options` to make sure that they are passing a string key instead of symbol.

wdyt?

h3poteto:
@phstc Thank you for proposal.
I checked [yaml parser](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L360) in Sidekiq.
Sidekiq recommend symbol syntax for yaml in [wiki](https://github.com/mperham/sidekiq/wiki/Advanced-Options), and Sidekiq ignore a string key like this:
```ruby
YAML.load("aws:\n access_key_id: hoge\n")
=> {"aws"=>{"access_key_id"=>"hoge"}}
```
 I considered this.

I thought that like below:
> we should not remove deep_symbolize_keys, because maybe many developers did not use symbol syntax for yaml

but,  now I think that we can remove `deep_symbolize_keys` and  we should announce new `shoryuken.yml` about symbol syntax for yaml in README and wiki like Sidekiq.

Maybe, new developers will read README, when new developers use shoryuken.
And existing developers will read release note or new README, when upgrade shoryuken.

So, developers will detect new `shoryuken.yml` and can use symbol syntax for yaml like this:
```yaml
:aws:
  :access_key_id: hoge
```
I think that symbol syntax is fine with developers.

What do you think?

phstc:
hi @h3poteto 

I personally prefer:

```yaml
aws:
  access_key_id: hoge
```

But in favor of styling preferences in something that widely used `YAML`, backwards compatibility with previous versions of Shoryuken and simplicity, I merged [your change](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/core_ext.rb), and I've also added [some specs](https://github.com/phstc/shoryuken/blob/master/spec/shoryuken/core_ext_spec.rb).

Thanks a lot for your PR and all researched! :beers:

PR #176, opened 2016-01-26
update bundler before travis builds

suzan2go:
Latest builds on travis are broken because bundler is too old and contains some issues.
Refs: bundler/bundler#3558 travis-ci/travis-ci#3531


phstc:
you rock! tks :star: Added to master :beers:

PR #177, opened 2016-01-28
Allow symbol as a queue name in shoryuken_options

serihiro:
# What does this PullRequest change?

This PullRequest allows to use symbol for queue name in shoryuken_options.

e.g.

```ruby
class SymbolQueueTestWorker
  include Shoryuken::Worker

  shoryuken_options queue: :default
end
```

In this case, `:default` will be converted to `'default'`, and work.

# Why is this need?

To use Amazon SQS, I was about to convert a sidekiq worker class to a shoryuken worker class.
But my worker was not be handled by shoryuken worker process though I enqueued to SQS with valid format message...

The converted shoryuken worker class is like the following:

```ruby
class WhatOneceWasSidekiqWorker
  include Shoryuken::Worker

  shoryuken_options queue: :myworker
end
```

And I noticed that sidekiq worker accepts symbol queue name, but shoryuken does not.

But as far as I read shoryuken's codes , I could not find any special reasons that Shoryken has to force the queue name to be Symbol.

So if there are not any special reasons, I think it is better to allow the Symbol for portability from Sidekiq.

### DIFF
--- a/lib/shoryuken/worker.rb
+++ b/lib/shoryuken/worker.rb
@@ -68,6 +68,13 @@ module Shoryuken
           @shoryuken_options['queue'] = queue
         end
 
+        case @shoryuken_options['queue']
+        when Array
+          @shoryuken_options['queue'].map!(&:to_s)
+        when Symbol
+          @shoryuken_options['queue'] = @shoryuken_options['queue'].to_s
+        end
+
         [@shoryuken_options['queue']].flatten.compact.each(&method(:register_worker))
       end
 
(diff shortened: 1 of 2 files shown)


serihiro:
https://travis-ci.org/phstc/shoryuken/jobs/105426349 ruby2.2.0 build is failed. 
I think this is same problem as this https://github.com/phstc/shoryuken/pull/176.

phstc:
Tks @serihiro! Great fix, I've also merged #176 in.

serihiro:
@phstc Thank you for reviewing and merging! :beers:

PR #178, opened 2016-02-04
Check that concurrency is set to a positive number. 

bbaja42:
Without this check, setting concurrency to -1 or to 0; will actually start the manager, but manager will not spawn any workers.

I've made the check at this level, alternatively perhaps cli.rb should fail as well if number passed is negative.
What I prefer with current approach is that argument verification is next to the code actually using the argument.

`@ready = @count.times.map { build_processor }` this is a `NOP` command if `@count` is negative or zero.

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -12,6 +12,7 @@ module Shoryuken
 
     def initialize(condvar)
       @count  = Shoryuken.options[:concurrency] || 25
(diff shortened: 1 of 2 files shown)


bbaja42:
Thnx for the feedback. Will fix shortly.

bbaja42:
Any chance to get some feedback on this PR?

phstc:
@bbaja42 :beers: 

PR added to master :smile:

bbaja42:
Wooohoo, my first ruby related open source PR :)

phstc:
@bbaja42 well done :tada:

PR #184, opened 2016-02-24
add shoryukenctl control executable for shutting down

AlexWiles:
I've just set up the deployment for some servers with Shoryuken processes using the sidekiqctl script to send shutdown signals. Here I just copied the script and s/Sidekiq/Shoryuken since the signals are the same. Thought it might be useful for others :smile:

### DIFF
--- a/bin/shoryukenctl
+++ b/bin/shoryukenctl
@@ -1,0 +1,99 @@
+#!/usr/bin/env ruby
+
+require 'fileutils'
+
+class Shoryukenctl
+  DEFAULT_KILL_TIMEOUT = 10
+
+  attr_reader :stage, :pidfile, :kill_timeout
+
+  def self.print_usage
+    puts "#{File.basename($0)} - stop a Shoryuken process from the command line."
+    puts
+    puts "Usage: #{File.basename($0)} <command> <pidfile> <kill_timeout>"
+    puts " where <command> is either 'quiet' or 'stop'"
+    puts "       <pidfile> is path to a pidfile"
+    puts "       <kill_timeout> is number of seconds to wait until Shoryuken exits"
+    puts "       (default: #{Shoryukenctl::DEFAULT_KILL_TIMEOUT}), after which Shoryuken will be KILL'd"
+    puts
+    puts "Be sure to set the kill_timeout LONGER than Shoryuken's -t timeout.  If you want"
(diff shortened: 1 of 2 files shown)


achiinto:
:+1:

Senjai:
@AlexWiles How do you manage this on your servers? This should be handled by systemd/upstart/runit/monit

AlexWiles:
@Senjai I'm using upstart to manage a set of shoryuken processes. I'm doing the same thing with sidekiq in a different service and I like the convenience of sidekiqctl handling USR1 and TERM signals. I have `shoryukenctl stop $PID_FILE` in my shoryuken-upstart.conf pre-stop script.

phstc:
hm I usually use only `pkill -f` or `kill` directly. So this CLI would be just a shortcut to use `stop` instead of `USR1`?

mariokostelac:
I feel that this layer could introduce extra layer of confusion since it introduces another layer of graceful shutdown. With this new setup we have three:
1. Upstart/Monit/any process manager
2. ShoryukenCtl
3. Shoryuken (yes, there is a `timeout` param that tries to kill all processors after that period).

@AlexWiles  what's the real benefit of this one? Could you use standard `kill timeout` upstart feature?

bin/shoryukenctl
```diff
@@ -0,0 +1,99 @@
...
+    puts "#{File.basename($0)} - stop a Shoryuken process from the command line."
+    puts
+    puts "Usage: #{File.basename($0)} <command> <pidfile> <kill_timeout>"
+    puts " where <command> is either 'quiet' or 'stop'"
+    puts "       <pidfile> is path to a pidfile"
+    puts "       <kill_timeout> is number of seconds to wait until Shoryuken exits"
+    puts "       (default: #{Shoryukenctl::DEFAULT_KILL_TIMEOUT}), after which Shoryuken will be KILL'd"
+    puts
+    puts "Be sure to set the kill_timeout LONGER than Shoryuken's -t timeout.  If you want"
+    puts "to wait 60 seconds for jobs to finish, use `sidekiq -t 60` and `sidekiqctl stop"
```
eugeneius: sidekiq 😒

phstc:
I think it's an interesting idea, but just to keep the code base simpler, I'm inclined to do not add it.

PR #191, opened 2016-04-12
Fix initialize logger

phstc:
Experiment based on #181

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -15,13 +15,12 @@ module Shoryuken
     end
 
     def load
+      Shoryuken.options.merge!(config_file_options)
+      Shoryuken.options.merge!(options)
       initialize_logger
       load_rails if options[:rails]
-      Shoryuken.options.merge!(config_file_options)
-      reset_logger
       merge_cli_defined_queues
       prefix_active_job_queue_names
-      Shoryuken.options.merge!(options)
       parse_queues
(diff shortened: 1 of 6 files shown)


lib/shoryuken/environment_loader.rb
```diff
@@ -15,12 +15,12 @@ def initialize(options)
     end
 
     def load
+      Shoryuken.options.merge!(config_file_options)
+      Shoryuken.options.merge!(options)
```
phstc: @h3poteto in continuation of https://github.com/phstc/shoryuken/pull/181#issuecomment-208650924. [This](https://github.com/phstc/shoryuken/blob/crowdworks-fix-initialize-logger/lib/shoryuken/environment_loader.rb#L37) will logger out in the `stdout`, as the log wasn't initialized yet. But thinking more on it, I believe it should show an error and exit if the config file supplied is invalid. WDYT?

h3poteto: > I believe it should show an error and exit if the config file supplied is invalid.

I think it is right behavior.
LGTM

waynerobinson: This breaks loading the queues from the config when starting with the command-line client.

Basically, the default https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/cli.rb#L115 will cause `options` to always have a `:queues` key and clobber whatever is supplied in the config file.

PR #195, opened 2016-04-17
Fix option initialization for queues

laCour:
This corrects option initialization when invoking `shoryuken` via the command line, and also fixes setting configuration options using the Rails environment with gems like dotenv.

Since PR #191, queues from the configuration file were being overwritten by the command line's default options—effectively being ignored.

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -15,8 +15,7 @@ module Shoryuken
     end
 
     def load
-      Shoryuken.options.merge!(config_file_options)
-      Shoryuken.options.merge!(options)
+      initialize_options
       initialize_logger
       load_rails if options[:rails]
       merge_cli_defined_queues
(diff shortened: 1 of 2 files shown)


lib/shoryuken/environment_loader.rb
```diff
@@ -92,6 +92,7 @@ def load_rails
         require File.expand_path('config/environment.rb')
       end
 
+      Shoryuken.options.merge!(config_file_options)
```
phstc: @laCour is it a possible fix for what @waynerobinson commented out [here](https://github.com/phstc/shoryuken/pull/191/files#r59998676)?

laCour: @phstc I've updated the PR to fix this issue as well. The options passed from the CLI were destructively overwriting the options from the configuration file on merge, because an empty array was being passed.

phstc: ⭐

phstc:
@laCour Thanks a lot! Added to master 🍻

PR #197, opened 2016-04-20
Fix initialization when the Load Rails option is specified

yujideveloper:
I want to use gems like SettingsLogic with Rails environements in the Shoryuken configuration file.
So, Rails must be loaded before loading the Shoryuken configuration file.

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -15,9 +15,9 @@ module Shoryuken
     end
 
     def load
+      load_rails if options[:rails]
       initialize_options
       initialize_logger
-      load_rails if options[:rails]
       merge_cli_defined_queues
       prefix_active_job_queue_names
       parse_queues
@@ -95,11 +95,6 @@ module Shoryuken
         require 'shoryuken/extensions/active_job_adapter' if defined?(::ActiveJob)
         require File.expand_path('config/environment.rb')
       end
-
-      # Reload options with Rails environment (see PR #195)
-      initialize_options
-
-      Shoryuken.logger.info { 'Rails environment loaded' }
     end
 
     def merge_cli_defined_queues


lib/shoryuken/environment_loader.rb
```diff
@@ -15,9 +15,12 @@ def initialize(options)
...
 
     def load
-      initialize_options
-      initialize_logger
-      load_rails if options[:rails]
+      if options[:rails]
+        load_rails
+      else
+        initialize_options
+        initialize_logger
```
phstc: > I want to use gems like SettingsLogic

@yujideveloper how would that let people to use `--rails` and load settings without using `SettingsLogic` or similar?

pbernays: ERB can't be processed if using something from the Rails project, eg.
```
queues:
  - [<%= Rails.application.config.some_setting %>, 1]
  - [<%= Settings.some_other_setting %>, 1]
```
because options are initialised before the Rails environment is loaded. That config file results in an `uninitialized constant` error.

lib/shoryuken/environment_loader.rb
```diff
@@ -98,6 +101,7 @@ def load_rails
 
       # Reload options with Rails environment (see PR #195)
       initialize_options
+      initialize_logger
```
mariokostelac: why to initialise logger here?

mariokostelac:
I think you are right, we have to load rails upfront. This functionality was somehow broken in last release.

lib/shoryuken/environment_loader.rb
mariokostelac: Ok, I see.

mariokostelac:
I think approach is good, but let's add some clarity to the codeflow.
`load_rails` loads config and initialises the logger in https://github.com/yujideveloper/shoryuken/blob/654536c35187ead505e8389c1635f58f9327c065/lib/shoryuken/environment_loader.rb#L102-L104. This makes `initialize` method messy because it's not clear why logger and options do not have to be loaded if we are loading rails.
I'd rather remove log line from `load_rails` (because it requires logger) and have initializer like:
```ruby
load_rails if options[:rails]
initialize_options
initialize_logger
...
```

lib/shoryuken/environment_loader.rb
yujideveloper: Sorry for slow response.
What I want to say is same as [this comment](https://github.com/phstc/shoryuken/pull/197#discussion_r63645034).

yujideveloper:
@mariokostelac @phstc 
Thank you for review.
I've update codes. How do you think?

mariokostelac:
I've checked manually against the codebase that was not able to have shoryuken upgraded because of loading config before rails and I am happy to say it does work!
Awesome work @yujideveloper.

PR #199, opened 2016-04-23
Isolate MessageVisibilityExtender as a middleware

mariokostelac:
First cut to isolate MessageVisibilityExtender as a middleware + added some tests for that functionality.
I do not like the fact that tests depend on time (and inherently slow down the suite by 7 seconds min), but was not able to mock `every` and yield to passed block.


spec/shoryuken/middleware/server/auto_extend_visibility_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  def build_message
+    double Shoryuken::Message, queue_url: queue, body: 'test', receipt_handle: SecureRandom.uuid
+  end
+
+  # We need to run our worker inside actor context.
+  class Runner
+    include Celluloid
+
+    def run_and_sleep(worker, queue, sqs_msg, sleep_interval)
+      Shoryuken::Middleware::Server::AutoExtendVisibility.new.call(worker, queue, sqs_msg, sqs_msg.body) {}
```
phstc: hi @mariokostelac is `sleep_interval` needed? It does not seem to be used.

mariokostelac: oh sorry, it's a leftover. Will remove it.

phstc:
@mariokostelac great work 🍻 

Could you also check [this spec](https://github.com/phstc/shoryuken/blob/master/spec/shoryuken/processor_spec.rb#L274)? It isn't passing for me. Because of the `slow: true`, you need to run it as follows: `SPEC_ALL=true rspec ./spec/shoryuken/processor_spec.rb:273`.

mariokostelac:
No problem.

mariokostelac:
Hm, all tests pass for me :S.

mariokostelac:
The problem is that body parser runs before we hit the middleware so it pauses before the the `sleep` call. I am happy to harden that test suite a little bit more.

mariokostelac:
I guess body parser could also be a middleware so all of those could be wrapped in metrics, logging or any other middleware. That should be separate PR though, but would make processor even lighter.

phstc:
@mariokostelac amazing work on this. Thanks a lot, the processor cleaner than before ⭐ 🍻 

> I guess body parser could also be a middleware so all of those could be wrapped in metrics, logging or any other middleware. That should be separate PR though, but would make processor even lighter.

I like your idea ^, we could call it right after the [`Timing` middleware](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken.rb#L143). 

_@phstc is thinking out loud:_ Or also, just extracting into a class `BodyParser`, calling it in the `Processor` instead of having it implementing the parser logic. The advantage of having it in a middleware, is that people can remove, in case they want a custom parser, but as [the parser option is very flexible, it accepts a class, block etc](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/processor.rb#L41-L57), maybe just a `BodyParser` being called in the `Processor` instead of a middleware would be good enough, WDYT?

mariokostelac:
I guess you are right. The only drawback is not timing body parsing.

PR #200, opened 2016-04-27
require 'rails' before initialize config

highwide:
In order to fix this👇  (I guess it's been since https://github.com/phstc/shoryuken/pull/191 )

config/shoryuken.yml
```
queues:
  - [<%= Rails.env %>-foo-bar-queue, 2]
```

```
% bundle exec shoryuken -R -C config/shoryuken.yml
uninitialized constant Rails
...
```

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -15,6 +15,7 @@ module Shoryuken
     end
 
     def load
+      require 'rails' if options[:rails]
       initialize_options
       initialize_logger
       load_rails if options[:rails]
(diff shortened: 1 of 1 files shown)


phstc:
Hi @highwide 

I was wondering if we should ` - [<%= ENV['RAILS_ENV'] %>-foo-bar-queue, 2]` instead. Sidekiq only requires Rails, after loading the config files: https://github.com/mperham/sidekiq/blob/78fc103d561792b876854706f0e8acdc3b3379f4/lib/sidekiq/cli.rb#L222

I'm not sure if there's a specific reason for that.

highwide:
Thank you @phstc 

It's a pity but I understood that I can't use like `<%= Rails.env.production? ? '': "#{Rails.env}-" %>`
I'll use `RAILS_ENV` and close this PR.

PR #204, opened 2016-05-04
Allow Shoryuken AWS credentials to be overridden for job enqueuing

gtaschuk:
We use more than one set of keys for Aws.  While ```config/shoryuken.yml``` works well for initializing workers to use alternate Aws credentials, in order to enqueue jobs correctly from the web worker, we needed to override the credentials that ```Aws::SQS::Client``` uses.

This allows us to specify in ```config/initializers/shoryuken.rb```
```ruby
Shoryuken.options[:aws] = { 
  access_key_id: ENV['AWS_API_ACCESS_KEY_ID'],
  secret_access_key: ENV['AWS_API_SECRET_ACCESS_KEY']
}
```

### DIFF
--- a/lib/shoryuken/client.rb
+++ b/lib/shoryuken/client.rb
@@ -31,10 +31,13 @@ module Shoryuken
 
       def aws_client_options(service_endpoint_key)
         environment_endpoint = ENV["AWS_#{service_endpoint_key.to_s.upcase}"]
-        explicit_endpoint = Shoryuken.options[:aws][service_endpoint_key] || environment_endpoint
-        options = {}
-        options[:endpoint] = explicit_endpoint unless explicit_endpoint.to_s.empty?
-        options
+        aws_options = Shoryuken.options[:aws]
+        explicit_endpoint = aws_options[service_endpoint_key] || environment_endpoint
+        Hash.new.tap do |options|
+          options[:endpoint] = explicit_endpoint unless explicit_endpoint.to_s.empty?
+          options[:access_key_id] = aws_options[:access_key_id] unless aws_options[:access_key_id].nil?
+          options[:secret_access_key] = aws_options[:secret_access_key] unless aws_options[:secret_access_key].nil?
+        end
       end
     end
   end
--- a/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
+++ b/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
@@ -2,6 +2,7 @@ module Shoryuken
   module Middleware
     module Server
       class ExponentialBackoffRetry
+        MODIFIER_BASE = 10
         include Util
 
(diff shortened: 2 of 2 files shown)


phstc:
Hi @gtaschuk 

I don't think I could follow you. Could you give maybe a code sample of what you need to do?

mariokostelac:
@gtaschuk  I can understand that there could be a problem where you use queues from different AWS accounts and you have to push to both of them. Your solution tries to solve a different problem though. You still allow to use just one credentials pair per process.
As far as I understand, aws sdk will get credentials from environment variables, if they're not explicitly passed (or `.aws/credentials` file) (source: http://docs.aws.amazon.com/sdkforruby/api/#Configuration).

vemv:
@gtaschuk shoryuken.yml already accepts ERB, so you could read ENV from there, right?

mariokostelac:
@vemv it is currently broken if using constants loaded from rails, but fix is already in master.

lib/shoryuken/client.rb
```diff
@@ -31,10 +31,13 @@ def topics(name)
...
-        options = {}
-        options[:endpoint] = explicit_endpoint unless explicit_endpoint.to_s.empty?
-        options
+        aws_options = Shoryuken.options[:aws]
+        explicit_endpoint = aws_options[service_endpoint_key] || environment_endpoint
+        Hash.new.tap do |options|
+          options[:endpoint] = explicit_endpoint unless explicit_endpoint.to_s.empty?
+          options[:access_key_id] = aws_options[:access_key_id] unless aws_options[:access_key_id].nil?
+          options[:secret_access_key] = aws_options[:secret_access_key] unless aws_options[:secret_access_key].nil?
+        end
```
vemv: this code is quite ugly, can't it be DRYed up?

mariokostelac:
@vemv @gtaschuk could you explain what problem you're trying to solve?

lib/shoryuken/middleware/server/exponential_backoff_retry.rb
```diff
@@ -2,6 +2,7 @@ module Shoryuken
   module Middleware
     module Server
       class ExponentialBackoffRetry
+        MODIFIER_BASE = 10
```
mariokostelac: what is this?

mariokostelac:
@gtaschuk i got it now. Could we somehow use the same code for setting these credentials on both sides (producer and consumer)?

phstc:
Shoryuken actually sets the [`Aws.config`](https://github.com/phstc/shoryuken/blob/555953fb5e4c37933335109adc1e23d5197c382b/lib/shoryuken/environment_loader.rb#L74). So if you want to change that dynamically at runtime, re-setting it `Aws.config = config` would change the creds.

PR #205, opened 2016-05-09
Add rspec task

potomak:
See https://www.relishapp.com/rspec/rspec-core/docs/command-line/rake-task.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,12 @@
 require 'bundler/gem_tasks'
 $stdout.sync = true
 
+begin
+  require 'rspec/core/rake_task'
+  RSpec::Core::RakeTask.new(:spec)
+rescue LoadError
+end
+
 desc 'Open Shoryuken pry console'
 task :console do
   require 'pry'


phstc:
Hi @potomak 

Could you give some examples of the tasks you use? I usually execute specs only by calling `rspec`, I'm not familiar with the rake task.

potomak:
This PR adds a new `rake spec` task. It does the same thing as running `rspec`, but I added it anyway because I'm used to find it in gems and projects that use rspec, and generally I always make it the *default* rake task.

It could be useful if you want to combine it with other tasks, like making it a dependency if the *build* task to be sure that all specs are green before a new build.

mariokostelac:
Is it going to be a default rake task or task that runs before any other task?
It does seem that it will run before console task, too. Am I wrong?
Also, I'd like to have some verbosity on LoadError.

phstc:
I believe it's more a matter of being used to or not. TBH I'm not use to `rake spec`, I usually go straight with `rspec`. 

I'm trying to avoid adding more code to Shoryuken unless it's like must have, just to keep Shoryuken more stable. I know that this change is minor, but it isn't must have.

PR #206, opened 2016-05-09
Minor fetch method refactor

potomak:
Call `async.dispatch` only once at the end of the `begin` block.


phstc:
Thanks! Added to master 🍻

potomak:
👍

PR #209, opened 2016-05-10
load rails earlier

leemhenson:
Hi. A recent change has broken the boot sequence in my app. I was relying on the rails environment being loaded by the time `shoryuken.yml` was parsed, but since PR 195 the file is parsed before load rails happens. 

As far as I can see, there's no issue with moving `load_rails` to the beginning of the boot sequence and calling `initialize_options` afterwards. What do you think?

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -15,9 +15,13 @@ module Shoryuken
     end
 
     def load
+      load_rails if options[:rails]
+
       initialize_options
       initialize_logger
-      load_rails if options[:rails]
+
+      Shoryuken.logger.info { 'Rails environment loaded' } if options[:rails]
+
       merge_cli_defined_queues
       prefix_active_job_queue_names
       parse_queues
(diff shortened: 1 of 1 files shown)


leemhenson:
fyi I'm still testing this so this is for discussion only at the moment.

phstc:
> fyi I'm still testing this so this is for discussion only at the moment.

@leemhenson thanks a lot for that ^ :beers: Unfortunately, I didn't test well the latest changes (some recently merged in pull requests) on this initialization, and it started to break for some people, each PR fixed something, but broke another 😢 

It started to 💣 after 14ce22b3142c68668cc8fbd279f867b8d368148d ([history](https://github.com/musicglue/shoryuken/commits/load-rails-ordering/lib/shoryuken/environment_loader.rb)).

Any deep drive on this would be very helpful ❤️

leemhenson:
Yeah I just encountered further problems with the interaction around loading celluloid at the right time vs including it like this:

https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/middleware/server/auto_extend_visibility.rb#L19

I can't budget time to go deeper, at this point sorry! I will pin to v2.0.5.

PR #213, opened 2016-05-23
Fix auto extend error when using an ActiveJob wrapper

takeyuweb:
When using auto extend

> ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper
Could not auto extend the message ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper/<queue_name>/<message> visibility timeout. Error: undefined method `[]' for nil:NilClass

### DIFF
--- a/lib/shoryuken/util.rb
+++ b/lib/shoryuken/util.rb
@@ -42,7 +42,8 @@ module Shoryuken
           && !sqs_msg.is_a?(Array) \
           && sqs_msg.message_attributes \
           && sqs_msg.message_attributes['shoryuken_class'] \
-          && sqs_msg.message_attributes['shoryuken_class'][:string_value] == ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper.to_s
+          && sqs_msg.message_attributes['shoryuken_class'][:string_value] == ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper.to_s \
+          && body
 
         "ActiveJob/#{body['job_class']}"
       else
--- a/lib/shoryuken/util.rb
+++ b/lib/shoryuken/util.rb
@@ -42,8 +42,8 @@ module Shoryuken
           && !sqs_msg.is_a?(Array) \
           && sqs_msg.message_attributes \
           && sqs_msg.message_attributes['shoryuken_class'] \
-          && sqs_msg.message_attributes['shoryuken_class'][:string_value] == \
-            ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper.to_s \
+          && sqs_msg.message_attributes['shoryuken_class'][:string_value] \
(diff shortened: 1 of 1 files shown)


lib/shoryuken/util.rb
```diff
@@ -42,7 +42,9 @@ def worker_name(worker_class, sqs_msg, body = nil)
           && !sqs_msg.is_a?(Array) \
           && sqs_msg.message_attributes \
           && sqs_msg.message_attributes['shoryuken_class'] \
-          && sqs_msg.message_attributes['shoryuken_class'][:string_value] == ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper.to_s
+          && sqs_msg.message_attributes['shoryuken_class'][:string_value] \
+          == ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper.to_s \
+          && body
```
phstc: Hi @takeyuweb 

So can `body` be empty when `ActiveJob?

takeyuweb:
Hi @phstc 
It is here.
https://github.com/phstc/shoryuken/blob/6a3c94df1a56f4cca8bf900e48c43935b1ac1df7/lib/shoryuken/middleware/server/auto_extend_visibility.rb#L30

phstc:
> It is here.
> https://github.com/phstc/shoryuken/blob/6a3c94df1a56f4cca8bf900e48c43935b1ac1df7/lib/shoryuken/middleware/server/auto_extend_visibility.rb#L30

@takeyuweb hm good catch. I'm wondering if we should pass the `body` in there, otherwise the `worker_name` will return `JobWrapper` instead of the job class.

cc/ @mariokostelac

mariokostelac:
New code won't break, that's for sure. As far as I understand, ActiveJob can't be represented by a blank body.
If we look at ActiveJob::Base serialisation (https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/activejob/lib/active_job/core.rb#L75-L84), we can see that it has to be a valid json, one with `job_class` set.
Was blank message representing valid ActiveJob at some point?

mariokostelac:
I think we can just remove calling `worker_name` for the purpose of logging. Since `worker_name` does not work for batch workers either, it does not bring extra value.
We can just log queue name and message handle (or some sort of identifier).

phstc:
@takeyuweb thanks, I could check it deeper now and I understood. I branched from your branch and created #240 with your fix, and also starting to pass along `body` from the auto extender middleware to the method `worker_name`. Actually, checking Shoryuken code base, only the auto extender and timing middlewares use  `worker_name` - TBH I would love to remove that method, I understand it's handy to see more details in the log, but not everybody use those middlewares. I think it could simplify a bit removing it, but as it's public on `Util`, I can't remove it without introduce a major release. 

Closing in favor of #240.

PR #214, opened 2016-05-25
Make fetcher pause interval configurable

darraghotoole:
The fetcher currently pauses for one second when all processes are busy.  When running in single-threaded mode this effectively means we can only process one job per second.   This PR allows us to configure the pause interval (eg: 0.1 for 100ms) instead of using the hardcoded value.

### DIFF
--- a/lib/shoryuken/cli.rb
+++ b/lib/shoryuken/cli.rb
@@ -119,6 +119,10 @@ module Shoryuken
           opts[:concurrency] = Integer(arg)
         end
 
+        o.on '-i', '--fetcher-pause-interval DECIMAL', 'Interval to pause fetching when all workers are busy' do |arg|
+          opts[:fetcher_pause_interval] = Float(arg)
+        end
+
         o.on '-d', '--daemon', 'Daemonize process' do |arg|
           opts[:daemon] = arg
         end
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -12,7 +12,17 @@ module Shoryuken
 
     def initialize(condvar)
       @count = Shoryuken.options[:concurrency] || 25
-      raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number") unless @count > 0
+      @fetcher_pause_interval = Shoryuken.options[:fetcher_pause_interval] || 1
+
+      unless @count > 0
+        raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number")
+      end
+
+      if @fetcher_pause_interval < 0
+        message = "Fetcher pause interval value #{@fetcher_pause_interval} is invalid, it cannot be negative"
+        raise(ArgumentError, message)
+      end
+
       @queues = Shoryuken.queues.dup.uniq
       @finished = condvar
 
@@ -122,16 +132,15 @@ module Shoryuken
       after(Shoryuken.options[:delay].to_f) { async.restart_queue!(queue) }
     end
 
-
     def dispatch
       return if stopped?
 
       logger.debug { "Ready: #{@ready.size}, Busy: #{@busy.size}, Active Queues: #{unparse_queues(@queues)}" }
 
       if @ready.empty?
-        logger.debug { 'Pausing fetcher, because all processors are busy' }
+        logger.debug { 'Pausing fetcher for #{@fetcher_pause_interval} seconds, because all processors are busy' }
 
-        after(1) { dispatch }
+        after(@fetcher_pause_interval) { dispatch }
 
         return
       end
--- a/spec/shoryuken/manager_spec.rb
+++ b/spec/shoryuken/manager_spec.rb
@@ -11,10 +11,18 @@ RSpec.describe Shoryuken::Manager do
   describe 'Invalid concurrency setting' do
     it 'raises ArgumentError if concurrency is not positive number' do
       Shoryuken.options[:concurrency] = -1
-      expect { Shoryuken::Manager.new(nil) }
-        .to raise_error(ArgumentError, 'Concurrency value -1 is invalid, it needs to be a positive number')
+      expect { Shoryuken::Manager.new(nil) }.
+        to raise_error(ArgumentError, 'Concurrency value -1 is invalid, it needs to be a positive number')
     end
 
+  end
+
+  describe 'Invalid fetch pause interval setting' do
+    it 'raises ArgumentError if fetcher pause interval is not positive number' do
+      Shoryuken.options[:fetcher_pause_interval] = -1
+      expect { Shoryuken::Manager.new(nil) }.
+        to raise_error(ArgumentError, 'Fetcher pause interval value -1 is invalid, it cannot be negative')
+    end
   end
 
   describe 'Auto Scaling' do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -54,6 +54,7 @@ RSpec.configure do |config|
 
     Shoryuken.queues.clear
 
+    Shoryuken.options[:fetcher_pause_interval] = 1
     Shoryuken.options[:concurrency] = 1
     Shoryuken.options[:delay]       = 1
     Shoryuken.options[:timeout]     = 1


lib/shoryuken/manager.rb
```diff
@@ -12,7 +12,16 @@ class Manager
...
     def initialize(condvar)
       @count = Shoryuken.options[:concurrency] || 25
-      raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number") unless @count > 0
+      @dispatch_interval = Shoryuken.options[:dispatch_interval] || 1
+
+      unless @count > 0
+        raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number")
+      end
+
+      unless @dispatch_interval > 0
```
mstewart: I think a value of `0` should be allowed here; it's valid to `dispatch(0)` on celluloid if you want to configure a tight loop.

lib/shoryuken/cli.rb
```diff
@@ -119,6 +119,10 @@ def parse_cli_args(argv)
           opts[:concurrency] = Integer(arg)
         end
 
+        o.on '-i', '--dispatch-interval DECIMAL', 'Dispatch interval' do |arg|
```
bbaja42: Could you add a note what does this value represent, and what will change depending on this value?
Readme.md sounds like a good place

For instance, in case jobs are taking less than a second, let's say 200ms without this change; shoryuken will spend 800ms doing nothing; which ends up in doing just 20% of possible throughput.
On other side, this results in more calls to SQS, in case user is price sensitive

mstewart: Is `dispatch-interval` the right name for this parameter?  There are a few calls to dispatch in the codebase, only one of which is governed by this param.

It'd be nice to find a name which more specifically calls out that it sets the interval used when all processor threads are busy.  E.g. `--busy-wait-time` or something?

lib/shoryuken/manager.rb
```diff
@@ -12,7 +12,17 @@ class Manager
...
     def initialize(condvar)
       @count = Shoryuken.options[:concurrency] || 25
-      raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number") unless @count > 0
+      @fetcher_pause_interval = Shoryuken.options[:fetcher_pause_interval] || 1
+
+      unless @count > 0
+        raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number")
+      end
+
+      unless @fetcher_pause_interval > 0
```
bbaja42: I think I experimented with this value being set to 0 without any problems.
https://github.com/celluloid/timers/blob/fcb443006ed1770d3ba2ebf3db0e0d2e570c6672/lib/timers/group.rb#L37
So is perhaps instead if @fetcher_pause_interval < 0 is better?

lib/shoryuken/manager.rb
```diff
@@ -131,7 +140,7 @@ def dispatch
       if @ready.empty?
         logger.debug { 'Pausing fetcher, because all processors are busy' }
```
bbaja42: Could you add info in the message for how long is the fetcher being paused?

darraghotoole:
@bbaja42 @mstewart Renamed form 'dispath_interval' to 'fetcher_pause_interval'.  Now accepts 0.

lib/shoryuken/manager.rb
```diff
@@ -12,7 +12,17 @@ class Manager
...
-      raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number") unless @count > 0
+      @fetcher_pause_interval = Shoryuken.options[:fetcher_pause_interval] || 1
+
+      unless @count > 0
+        raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number")
+      end
+
+      if @fetcher_pause_interval < 0
+        message = "Fetcher pause interval value #{@fetcher_pause_interval} is invalid, it cannot be negative"
+        raise(ArgumentError, message)
```
phstc: Unrelated to the PR: I'm wondering if we should move these `ArgumentError` errors to `cli.rb`, consequently failing in there, instead of later in the startup process. WDYT?

phstc:
Looks good @darraghotoole 🍻 

@mariokostelac do you mind double checking it?

lib/shoryuken/manager.rb
```diff
@@ -12,7 +12,17 @@ class Manager
 
     def initialize(condvar)
       @count = Shoryuken.options[:concurrency] || 25
-      raise(ArgumentError, "Concurrency value #{@count} is invalid, it needs to be a positive number") unless @count > 0
+      @fetcher_pause_interval = Shoryuken.options[:fetcher_pause_interval] || 1
```
mariokostelac: are we using the wrong symbol here?
We are using `:dispatch_interval` and `:fetcher_dispatch_interval`.

mariokostelac: we are using two different symbols: 
`:dispatch_interval` and `:fetcher_pause_interval`

mariokostelac:
It seems we can't set that value since we are using two different symbols.

lib/shoryuken/manager.rb
darraghotoole: This was missed during refactor, good catch.

lib/shoryuken/manager.rb
darraghotoole: Makes sense, moving the relevant tests would also be the first tests against `cli.rb`, which seems to be untested at the moment.

lib/shoryuken/cli.rb
```diff
@@ -119,6 +119,10 @@ def parse_cli_args(argv)
           opts[:concurrency] = Integer(arg)
         end
 
+        o.on '-i', '--fetcher-pause-interval DECIMAL', 'Interval to pause fetching when all workers are busy' do |arg|
+          opts[:fetcher_pause_interval] = Float(arg)
```
darraghotoole: @mariokostelac Fixed ^^

mariokostelac:
I've been thinking more about this one. You're solving a problem with optionally decreasing that time, but you're also introducing a problem of manager hitting the Q very often when there are no jobs. Making that value very close to 0 would mean constantly polling the Q.

I feel we can do very similar thing like I did in https://github.com/phstc/shoryuken/pull/215.
We are pretty sure that we can assign a new job to the worker in two situations:
- when we just started/restarted a manager
- when certain process is done with the work and moves from `busy` to `ready` collection, which could be when:
  - job is done
  - job crashed

We already call `dispatch` on start so have to do that same after updating `@busy` collection in `processor_done` and `processor_died`.

I am fine with adding this extra option on boot, but I'd like to make sure we are improving the code with every change. 

Could we add those dispatch calls in the same PR @darraghotoole?

@phstc what do you think about this one?

phstc:
> @phstc are you good with my approach?

@mariokostelac I'm good with it.

> I am fine with adding this extra option on boot, but I'd like to make sure we are improving the code with every change.

Loved :)

mariokostelac:
Fix I've been talking about is merged in https://github.com/phstc/shoryuken/commit/f4640d97950c1783a061195855d93994725ed64a.

lib/shoryuken/manager.rb
mariokostelac: Can you move that at least to the `initialize` method? That's where it's set.

lib/shoryuken/manager.rb
```diff
@@ -122,16 +132,15 @@ def pause_queue!(queue)
...
 
-
     def dispatch
       return if stopped?
 
       logger.debug { "Ready: #{@ready.size}, Busy: #{@busy.size}, Active Queues: #{unparse_queues(@queues)}" }
 
       if @ready.empty?
-        logger.debug { 'Pausing fetcher, because all processors are busy' }
+        logger.debug { 'Pausing fetcher for #{@fetcher_pause_interval} seconds, because all processors are busy' }
```
eugeneius: Single-quoted strings are not interpolated, so the log will contain the literal text:

```
Pausing fetcher for #{@fetcher_pause_interval} seconds, because all processors are busy
```

You probably want to use double quotes here.

eugeneius:
This seems conceptually similar to the `delay` option: https://github.com/phstc/shoryuken/wiki/Shoryuken-options#delay

It's a bit inconsistent that `delay` can be set in the config file but not from the command line, and now this option will be settable from the command line but not in the config file.

mariokostelac:
@eugeneius Yes, I do not see any useful difference right now. Since we've fixed the original issue on better way, I am inclined to avoid introducing extra complexity and close this PR.
@darraghotoole @mstewart @bbaja42 thanks for raising the issue and providing a solution for that. If you think that I am wrong with closing this one, please reopen.

PR #219, opened 2016-06-06
Daemonize before loading environment

eugeneius:
Daemonizing forks the process, which kills all other running threads. If your Rails app starts threads during initialization, they will currently be killed almost immediately.

Sidekiq daemonizes before loading Rails:
https://github.com/mperham/sidekiq/blob/v4.1.2/lib/sidekiq/cli.rb#L42-L43
https://github.com/mperham/sidekiq/blob/v4.1.2/lib/sidekiq/cli.rb#L50

### DIFF
--- a/lib/shoryuken/cli.rb
+++ b/lib/shoryuken/cli.rb
@@ -31,10 +31,11 @@ module Shoryuken
 
       options = parse_cli_args(args)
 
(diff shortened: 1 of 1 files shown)


mariokostelac:
I agree, writing PID and daemonizing are related just to the way we run the process. We should do that as soon as possible.

PR #221, opened 2016-06-07
Fix error when initialize shoryuken daemonize

purintai:
repro
```bundle exec shoryuken -d -L ./log/shoryuken.log```

PR #222, opened 2016-06-07
Lazy load AutoExtendVisibility middleware

mariokostelac:
Since `auto_extend_visibility` needs Celluloid, we need to load it lazily.
Also, `load_celluloid` is moved before loading environment because:
- it's a system dependency
- does not require anything from environment
- **breaks pretty hard, which means that loading celluloid after daemonization and before load_celluloid could break, even if it should not.**


mariokostelac:
Closing in favour of https://github.com/phstc/shoryuken/pull/221.

PR #224, opened 2016-06-08
Fix daemonization broken in #219

mariokostelac:


lib/shoryuken/cli.rb
```diff
@@ -31,13 +31,13 @@ def run(args)
...
-      daemonize
-      write_pid
-
-      load_celluloid
+      daemonize(options)
+      write_pid(options)
 
       EnvironmentLoader.load(options)
 
+      load_celluloid
```
mariokostelac: we have to load celluloid later so we could attach logger properly.

lib/shoryuken/cli.rb
```diff
@@ -64,21 +64,27 @@ def run(args)
...
+      defined?(::Celluloid)
+    end
+
+    def daemonize(options)
+      return unless options[:daemon]
 
-      fail ArgumentError, "You really should set a logfile if you're going to daemonize" unless Shoryuken.options[:logfile]
+      fail ArgumentError, "You really should set a logfile if you're going to daemonize" unless options[:logfile]
+
+      if celluloid_loaded?
```
mariokostelac: Daemonizing should fail if there is celluloid loaded, not loading celluloid, as we had before.

spec/shoryuken/cli_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'spec_helper'
+require 'shoryuken/cli'
+require 'shoryuken/launcher'
+
+RSpec.describe Shoryuken::CLI do
```
phstc: ❤️

mariokostelac:
@phstc you good with this PR? :)

lib/shoryuken/cli.rb
phstc: @mariokostelac this is only to prevent users that manually `require celluloid` before requiring shoryuken?

mariokostelac: yes, or us accidentally breaking it :)

lib/shoryuken/cli.rb
```diff
@@ -64,21 +64,27 @@ def run(args)
     private
 
     def load_celluloid
-      raise "Celluloid cannot be required until here, or it will break Shoryuken's daemonization" if defined?(::Celluloid) && Shoryuken.options[:daemon]
```
mariokostelac: @phstc it's just moved from here :)

phstc:
@mariokostelac looks good to me :shipit:

eugeneius:
Sorry for breaking this 😩 and thanks for fixing it @mariokostelac !

PR #230, opened 2016-06-15
Ouput the backtrace when a processor dies.

barisbalic:
I couldn't see any obvious reason why the backtraces aren't logged so I thought I'd suggest this.  Let me know if this is contrary to some intention you had/choice you made.

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -78,6 +78,7 @@ module Shoryuken
     def processor_died(processor, reason)
       watchdog("Manager#processor_died died") do
         logger.error { "Process died, reason: #{reason}" unless reason.to_s.empty? }
+        reason.backtrace.each { |bt| logger.error { bt } } unless reason.backtrace.nil?
 
         @threads.delete(processor.object_id)
         @busy.delete processor


mariokostelac:
@phstc any obvious reason?

lib/shoryuken/manager.rb
```diff
@@ -78,6 +78,7 @@ def processor_done(queue, processor)
     def processor_died(processor, reason)
       watchdog("Manager#processor_died died") do
         logger.error { "Process died, reason: #{reason}" unless reason.to_s.empty? }
+        reason.backtrace.each { |bt| logger.error { bt } } unless reason.backtrace.nil?
```
phstc: @barisbalic could you give a sample output showing how it looks like? I'm wondering if the backtrace goes all the way to the worker that died or if it will always shows the same backtrace listing something internal only.

WDYT?

```ruby
logger.error { reason.backtrace.to_a.join("\n") }
```

But be careful, I remember getting a lot of `reason = nil` when I was testing `processor_died`. So we may need `if reason and !reason.backtrace.to_a.empty?` cc/ @mariokostelac

barisbalic: I've not encountered `reason = nil` but I haven't checked/tested extensively, `backtrace = nil` was certainly true for some failures.  I would be happy to refactor as you suggested.

It's been so long that I honestly can't remember the specific problem that drove me to suggest this change, but I'll try recreating and provide a sample output when I get some time.  However, I would argue regardless of how far back it goes, something is better than nothing.

phstc:
```ruby
2016-11-02T22:38:50Z 21002 TID-ow6zrttks ERROR: Manager#processor_died died
2016-11-02T22:38:50Z 21002 TID-ow6zrttks ERROR: undefined method `backtrace' for nil:NilClass
```

@barisbalic just got `reason = nil` ^ 

Actually, I'm having a hard time to find the celluloid documentation on `trap_exit` to understand what `reason` is, so for now, I would say it's better to not add it, as it isn't reliable to be present.

phstc:
Thanks @barisbalic. But I think it would be safer to close it for now. See https://github.com/phstc/shoryuken/pull/230/files#r86258727

lib/shoryuken/manager.rb
barisbalic: @phstc sorry, been horribly busy and keep forgetting about this.  It is as you suspected before but I'd still wager it's better to guard against this and output a backtrace whenever you can.  It's certainly helped me in the last ~5 months, but if you're dead set against it, I'll defer to your judgement.

PR #236, opened 2016-08-13
Refactor fetcher, polling strategy and manager

mariokostelac:
Simplifying manager, fetcher and polling strategy in order to make whole system easier to understand, test and more performant
- Fetcher became synchronous and quite simple. It became totally unaware of the manager so we finally lost that bidirectional dependency.
- Manager became far slimmer. It lost responsibility of deciding on next queue (moved to polling strategy) and calls fetcher and assign method synchronously for the sake of easier testing.
- Polling strategy it quite simple and can be replaced with something else. WeightedRoundRobin actually implements the old implicit strategy.

I've done some testing locally for next setup
- one 1, two workers
- workers just re-enq the job back to the q
- run OS X 10.11.6, MacBook Pro (Retina, 13-inch, Early 2015)
- run workers for 5 minutes

For concurrency set to 1, throughput increased ~50%.
For concurrency set to 10, throughput increased ~30%.
For concurrency set to 20, throughput increased ~20%.

It's important to note that increase becomes higher if jobs gets faster (shorter).

WDYT @phstc?

### DIFF
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -1,15 +1,8 @@
   module Shoryuken
     class Fetcher
-      include Celluloid
       include Util
 
       FETCH_LIMIT = 10
-
-      def initialize(manager, polling_strategy)
-        @manager = manager
-        @polling_strategy = polling_strategy
-        @delay = Shoryuken.options[:delay].to_f
-      end
 
       def fetch(queue, available_processors)
         watchdog('Fetcher#fetch died') do
@@ -18,36 +11,19 @@
           logger.debug { "Looking for new messages in '#{queue}'" }
 
           begin
-            batch = Shoryuken.worker_registry.batch_receive_messages?(queue.name)
-            limit = batch ? FETCH_LIMIT : available_processors
+            limit = available_processors > FETCH_LIMIT ? FETCH_LIMIT : available_processors
 
             sqs_msgs = Array(receive_messages(queue, limit))
             logger.info { "Found #{sqs_msgs.size} messages for '#{queue}'" }
 
-            if batch
-              @manager.async.assign(queue.name, patch_sqs_msgs!(sqs_msgs))
-            else
-              sqs_msgs.each { |sqs_msg| @manager.async.assign(queue.name, sqs_msg) }
-            end
-
-            @polling_strategy.messages_found(queue, sqs_msgs.size)
-
             logger.debug { "Fetcher for '#{queue}' completed in #{elapsed(started_at)} ms" }
+
+            sqs_msgs
           rescue => ex
             logger.error { "Error fetching message: #{ex}" }
             logger.error { ex.backtrace.first }
           end
-
-          @manager.async.dispatch
         end
-      end
-
-      def next_queue(*args)
-        @polling_strategy.next_queue(*args)
-      end
-
-      def active_queues(*args)
-        @polling_strategy.active_queues(*args)
       end
 
       private
@@ -64,16 +40,6 @@
         options.merge!(queue.options)
 
         Shoryuken::Client.queues(queue.name).receive_messages options
-      end
-
-      def patch_sqs_msgs!(sqs_msgs)
-        sqs_msgs.instance_eval do
-          def message_id
-            "batch-with-#{size}-messages"
-          end
-        end
-
-        sqs_msgs
       end
     end
   end
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -15,13 +15,12 @@
 
             sqs_msgs = Array(receive_messages(queue, limit))
             logger.info { "Found #{sqs_msgs.size} messages for '#{queue}'" }
-
             logger.debug { "Fetcher for '#{queue}' completed in #{elapsed(started_at)} ms" }
-
             sqs_msgs
           rescue => ex
             logger.error { "Error fetching message: #{ex}" }
             logger.error { ex.backtrace.first }
+            []
           end
         end
       end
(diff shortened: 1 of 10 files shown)


lib/shoryuken/fetcher.rb
```diff
@@ -1,73 +1,44 @@
...
+          limit = available_processors > FETCH_LIMIT ? FETCH_LIMIT : available_processors
 
+          sqs_msgs = Array(receive_messages(queue, limit))
+          logger.info { "Found #{sqs_msgs.size} messages for '#{queue}'" }
           logger.debug { "Fetcher for '#{queue}' completed in #{elapsed(started_at)} ms" }
+          sqs_msgs
         rescue => ex
           logger.error { "Error fetching message: #{ex}" }
           logger.error { ex.backtrace.first }
+          []
```
phstc: @mariokostelac not sure if we need `rescue => ex` where, as we are using [`watchdog`](https://github.com/phstc/shoryuken/blob/bf65b6e90bd6648279c89eeaafeb3f21cbbfd1c1/lib/shoryuken/util.rb#L3) wrapping the method body.

lib/shoryuken/fetcher.rb
```diff
@@ -1,73 +1,44 @@
...
 
-    def patch_sqs_msgs!(sqs_msgs)
-      sqs_msgs.instance_eval do
-        def message_id
-          "batch-with-#{size}-messages"
-        end
-      end
+    def receive_messages(queue, limit)
+      # AWS limits the batch size by 10
+      limit = limit > FETCH_LIMIT ? FETCH_LIMIT : limit
```
phstc: @mariokostelac we are doing this same `FETCH_LIMIT` checking in the `fetch` method too, I believe we should do it only on the `receive_messages` method, as it's the one responsible for receiving messages.

lib/shoryuken/manager.rb
```diff
@@ -154,61 +128,41 @@ def real_thread(proxy_id, thr)
...
-          dispatch
-        end
+        processor.async.process(queue, sqs_msg)
       end
     end
 
-    def current_queue_weight(queue)
-      queue_weight(@queues, queue)
+    def dispatch_batch(queue)
+      batch = fetcher.fetch(queue, BATCH_LIMIT)
```
phstc: @mariokostelac I believe we could use something like `Fetcher::FETCH_LIMIT` instead of `BATCH_LIMIT` as it should be the same, in case of batch we want the maximum possible to fetch anyway == fetch limit, and the fetch is the one that knows the what's the maximum.

phstc: To reduce the dependency with constants, we could also have specific methods `fetch_max` and `fetch_limited` or something like that, but I don't see problems with `Fetcher::FETCH_LIMIT` as `Manager` already knows `Fetcher`.

lib/shoryuken/manager.rb
```diff
@@ -241,5 +195,15 @@ def hard_shutdown_in(delay)
         end
       end
     end
+
+    def patch_batch!(sqs_msgs)
+      sqs_msgs.instance_eval do
+        def message_id
+          "batch-with-#{size}-messages"
+        end
```
phstc: @mariokostelac why is this needed?

phstc:
Hey @mariokostelac it's looking super good. I did an initial PR code review, but I'm going to checkout it and test it locally, so I can understand it better.

lib/shoryuken/fetcher.rb
mariokostelac: We should use one or another, not both, that's for sure. I do like having error handling here because SQS does return 500 now and then and it'd be nice to return empty array / retry fetch in that case.

lib/shoryuken/manager.rb
mariokostelac: It was logical separation, but I am good with referencing fetch limit. Other approach would be having `BATCH_LIMIT = Fetcher::FETCH_LIMIT`

lib/shoryuken/manager.rb
mariokostelac: i have to check it out, it was part of fetcher before and i just moved it here. I'd like to get rid of that if possible. Will check if we need this (it think we did need it for logging).

lib/shoryuken/manager.rb
phstc: @mariokostelac I wouldn't introduce `BATCH_LIMIT`, because we don't have a such thing. In this case we want: give the maximum you can. Just to make it simpler for future maintenance.

lib/shoryuken/fetcher.rb
phstc: @mariokostelac hm nvm just saw that the [current implementation](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/fetcher.rb#L51-L54)  does the same.

phstc:
@mariokostelac I could perform an end-to-end test today. It seems to be working fine, but `ctrl-c` does not seem to be working, it seems it isn't catching the shutdown signal, besides that and two feedbacks I left to you, I believe it's ready to go.

mariokostelac:
@phstc ok, I will apply changes based on your comments and take a look at Ctrl-C issue. It worked for me (that's how I stopped it), but could be I broke something.

phstc:
@mariokostelac if `Ctrl-C` (`SIGINT`) works fine for you, ping me and I can test it again. I don't see anything related to trapping signals in your PR, might be something on my test, but it's good to double check it.

atyndall:
Has any progress been made on this? The `StrictPriorityPolling` implementation in pull #106 would be ideal for our circumstances.

phstc:
@mariokostelac let's merge it :shipit: 🍻 so, we can also merge #263. Could you check the merge the conflicts?

mariokostelac:
@phstc I see I have not write rights anymore. Do you prefer creating another PR for that (from forked repo)?

phstc:
@mariokostelac I think that makes sense.

mariokostelac:
Closing in favor of https://github.com/phstc/shoryuken/pull/236!

Issue #235, opened 2016-08-09
DefaultWorkerRegistry and a dead letter queue


tcostermans:
Atm I don't fully understand how to listen to messages on a DQL.
I configured 2 queues: queue1 and dlq1

After 5 delivery attempt a msg is moved to the DLQ.
This seems to be working fine. I find msgs on the DLQ.

Following code snippets describes my setup:
```
class Worker1Job
  include Shoryuken::Worker
  shoryuken_options 'queue1', auto_delete: true
```

```
class DlqListenerJob
  include Shoryuken::Worker
  shoryuken_options queue: 'dlq1', auto_delete: true
```

Main goal: Log a statement when a message arrives on the DLQ and sent a notification to a chat client.

Shoryuken seems to have a worker registry. (DefaultWorkerRegistry)

`message.message_attributes['shoryuken_class'][:string_value] = Worker1Job`

So, when a msg on the DLQ arrives, Worker1Job is called instead of DlqListenerJob.
This might not be what you want when picking up msgs from a DLQ.

In this particular case I just want to log "Async processing of dbs items with ID x,y,z,... failed"
I'm looking for a mechanism to monitor the DLQ. As a first step I just want to be notified if a msg arrives on the DLQ.
How do I achieve this using shoryuken?

phstc:
@tcostermans did you add your DL in your `shoryuken.yml`?

```yaml
queues:
  - [queue1, 1]
  - [dlq1, 1]
```

tcostermans:
@phstc yes, forget to mention that. But that's how my shoryuken.yml looks like.

My log looks like this when a msg is found on the DLQ:


```
14:08:03 web-worker.1 | 2016-08-09T12:08:03Z 94617 TID-oxvw0mlac INFO: Found 1 messages for 'dlq1'
14:08:03 web-worker.1 | 2016-08-09T12:08:03Z 94617 TID-oxvwjo0x4 INFO: Increasing 'dlq1' weight to 2, max: 2

```
And then following code is hit:
```
14:08:04 web-worker.1 | [12, 21] in /Users/user1/.rbenv/versions/2.1.6/lib/ruby/gems/2.1.0/gems/shoryuken-2.0.2/lib/shoryuken/default_worker_registry.rb
14:08:04 web-worker.1 |    12:       @workers.clear
14:08:04 web-worker.1 |    13:     end
14:08:04 web-worker.1 |    14:
[...]

PR #237, opened 2016-08-14
Added proof of concept support for DLQ. Should work for both Shoryuke…

tcostermans:
Probably a lot of polishing still needed, but this is a first proof of concept.
There are some tests that are still failing. So, don't merge yet!

For ActiveJob I had to use a convention
batch_worker (consumes from the normal queue)
batch_dlq_worker (I assume the there is a dlq_worker class if there a Redrive policy defined)

### DIFF
--- a/lib/shoryuken/default_worker_registry.rb
+++ b/lib/shoryuken/default_worker_registry.rb
@@ -18,8 +18,10 @@ module Shoryuken
         message.message_attributes['shoryuken_class'] &&
         message.message_attributes['shoryuken_class'][:string_value]
 
-      approximate_receive_count = message.attributes['ApproximateReceiveCount'].to_i
-      if approximate_receive_count > 1 && 'ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper' != worker_class
+      # TODO: get rid of this code
+      # it's only here to pick the correct Worker in case of msg redelivery, when AJ is not used.
+      # by clearing the worker_class, the correct worker will be fetched based on the queue name.
+      if !message.is_a?(Array) && message.redelivery? && active_job_woker?(worker_class)
         worker_class = nil
       end
 
@@ -46,6 +48,12 @@ module Shoryuken
 
     def workers(queue)
       [@workers.fetch(queue, [])].flatten
+    end
+
+    private
+
+    def active_job_woker?(worker_class)
+      'ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper' != worker_class
     end
   end
 end
--- a/lib/shoryuken/default_worker_registry.rb
+++ b/lib/shoryuken/default_worker_registry.rb
@@ -18,6 +18,11 @@ module Shoryuken
         message.message_attributes['shoryuken_class'] &&
         message.message_attributes['shoryuken_class'][:string_value]
 
+      approximate_receive_count = message.attributes['ApproximateReceiveCount'].to_i
+      if approximate_receive_count > 1 && 'ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper' != worker_class
+        worker_class = nil
+      end
+
       worker_class = (worker_class.constantize rescue nil) || @workers[queue]
 
       worker_class.new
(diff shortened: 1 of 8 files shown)


lib/shoryuken/extensions/active_job_adapter.rb
```diff
@@ -75,6 +75,16 @@ class JobWrapper #:nodoc:
         shoryuken_options body_parser: :json, auto_delete: true
 
         def perform(sqs_msg, hash)
+          approximate_receive_count = sqs_msg.attributes['ApproximateReceiveCount'].to_i
+          if approximate_receive_count > 1
+            message_queue_name = JSON.parse(sqs_msg.body)['queue_name']
+            redrive_policy_json = JSON.parse(Shoryuken::Client.queues(message_queue_name).redrive_policy)
```
tcostermans: Todo: it's possible a queue doesn't have a redrive_policy. Don't crash in that case.

lib/shoryuken/default_worker_registry.rb
```diff
@@ -18,6 +18,11 @@ def fetch_worker(queue, message)
         message.message_attributes['shoryuken_class'] &&
         message.message_attributes['shoryuken_class'][:string_value]
 
+      approximate_receive_count = message.attributes['ApproximateReceiveCount'].to_i
+      if approximate_receive_count > 1 && 'ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper' != worker_class
```
mariokostelac: Can you try to explain this one?

lib/shoryuken/extensions/active_job_adapter.rb
mariokostelac: It's too expensive to check the redrive policy for a q every single time

tcostermans: It's only fetched for messages that are redelivered (approximate_receive_count > 1 ). But you have a point. Some extra caching can be done.

lib/shoryuken/default_worker_registry.rb
tcostermans: This code needs to disappear, but is was a hacky way add support for Shoryuken workers as wel.

When you are using the worker class implementation, the worker should be fetched based on the queue name. (@workers[queue])
In case of redelivery, you want to ignore the message attributes, and clear the worker_class if it's already set so @workers[queue] is executed.

When using a Shoryuken worker the implementation can be really clean, just @workers[queue].new

The AJ adapter is complicating things and making this method complex.

lib/shoryuken/extensions/active_job_adapter.rb
mariokostelac: I'd rather rethink about approach than introduce caching. Caching is nasty stuff.
I think we can infer things from configuration, rather than live system state.

tcostermans: At startup you don't know all your queues. It's only at runtime you can start to infer this kind of information. Adding a msg to the queue and fetching a msg from the key are the only 2 moments when you have the queue name.

A queue has a redrive policy. Once set, it never changes until the queue is deleted and created again. I tried to update the redrive policy but that is not allowed.
In the updated code the policy is stored in a hash with as key the queue name. 
This is the best approach I could come up with at this point.

I'm really eager to get this issue sorted, because consuming a msg from the DLQ can be very useful. So, if you can give me a few hints on how you would solve it, that would be great.

lib/shoryuken/message.rb
```diff
@@ -66,5 +66,17 @@ def md5_of_message_attributes
     def message_attributes
       data.message_attributes
     end
+
+    def approximate_receive_count
+      data.attributes['ApproximateReceiveCount'].to_i
+    end
+
+    def redelivery?
+      approximate_receive_count > 1
```
phstc: That depends on the redrive policy. it can more than 1.

phstc:
@tcostermans thanks for investigating it. I really appreciate it, and it helped me on [this decision](https://github.com/phstc/shoryuken/issues/235#issuecomment-244393560).

cc/ @mariokostelac

PR #243, opened 2016-09-06
changing log verbosity

Sof31t:
for production environment, we dont need to know how much messages we have received. It pollutes the logs specially when we can collect these informations differently (we format logs in our own way)

### DIFF
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -32,7 +32,7 @@ module Shoryuken
           limit = batch ? FETCH_LIMIT : available_processors
 
           if (sqs_msgs = Array(receive_messages(queue, limit))).any?
-            logger.info { "Found #{sqs_msgs.size} messages for '#{queue}'" }
+            logger.debug { "Found #{sqs_msgs.size} messages for '#{queue}'" }
 
             if batch
               @manager.async.assign(queue, patch_sqs_msgs!(sqs_msgs))


lib/shoryuken/fetcher.rb
```diff
@@ -32,7 +32,7 @@ def fetch(queue, available_processors)
           limit = batch ? FETCH_LIMIT : available_processors
 
           if (sqs_msgs = Array(receive_messages(queue, limit))).any?
-            logger.info { "Found #{sqs_msgs.size} messages for '#{queue}'" }
+            logger.debug { "Found #{sqs_msgs.size} messages for '#{queue}'" }
```
phstc: @mariokostelac WDYT about it? I kind of like this change, but it can mislead people that were used to these messages.

Sof31t: @phstc  @mariokostelac  Is it really useful in production environment ? changing it to debug will not affect development environment, in the other hand, it will help a lot of people who are using paying solutions to manage logs ;-)
Not to mention the fact that we can analyze all the trafic for the last 2 weeks in the Amazon monitoring tab. And if we really want this kind of repetitive logs, it is very easy to add it in the worker.

phstc: @Sof31t I'm in favor of your change. I'm just wondering about people that are used to those logs, they might think something wrong is going on if they stop seeing it after a bump.

Sof31t:
any news guys ? @phstc @mariokostelac

phstc:
@Sof31t added to master! Thanks 🍻

PR #244, opened 2016-09-07
fix README

rubemz:
This fixes the README by removing the "DIR" option from "-r" CLI param.

When I passed a DIR it raised, since it calls `require dir` in ruby code, what doesn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ shoryuken [options]
     -c, --concurrency INT            Processor threads to use
     -d, --daemon                     Daemonize process
     -q, --queue QUEUE[,WEIGHT]...    Queues to process with optional weights
-    -r, --require [PATH|DIR]         Location of the worker
+    -r, --require [PATH]             Location of the worker
     -C, --config PATH                Path to YAML config file
     -R, --rails                      Attempts to load the containing Rails project
     -L, --logfile PATH               Path to writable logfile


README.md
```diff
@@ -168,7 +168,7 @@ shoryuken [options]
     -c, --concurrency INT            Processor threads to use
     -d, --daemon                     Daemonize process
     -q, --queue QUEUE[,WEIGHT]...    Queues to process with optional weights
-    -r, --require [PATH|DIR]         Location of the worker
+    -r, --require [PATH]             Location of the worker
```
phstc: @rubemz seems right! I was wondering if it would be better to change [this](https://github.com/phstc/shoryuken/blob/555953fb5e4c37933335109adc1e23d5197c382b/lib/shoryuken/environment_loader.rb#L156) instead to check `File.directory?` and load all files inside a directory in case of a directory is supplied? WDYT?

rubemz: @phstc yeah I agree it would be good to check if it is a dir and load all files inside. Will create a PR for that.

yusukegoto: 👍  I encountered the same issue.

phstc:
Added to master 🍻

yusukegoto:
So need to revert this merge commit when applying the filetype check 😄

phstc:
@yusukegoto good call, will do 🍻

rubemz:
@phstc ah nice. I'd work on that Today, but glad to see it is already in master

phstc:
@rubemz cool, are you going to work on the `DIR` lookup? 🙏

rubemz:
@phstc I'd work on that but I see @yusukegoto already started working in the DIR lookup https://github.com/yusukegoto/shoryuken/commit/970339600528d846f7c480b93d776b869c6c6313

@yusukegoto are you going to create a PR for that?

yusukegoto:
@phstc No! sorry for giving confusion! I don't have any intention to create this PR!! I've waited for your PR.

yusukegoto:
@rubemz  or can i..?

rubemz:
@yusukegoto sure, got ahead an create one. I'd probably would do very similar to what you already did 😄  so feel free to create the PR

yusukegoto:
@rubemz thanks! I'll create!

PR #248, opened 2016-09-16
Cli worker option

yusukegoto:
fix worker directory option as mentioned at #244


phstc:
@yusukegoto thanks! Added to master 🍻

PR #252, opened 2016-09-23
Add aws configuration in 'Shoryuken.configure_server'

### LINKED ISSUE #249: Can't read logfile option in config file
I can't work fine, for example:
```
rails@085a8fd8cf4a:~/app$ bundle exec shoryuken -R -C config/shoryuken.yml -d
You really should set a logfile if you're going to daemonize
/home/rails/app/vendor/bundle/ruby/2.3.0/gems/shoryuken-2.0.11/lib/shoryuken/cli.rb:80:in `daemonize'
/home/rails/app/vendor/bundle/ruby/2.3.0/gems/shoryuken-2.0.11/lib/shoryuken/cli.rb:34:in `run'
/home/rails/app/vendor/bundle/ruby/2.3.0/gems/shoryuken-2.0.11/bin/shoryuken:6:in `<top (required)>'
/home/rails/app/vendor/bundle/ruby/2.3.0/bin/shoryuken:23:in `load'
/home/rails/app/vendor/bundle/ruby/2.3.0/bin/shoryuken:23:in `<main>'
```

In `shoryuken.yml`, I wrote logfile option:
```yaml
verbose: false
concurrency: 20
delay: 25
pidfile: ./tmp/pids/shoryuken.pid
logfile: ./log/shoryuken.log
```
but shoryuken didn't read logfile option.

I expect this case occur after https://github.com/phstc/shoryuken/pull/219.
`EnvironmentLoader.load` read config file, and initialize logger. And `daemonize` check `Shoryuken.options[:logfile]`. So, if call `daemonize` before `EnvironmentLoader.load`, `daemonize` fail.

Current shoryuken source code, 
https://github.com/phstc/shoryuken/blob/v2.0.11/lib/shoryuken/cli.rb#L32-L39
call `daemonize` before `EnvironmentLoader.load`, so didn't read logifle option in config file.

At the same time, I understand https://github.com/phstc/shoryuken/pull/219 is required.

In sidekiq, read config file options and initialize logger, before `daemonize`.
https://github.com/mperham/sidekiq/blob/v4.1.2/lib/sidekiq/cli.rb#L39-L43

I thought that call initialize logger before `daemonize` in the same manner as sidekiq, but shoryuken have to load Rails before read config file: https://github.com/phstc/shoryuken/pull/197
therefore I can't fix.

What do you think?

h3poteto:
I add aws configuration in `Shoryuken.configure_server`, like sidekiq: 
https://github.com/mperham/sidekiq/blob/master/lib/sidekiq.rb#L75-L83

Now, I have to call `EnvironmentLoader.load` in initializer of my Rails application to setup aws configuration. For example:
```ruby
EnvironmentLoader.load(config_file: "config/shoryuken.yml")
```

I want to divide `EnvironmentLoader.load` method, because I try to fix https://github.com/phstc/shoryuken/issues/249.

To fix #249, shoryuken have to read `shoryuken.yml` before `daemonize`, and then load Rails.
So, I want to divide `EnvironmentLoader.load`.

But, now `EnvironmentLoader.load` is used in initializer of user's Rails application, therefore I propose alternate method to configure aws in initializer. That is `Shoryuken.configure_server`.

In sidekiq, user can configure redis information in initializer.
```ruby
Sidekiq.configure_server do |config|
  config.redis = { url: "redis://localhost:6379" }
end

Sidekiq.configure_client do |config|
 config.redis = { url: "redis://localhost:6379" }
end
```

I aim the same way as sidekiq.
```ruby
  config.aws = { :sqs_endpoint => '...', :access_key_id: '...', :secret_access_key: '...', region: '...' }
end

  config.aws = { :sqs_endpoint => '...', :access_key_id: '...', :secret_access_key: '...', region: '...' }
end
```

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -6,6 +6,7 @@ require 'shoryuken/version'
 require 'shoryuken/core_ext'
 require 'shoryuken/util'
 require 'shoryuken/logging'
+require 'shoryuken/aws_config'
 require 'shoryuken/environment_loader'
 require 'shoryuken/queue'
 require 'shoryuken/message'
@@ -72,6 +73,12 @@ module Shoryuken
       @@active_job_queue_name_prefixing = prefixing
     end
 
+    ##
+    # Configuration for Shoryuken server, use like:
+    #
+    #   Shoryuken.configure_server do |config|
+    #     config.aws = { :sqs_endpoint => '...', :access_key_id: '...', :secret_access_key: '...', region: '...' }
+    #   end
     def configure_server
       yield self if server?
     end
@@ -82,8 +89,14 @@ module Shoryuken
       @server_chain
     end
 
+    ##
+    # Configuration for Shoryuken client, use like:
+    #
+    #   Shoryuken.configure_client do |config|
+    #     config.aws = { :sqs_endpoint => '...', :access_key_id: '...', :secret_access_key: '...', region: '...' }
+    #   end
     def configure_client
-      yield self
+      yield self unless server?
     end
 
     def client_middleware
@@ -116,6 +129,10 @@ module Shoryuken
 
     def on_stop(&block)
       @stop_callback = block
+    end
+
+    def aws=(hash)
+      @aws = Shoryuken::AwsConfig.setup(hash)
     end
 
     # Register a block to run at a point in the Shoryuken lifecycle.
--- a/lib/shoryuken/aws_config.rb
+++ b/lib/shoryuken/aws_config.rb
@@ -6,8 +6,8 @@ module Shoryuken
         @options ||= {}
       end
 
-      def set_options(hash)
-        Shoryuken::AwsConfig.options.merge!(hash)
+      def options=(hash)
+        @options = hash
       end
 
       def setup(hash)
@@ -15,7 +15,7 @@ module Shoryuken
         # when not explicit supplied
         return if hash.empty?
 
-        set_options(hash)
+        self.options = hash
 
         shoryuken_keys = %w(
           account_id
--- a/lib/shoryuken/aws_config.rb
+++ b/lib/shoryuken/aws_config.rb
@@ -1,0 +1,49 @@
+module Shoryuken
+  class AwsConfig
+
+    class << self
+      def options
+        @options ||= {}
+      end
+
+      def set_options(hash)
+        Shoryuken::AwsConfig.options.merge!(hash)
+      end
+
+      def setup(hash)
+        # aws-sdk tries to load the credentials from the ENV variables: AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
+        # when not explicit supplied
+        return if hash.empty?
+
+        set_options(hash)
+
+        shoryuken_keys = %w(
+          account_id
+          sns_endpoint
+          sqs_endpoint
+          receive_message
+        ).map(&:to_sym)
+
+        aws_options = hash.reject do |k, v|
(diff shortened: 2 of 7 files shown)


phstc:
Hi @h3poteto

Not sure if I could follow the change completely. Is it basically extracting the `Aws.config` setup? How will it fix #249?

h3poteto:
@phstc 
Thank you for your reply.

I create sample pull request in my repository: https://github.com/h3poteto/shoryuken/pull/2
I will divide `EnvironmentLoader.load` to fix #249, and intercept current `EnvironmentLoader.load`. 

But, now developers use `EnvironmentLoader.load`, so I prepare alternate load method which is this pull request.

lib/shoryuken.rb
```diff
@@ -118,6 +131,10 @@ def on_stop(&block)
       @stop_callback = block
     end
 
+    def aws=(hash)
+      @aws = Shoryuken::AwsConfig.setup(hash)
+    end
```
phstc: @h3poteto is `@aws` being used?

agustin:
Hi there! what do you think of this @phstc? looks like something that you would like to merge eventually? just wondering because I was looking for a good way to decouple the general AWS config of my application from the one that Shoryuken should use. 

I have 3 different users with different permissions in my app, one is for general purposes like store stuff in S3. Other is handling keys with the KMS, and was expecting to be able to have the last one to be in charge of the queues and jobs/workers, but the way is right now seems that if I passed the `[:aws]` key in the `shoryuken.yml` config file this will [override](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L74) the `Aws.config` right? or I'm maybe missed something here 😟 . 

And also on the `Shoryuken::Client` Class there is no way to pass over a different configuration to the SQS or SNS clients. I feel that, as I said before, there is something that I'm missing here.

Let me know if this is not clear at all, and also if you rather have this conversation on an issue or on other way, I wrote it in here because this PR looks like something that will help on this same topic. Maybe I'm wrong, don't hesitate to let me know!

Thanks a lot for all the hard work on all this! 🍺

phstc:
hi @agustin 

The change in this PR is still using `Aws.config` which configures the `Aws` globally. If I understood correct, you would like to configure the SQS client in Shoryuken using a specific creds and region and for other parts of your app to use a different creds, right?

For example, using this:

```ruby
sqs = Aws::SQS::Client.new(
  region: region_name,
  credentials: credentials,
  # ...
)
```

Instead of globally setting it `Aws.config`, is that right?

agustin:
exactly! @phstc! maybe if the [SNS](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/client.rb#L11) and the SQS method on the `Shoryuken::Client` accepts to pass the client directly as an optional argument and if not fallback to the actual way of instance it, maybe that could work. WDYT?

If you are ok with that maybe I can open a PR with that feature, let me know!
Thankss

phstc:
@agustin actually, I think we can continue what @h3poteto started in this PR and add `AwsConfig.sqs`and `AwsConfig.sns` instead of configuring `Aws.config` directly. WDYT?

> If you are ok with that maybe I can open a PR with that feature, let me know!

Please do!

phstc:
@h3poteto great work on this, thanks! Added to master 🍻 

@agustin ^

lib/shoryuken.rb
h3poteto: Sorry,  `@aws` is not used!

h3poteto:
thank you!

PR #259, opened 2016-10-21
Add ability to pass SQS and SNS sdks options directly in config

atyndall:
Adds a `sqs:` and `sns:` section to the config (other key names suggestions welcome) that will pass their options through to the `Aws::SQS::Client` and `Aws::SNS::Client` initializers respectively when created.

Includes tests and updated docs.

Also removes the below passage because as far as I can tell this isn't true as `aws_client_options` doesn't look at `Shoryuken.options[:aws]` to create the `options` key?

> and you can expect any other key defined in that block to be passed on untouched to ```Aws::SQS::Client#initialize```

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,10 @@ aws:
     attribute_names:
       - ApproximateReceiveCount
       - SentTimestamp
+sqs:
+  log_level: debug
+sns:
+  log_level: info
 concurrency: 25  # The number of allocated threads to process messages. Default 25
 delay: 25        # The delay in seconds to pause a queue when it's empty. Default 0
 queues:
@@ -127,7 +131,7 @@ queues:
   - [low_priority, 1]
 ```
 
-The ```aws``` section is used to configure both the Aws objects used by Shoryuken internally, and also to set up some Shoryuken-specific config. The Shoryuken-specific keys are listed below, and you can expect any other key defined in that block to be passed on untouched to ```Aws::SQS::Client#initialize```:
+The ```aws``` section is used to configure both the Aws objects used by Shoryuken internally, and also to set up some Shoryuken-specific config. The Shoryuken-specific keys are listed below:
 
 - ```account_id``` is used when generating SNS ARNs
 - ```sns_endpoint``` can be used to explicitly override the SNS endpoint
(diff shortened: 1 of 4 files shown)


lib/shoryuken/client.rb
```diff
@@ -9,15 +9,15 @@ def queues(name)
...
+        @sns ||= Aws::SNS::Client.new(aws_client_options(:sns))
       end
 
       def sns_arn
         @sns_arn ||= SnsArn
       end
 
       def sqs
-        @sqs ||= Aws::SQS::Client.new(aws_client_options(:sqs_endpoint))
+        @sqs ||= Aws::SQS::Client.new(aws_client_options(:sqs))
```
phstc: @atyndall this is a break change for people using `sqs_endpoint`, that would require a major release version.

atyndall: @phstc the below changes to `aws_client_options` preserve the old `sqs_endpoint` behaviour as well as introducing my new behaviour, there is no breaking change.

Unless `aws_client_options` is accessed outside this class? It shouldn't be, as it's private.

README.md
```diff
@@ -119,6 +119,10 @@ aws:
     attribute_names:
       - ApproximateReceiveCount
       - SentTimestamp
+sqs:
+  log_level: debug
+sns:
+  log_level: info
```
atyndall: The [`log_level`](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/Core/Configuration.html#log_level-instance_method) configuration option configures the log level that all AWS SDK API call messages log at.

The default it logs at is `info`. This pollutes our `info.log` file (which logs `info` and above) with every single SQS request that is made, which is very nosy and makes the logs very cluttered. By configuring it to `debug` log level, the SQS SDK logging is no longer output to `info.log`, and instead goes to our specific `debug.log`.

phstc: @atyndall the default [`logger`](https://docs.aws.amazon.com/AWSRubySDK/latest/AWS/Core/Configuration.html#logger-instance_method) is `nil`, so even if you change the level to info, it won't output `info.log`. If you are seeing the skd log in a file, it's probably because you setting it in somewhere else `AWS.config(logger: ...)`.

I might be missing something in here, but to have the sdk configured to log to a file or `stdout`, we also need to configure `logger: ...`, otherwise it won't output.

atyndall: Sorry, I wasn't completely clear. This is on a system where I have configured the ActiveJob integration with the initializer listed here; https://github.com/phstc/shoryuken/wiki/Rails-Integration-Active-Job

e.g.
```
end
```

`info.log` and `debug.log` are rails logs, this is what they look like pre and post the change; https://gist.github.com/atyndall/b44127df03a4717a76cc6de50a5262f8

As you can see, SQS is very noisy.

lib/shoryuken/client.rb
phstc: @atyndall something that's confusing me is `debug.log` and `info.log`? What's the `logfile` you are passing to Shoryuken? 

By default the AWS logger is set to `nil`, that confuses me, how is Aws logging into `debug.log` and `info.log`.

atyndall: @phstc I am passing no logfile to Shoryuken. `debug.log` and `info.log` are logs produced by Ruby on Rails. I assume Shoryuken is tied into their logging with this code, which I mention in our other conversion thread;

```
end
```

atyndall: Perhaps the AWS logging is a complex interaction of the gems used in my project, I am unsure, but I don't have an obvious answer.

pedros007:
See #353 if `Aws::SQS::Client` is logging at a `log_level` you didn't expect.

Issue #142, opened 2015-10-12
Celluloid 0.17


farski:
Are there any known issues updating the Celluloid dependency to `~> 0.17`?

mrsweaters:
@farski The only one I see is that shoryuken forces us to use `0.16.0` as celluloid `0.16.1` was pulled.

farski:
@mrsweaters Yeah, I'm wondering if there are any code-related issues with updating that dependency to `0.17`

phstc:
@farski @mrsweaters @vemv I couldn't test Shoryuken with `~> 0.17` yet, could you guys test it? I'm super ok with upgrading it in case it's working. Is there a specific change in Celulloid that you guys are looking for?

vemv:
I've been using my fork of shoryuken (which uses ~> 0.17) since I opened #149, that is 2 weeks ago.

Works absolutely fine! (my app is not in the wild though... only in staging)

The thing I'm looking for is compatibility with sidekiq.

farski:
@phstc Compatibility with other gems is the main reason I want it also.

phstc:
Could you guys submit a PR? :heart:

vemv:
Alright, lock acquired!

vemv:
Done!

azisaka:
Hi, what about this warning https://github.com/celluloid/celluloid/wiki/DEPRECATION-WARNING ?
Will you not upgrade the require statements?

phstc:
Hi @azisaka 

Good catch, I can check that, but I can't give you an ETA. Meanwhile, if you can submit a PR, I will be more than happy to review it 🍻

PR #260, opened 2016-10-22
Fix celluloid warning

phstc:
Fixes https://github.com/phstc/shoryuken/issues/142#issuecomment-254362716

### DIFF
--- a/lib/shoryuken/cli.rb
+++ b/lib/shoryuken/cli.rb
@@ -64,7 +64,7 @@ module Shoryuken
     private
 
     def load_celluloid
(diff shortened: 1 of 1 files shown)


lib/shoryuken/cli.rb
```diff
@@ -64,7 +64,7 @@ def run(args)
     private
 
     def load_celluloid
-      require 'celluloid/autostart'
+      require 'celluloid/current'
```
phstc: @azisaka ^ no more warnings 🍻

PR #262, opened 2016-10-24
Add AwsConfig.sns and AwsConfig.sqs, and avoid the override of Aws.config

agustin:
### Summary:

Previously (before the **PR #252**) when we're passing `:aws` config on the `shoryuken.yml` options the `Shoryuken::EnvironmentLoader` class where doing all the initialize and overriding the `Aws.config`. So the **SNS** and **SQS** clients where using that config to initialize the clients, this was a problem for those applications that want/need to have different AWS credentials running different services
### Proposal:

Following @phstc comments on the [pull request](https://github.com/phstc/shoryuken/pull/252#issuecomment-255617953) and taking advantage of the recent changes made by @h3poteto I added the `AwsConfig.sns` and `AwsConfig.sqs` that the `Shoryuken::Client` class will call. And remove the override of `Aws.config` and instead the `credentials` are passed directly to the clients.

Hey @phstc let me know what you think about it, I was unsure about your feelings on instance variables or if you have any other style preference or prefer another approach, please feel free to let me know! all comments are welcome 😃

### DIFF
--- a/lib/shoryuken/aws_config.rb
+++ b/lib/shoryuken/aws_config.rb
@@ -22,25 +22,42 @@ module Shoryuken
           receive_message
         ).map(&:to_sym)
 
-        aws_options = hash.reject do |k, _|
+        @aws_options = hash.reject do |k, _|
           shoryuken_keys.include?(k)
         end
 
         # assume credentials based authentication
         credentials = Aws::Credentials.new(
-          aws_options.delete(:access_key_id),
-          aws_options.delete(:secret_access_key)
+          @aws_options.delete(:access_key_id),
+          @aws_options.delete(:secret_access_key)
         )
 
         # but only if the configuration options have valid values
-        aws_options = aws_options.merge(credentials: credentials) if credentials.set?
+        @aws_options.merge!(credentials: credentials) if credentials.set?
 
         if (callback = Shoryuken.aws_initialization_callback)
           Shoryuken.logger.info { 'Calling Shoryuken.on_aws_initialization block' }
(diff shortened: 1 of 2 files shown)


agustin:
Hi @phstc just wondering if you had the time to look into it? or if you have any comment?

Thanks a lot!

lib/shoryuken/aws_config.rb
```diff
@@ -22,25 +22,46 @@ def setup(hash)
...
+        @aws_options ||= {}
+      end
+
+      def aws_client_options(service_endpoint_key)
+        environment_endpoint = ENV["AWS_#{service_endpoint_key.to_s.upcase}"]
+        explicit_endpoint = options[service_endpoint_key] || environment_endpoint
+        endpoint = {}.tap do |hash|
+          hash[:endpoint] = explicit_endpoint unless explicit_endpoint.to_s.empty?
+        end
+        aws_options.merge(endpoint)
```
phstc: @agustin WDYT about `@aws_options.to_h.merge(endpoint)`, so we can remove the `def aws_options`?

phstc:
@agustin sure thing, looks good! Just made a minor comment, other than that, it's ready for :shipit:

lib/shoryuken/aws_config.rb
cavi21: sure thing! on it

agustin:
I just commit that change and rebase it @phstc anything else please let me know! thanks!

phstc:
@agustin thanks! Added to master 🍻

PR #267, opened 2016-11-01
Terminate visibility extender in ensure block, Fixes #266

### LINKED ISSUE #266: Memory leak in auto_visibility_timeout
The ```@visibility_extender``` instance variable is always nil due to which a new ```MessageVisibilityExtender``` object is created for every message processed. This object is never garbage collected due to which the threads/memory of the worker grows with the number of messages processed.

https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/middleware/server/auto_extend_visibility.rb#L48

We didn't face this issue when not using auto_visibility_timeout, adding ```@visibility_extender.terminate``` as the last line in ```auto_visibility_timer``` method fixes the issue.

rkr090:
ref https://github.com/phstc/shoryuken/issues/266

### DIFF
--- a/lib/shoryuken/middleware/server/auto_extend_visibility.rb
+++ b/lib/shoryuken/middleware/server/auto_extend_visibility.rb
@@ -11,7 +11,10 @@ module Shoryuken
           begin
             yield
           ensure
-            timer.cancel if timer
+            if timer
+              timer.cancel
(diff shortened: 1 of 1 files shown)


phstc:
Thanks, added to master 🍻

PR #268, opened 2016-11-02
log reason celluloid process died

zldavis:


phstc:
Thanks, added to master 🍻

PR #269, opened 2016-11-09
Add deprecation warning in initialize_aws

h3poteto:
I add deprecation warning in `EnvironmentLoader#initialize_aws`.

In other times, I added aws configuration in `Shoryuken.configure_server`:  https://github.com/phstc/shoryuken/pull/252
Now, Shoryuken has two methods to configure aws.
The first way is the method described above which is use `Shoryuken.configure_server` in initializer.
The second way is the earlier method which is use `shoryuken.yml`.

I think it is intricate that there are two methods to configure aws.

So, I will remove the earlier method which is use `shoryuken.yml`.
Before remove that, I add deprecation warning in `EnvironmentLoader#initialize_aws`.

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -49,7 +49,11 @@ module Shoryuken
       YAML.load(ERB.new(IO.read(path)).result).deep_symbolize_keys
     end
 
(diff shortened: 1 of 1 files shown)


phstc:
Thanks @h3poteto. Added to master 🍻

PR #273, opened 2016-11-24
Adding FIFO Support

### LINKED ISSUE #272: FIFO Support
I'm using Shoryuken with a FIFO queue, and the AWS SDK is generating the following error when attempting to queue a job.
`Aws::SQS::Errors::MissingParameter (The request must contain the parameter MessageGroupId.)`

Research reveals that this is a required message attribute when working with FIFO queue. [As per MessageGroupID section in this link](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html).

I'll send a quick and dirty PR to generate a random one.

richseviora:
Fixes #272 

Adds FIFO queue type support to Shoryuken.

Changes:

* `Shoryuken::Queue` will now check if the queue is a FIFO queue or not. 
* If it is, messages will have a `MessageGroupId` attached. 
* If the queue is FIFO and does not have content based deduplication enabled, a UUID is assigned as the `MessageDeduplicationId`.
* If the queue is FIFO and the consumer submits a message with `DelaySeconds`, an error will be raised as the SDK would've raised an error any way.

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -41,7 +41,7 @@ module Shoryuken
     # Returns whether this queue is a FIFO queue or not.
     # @return [TrueClass, FalseClass]
     def is_fifo?
-      queue_attributes.attributes[FIFO_ATTRIBUTE] == 'true'
+      @__is_fifo ||= queue_attributes.attributes[FIFO_ATTRIBUTE] == 'true'
     end
 
     # Returns whether this queue has content based deduplication enabled or not.
@@ -57,9 +57,12 @@ module Shoryuken
     MESSAGE_GROUP_ID = 'ShoryukenMessage'
     VISIBILITY_TIMEOUT_ATTR = 'VisibilityTimeout'
 
+
     # @return [Aws::SQS::Types::GetQueueAttributesResult]
     def queue_attributes
-      client.get_queue_attributes(queue_url: url, attribute_names: [FIFO_ATTRIBUTE, CONTENT_DEDUP_ATTRIBUTE, VISIBILITY_TIMEOUT_ATTR])
+      # Note: Retrieving all queue attributes as requesting `FifoQueue` on non-FIFO queue raises error.
+      # See issue: https://github.com/aws/aws-sdk-ruby/issues/1350
+      client.get_queue_attributes(queue_url: url, attribute_names: ['All'])
     end
 
     # Returns sanitized messages, raising ArgumentError if any of the message is invalid.
@@ -111,16 +114,24 @@ module Shoryuken
       elsif !body.is_a?(String)
         fail ArgumentError, "The message body must be a String and you passed a #{body.class}"
       end
-      if is_fifo? && options[:delay_seconds].is_a?(Fixnum)
+      validate_fifo_message! options
+      options
+    end
+
+    # Validates a FIFO message with the queue configuration.
+    # @param [Hash] options - Message hash.
+    # @raise [ArgumentError] raises ArgumentError if the message configuration is incompatible with the queue configuration.
+    def validate_fifo_message!(options)
+      return unless is_fifo?
+      if options[:delay_seconds].is_a?(Fixnum)
         fail ArgumentError, 'FIFO queues do not accept DelaySeconds arguments.'
       end
-      if is_fifo? && options[:message_group_id].nil?
+      if options[:message_group_id].nil?
         fail ArgumentError, 'This queue is FIFO and no message_group_id was provided.'
       end
-      if is_fifo? && !has_content_deduplication? && options[:message_deduplication_id].nil?
-        fail ArgumentError, 'This queue is FIFO without ContentBasedDeduplication enabled, and no MessageDeduplicationId was supplied'
+      if !has_content_deduplication? && options[:message_deduplication_id].nil?
+        fail ArgumentError, 'This queue is FIFO without ContentBasedDeduplication enabled, and no MessageDeduplicationId was supplied.'
       end
-      options
     end
   end
 end
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -67,10 +67,10 @@ module Shoryuken
       options = case
                 when options.is_a?(Array)
                   { entries: options.map.with_index do |m, index|
-                    { id: index.to_s }.merge(m.is_a?(Hash) ? m : { message_body: m }).tap(&method(:add_fifo_attributes))
+                    { id: index.to_s }.merge(m.is_a?(Hash) ? m : { message_body: m }).tap(&method(:add_fifo_attributes!))
                   end }
                 when options.is_a?(Hash)
-                  options[:entries].each(&:add_fifo_attributes!)
+                  options[:entries].each(&method(:add_fifo_attributes!))
                   options
                 end
       validate_messages!(options)
@@ -83,6 +83,8 @@ module Shoryuken
 
       message_hash[:message_group_id] = MESSAGE_GROUP_ID
       message_hash[:message_deduplication_id] = SecureRandom.uuid unless has_content_deduplication?
+
+      message_hash
     end
 
     def sanitize_message!(options)
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -3,7 +3,7 @@ module Shoryuken
     attr_accessor :name, :client, :url
 
     def initialize(client, name)
-      self.name   = name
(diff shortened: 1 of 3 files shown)


lib/shoryuken/queue.rb
```diff
@@ -37,38 +37,68 @@ def send_messages(options)
...
-
       validate_messages!(options)
-
       options
     end
 
+    # Modifies the supplied hash and adds the required FIFO message attributes based on the queue configuration.
+    def add_fifo_attributes!(message_hash)
+      message_hash[:message_group_id]         = MESSAGE_GROUP_ID if is_fifo?
+      message_hash[:message_deduplication_id] = SecureRandom.uuid if is_fifo? && !has_content_deduplication?
```
phstc: @richseviora WDYT:

```ruby
def add_fifo_attributes!(message_hash)
   return unless is_fifo?
   
   message_hash[:message_group_id]         = MESSAGE_GROUP_ID
   message_hash[:message_deduplication_id] = SecureRandom.uuid unless has_content_deduplication?
end
```

lib/shoryuken/queue.rb
```diff
@@ -14,8 +14,8 @@ def initialize(client, name)
 
     def visibility_timeout
       client.get_queue_attributes(
-        queue_url: url,
-        attribute_names: ['VisibilityTimeout']
+          queue_url:       url,
+          attribute_names: ['VisibilityTimeout']
       ).attributes['VisibilityTimeout'].to_i
```
phstc: @richseviora could you keep the indentation, we use 2 spaces.

lib/shoryuken/queue.rb
```diff
@@ -37,38 +37,68 @@ def send_messages(options)
...
       options = case
-                when options.is_a?(String)
-                  # send_message('message')
-                  { message_body: options }
-                when options.is_a?(Hash)
-                  options
+                  when options.is_a?(String)
+                    # send_message('message')
+                    {message_body: options}
+                  when options.is_a?(Hash)
```
phstc: Better not changing the styling.

lib/shoryuken/queue.rb
```diff
@@ -37,38 +37,68 @@ def send_messages(options)
...
 
     private
 
+    FIFO_ATTRIBUTE          = 'FifoQueue'
+    CONTENT_DEDUP_ATTRIBUTE = 'ContentBasedDeduplication'
+    MESSAGE_GROUP_ID        = 'ShoryukenMessage'
+
+    # @return [Aws::SQS::Types::GetQueueAttributesResult]
+    def fifo_attributes
+      @fifo_attr ||= client.get_queue_attributes(queue_url: url, attribute_names: [FIFO_ATTRIBUTE, CONTENT_DEDUP_ATTRIBUTE])
```
phstc: I was wondering if we could do:

```ruby
def attributes
  @_attributes || client.get_queue_attributes(queue_url: url, attribute_names: [FIFO_ATTRIBUTE, CONTENT_DEDUP_ATTRIBUTE, 'VisibilityTimeout'])
end
```

So we could share the `attributes` with `visibility_timeout` and `fifo`.

lib/shoryuken/queue.rb
richseviora: Definitely cleaner. Will amend accordingly.

lib/shoryuken/queue.rb
richseviora: I like it! 

As I'm thinking through this, I'm starting to wonder if we'd need to worry about changes to queues while the workers are active. I know `FifoQueue` is readonly, but I believe `VisibilityTimeout` and `ContentBasedDeduplication` can be changed.

lib/shoryuken/queue.rb
richseviora: Will do. I don't see a code style statement in the readme so I'll add that in.

phstc: There's no specific code style, it's more like just to keep the style it currently uses.

richseviora: So then we'll make the current style the code style. I think it's beneficial to include so new contributors can write it correctly going forward, and it saves you time :)

lib/shoryuken/queue.rb
```diff
@@ -41,23 +38,55 @@ def receive_messages(options)
...
+
+    # Returns whether this queue has content based deduplication enabled or not.
+    # @return [TrueClass, FalseClass]
+    def has_content_deduplication?
+      queue_attributes.attributes[CONTENT_DEDUP_ATTRIBUTE] == 'true'
+    end
+
     private
 
+    FIFO_ATTRIBUTE = 'FifoQueue'
```
phstc: @richseviora I'm getting this:

```shell
     Aws::SQS::Errors::InvalidAttributeName:
       Unknown Attribute FifoQueue.
```

while trying to run `SPEC_ALL=true rspec` which performs an integration test end-to-end. I think that's because the idea of the generic `queue_attributes`, it's also requesting `FifoQueue` for non-FIFO queues.

I think we should use:

```ruby
# See https://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html#get_queue_attributes-instance_method
client.get_queue_attributes(queue_url: url, attribute_names: ['All'])
```

So it would return `FifoQueue` when available. WDYT?

lib/shoryuken/queue.rb
```diff
@@ -41,23 +38,55 @@ def receive_messages(options)
         map { |m| Message.new(client, self, m) }
     end
 
+    # Returns whether this queue is a FIFO queue or not.
+    # @return [TrueClass, FalseClass]
+    def is_fifo?
+      queue_attributes.attributes[FIFO_ATTRIBUTE] == 'true'
```
phstc: @richseviora I got your point that we shouldn't memoize the generic `queue_attributes` as the visibility timeout can be changed after the queue creation, that was a good call. But I think we could memoize `is_fifo?`, so we don't need to perform a request for every message we send.

```ruby
def fifo?
  @_is_fifo ||= queue_attributes.attributes[FIFO_ATTRIBUTE] == 'true'
end
```

WDYT?

lib/shoryuken/queue.rb
```diff
@@ -83,7 +111,15 @@ def validate_message!(options)
...
+      if is_fifo? && options[:delay_seconds].is_a?(Fixnum)
+        fail ArgumentError, 'FIFO queues do not accept DelaySeconds arguments.'
+      end
+      if is_fifo? && options[:message_group_id].nil?
+        fail ArgumentError, 'This queue is FIFO and no message_group_id was provided.'
+      end
+      if is_fifo? && !has_content_deduplication? && options[:message_deduplication_id].nil?
+        fail ArgumentError, 'This queue is FIFO without ContentBasedDeduplication enabled, and no MessageDeduplicationId was supplied'
+      end
       options
```
phstc: @richseviora WDYT:

```ruby
    def validate_message!(options)
      body = options[:message_body]
      if body.is_a?(Hash)
        options[:message_body] = JSON.dump(body)
      elsif !body.is_a?(String)
        fail ArgumentError, "The message body must be a String and you passed a #{body.class}"
      end

      validate_fifo_message!(options)

      options
    end

    def validate_fifo_message!(options)
      return unless fifo?

      if options[:delay_seconds].is_a?(Fixnum)
        fail ArgumentError, 'FIFO queues do not accept DelaySeconds arguments.'
      end

      if options[:message_group_id].nil?
        fail ArgumentError, 'This queue is FIFO and no message_group_id was provided.'
      end

      if options[:message_deduplication_id].nil? && !has_content_deduplication?
        fail ArgumentError, 'This queue is FIFO without ContentBasedDeduplication enabled, and no MessageDeduplicationId was supplied'
      end
    end
```

This `if options[:message_deduplication_id].nil? && !has_content_deduplication?` condition change is because `has_content_deduplication?` can perform a request, so it's better to check `options[:message_deduplication_id]` first, because we may not need to check the other which's more expensive.

lib/shoryuken/queue.rb
richseviora: Wow. I'm surprised that SQS would have a queue attribute that would generate an error if the value was false. I've opened an issue with their SDK. I'll amend the code as suggested.

lib/shoryuken/queue.rb
richseviora: Agreed! Will amend accordingly.

phstc:
Hi @richseviora 

I created a follow up PR #276 (branching from your branch, so all your commits are in there), with some ideas, would be able to review it?

PR #276, opened 2016-12-03
Continuation of #273

### LINKED ISSUE #272: FIFO Support

phstc:
Fixes #272

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -1,22 +1,21 @@
 module Shoryuken
   class Queue
+    FIFO_ATTR               = 'FifoQueue'
+    MESSAGE_GROUP_ID        = 'ShoryukenMessage'
+    VISIBILITY_TIMEOUT_ATTR = 'VisibilityTimeout'
+
     attr_accessor :name, :client, :url
 
     def initialize(client, name)
       self.name   = name
       self.client = client
-      begin
-        self.url = client.get_queue_url(queue_name: name).queue_url
-      rescue Aws::SQS::Errors::NonExistentQueue => e
-        raise e, "The specified queue '#{name}' does not exist"
-      end
+      self.url    = client.get_queue_url(queue_name: name).queue_url
+    rescue Aws::SQS::Errors::NonExistentQueue => e
+      raise e, "The specified queue '#{name}' does not exist."
     end
 
     def visibility_timeout
-      client.get_queue_attributes(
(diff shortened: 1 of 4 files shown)


lib/shoryuken/queue.rb
```diff
@@ -58,33 +68,48 @@ def sanitize_messages!(options)
       options
     end
 
+    def add_fifo_attributes!(options)
+      return unless fifo?
+
+      options[:message_group_id]         ||= MESSAGE_GROUP_ID
+      options[:message_deduplication_id] ||= Digest::SHA256.digest(options[:message_body])
```
phstc: @richseviora so in here I think we can set `message_group_id` and `message_deduplication_id` in case they are not supplied. I want to avoid checking `has_content_deduplication?`, otherwise we would need do a request to check if it's set or not per message, like N+1. WDYT?

lib/shoryuken/queue.rb
```diff
@@ -58,33 +68,48 @@ def sanitize_messages!(options)
...
+
       options
     end
+
+    def validate_fifo_message!(options)
+      return unless fifo?
+
+      if options[:delay_seconds]
+        fail ArgumentError, 'FIFO queues do not accept DelaySeconds arguments.'
+      end
```
phstc: @richseviora I removed other validation, because [this code](https://github.com/phstc/shoryuken/pull/276/files#r90748950) will prevent that to be `nil` anyways. WDYT?

phstc:
@richseviora WDYT https://github.com/phstc/shoryuken/wiki/FIFO-Queues?

lib/shoryuken/queue.rb
richseviora: Good point and agreed :)

lib/shoryuken/queue.rb
richseviora: Agreed. The code was something I put in place really for dev purposes, it's definitely redundant now.

PR #277, opened 2016-12-03
Port from #263

phstc:
Port from #263

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -134,8 +134,7 @@ module Shoryuken
       if @ready.empty?
         logger.debug { 'Pausing fetcher, because all processors are busy' }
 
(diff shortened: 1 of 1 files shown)


lib/shoryuken/manager.rb
```diff
@@ -154,6 +153,13 @@ def real_thread(proxy_id, thr)
 
     private
 
+    def dispatch_later
+      @_dispatch_timer ||= after(1) do
+        @_dispatch_timer = nil
+        dispatch
+      end
+    end
```
phstc: Hey @atyndall

I extracted your patch for the `dispatch` in this isolated PR.

phstc: @atyndall BTW I was able to cherry-pick your commit, to give you the credit for the patch 🍻

PR #279, opened 2016-12-05
Fix aws deprecation warning message

phstc:
Actualy `aws` is always set here https://github.com/phstc/shoryuken/blob/master/lib/shoryuken.rb#L29
so it's never nil, consequently always showing the warning message no
matter if `aws` is set or not

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -53,7 +53,9 @@ module Shoryuken
     # https://github.com/phstc/shoryuken/blob/a81637d577b36c5cf245882733ea91a335b6602f/lib/shoryuken.rb#L82
     # Please delete this method afert next release (v2.0.12 or later)
     def initialize_aws
-      Shoryuken.logger.warn { "[DEPRECATION] aws in shoryuken.yml is deprecated. Please use configure_server and configure_client in your initializer"} unless Shoryuken.options[:aws].nil?
+      unless Shoryuken.options[:aws].to_h.empty?
+        Shoryuken.logger.warn { '[DEPRECATION] aws in shoryuken.yml is deprecated. Please use configure_server and configure_client in your initializer' }
+      end
       Shoryuken::AwsConfig.setup(Shoryuken.options[:aws])
     end
 


lib/shoryuken/environment_loader.rb
```diff
@@ -53,7 +53,9 @@ def config_file_options
     # https://github.com/phstc/shoryuken/blob/a81637d577b36c5cf245882733ea91a335b6602f/lib/shoryuken.rb#L82
     # Please delete this method afert next release (v2.0.12 or later)
     def initialize_aws
-      Shoryuken.logger.warn { "[DEPRECATION] aws in shoryuken.yml is deprecated. Please use configure_server and configure_client in your initializer"} unless Shoryuken.options[:aws].nil?
+      unless Shoryuken.options[:aws].to_h.empty?
+        Shoryuken.logger.warn { '[DEPRECATION] aws in shoryuken.yml is deprecated. Please use configure_server and configure_client in your initializer' }
+      end
```
h3poteto: Sorry, I missed this https://github.com/phstc/shoryuken/blob/master/lib/shoryuken.rb#L29 .
Thank you for your fix.

I recommend use `configure_server` and `configure_client` in initializer. So, I add this `DEPRECATION` warning. 

Now, Shoryuken has multipe configure methods of aws.
First, worker side has two ways:

- Write aws information in `shoryuken.yml`
- Use `configure_client` block in initializer

Second, producer side has a few ways:

- Set environment variables, `AWS_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY`
- Use `~/.aws/credentials`
- Use Instance profile
- Use `configure_server` block in initializer

I thought this situation was complicated.
And I thought that setting aws in separate place is not good.

But, It may not be bad for multiple methods to exist, if we can control. 
What do you think?

phstc: @h3poteto @suzan2go I have a project which I don't use Rails, I use a standalone version of Shoryuken. So for this project, I would need to create a Ruby class, and make sure it's the first file required in the `-r` to set the AWS config. Also, there are still some people that during the deploy process (capistrano), copy a `shoryuken.yml` to the project's dir, for this people they would need to copy a class instead, which feels bad or change the way they initialize AWS. 

> Because if the configuration of aws is in shoryuken.yml, the initialize process of shoryuken becomes too complicated.

Unless I'm missing something I think the only thing it requires is this:
https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L55-L58

WDYT?

h3poteto: I answer this:
> I have a project which I don't use Rails, I use a standalone version of Shoryuken. So for this project, I would need to create a Ruby class, and make sure it's the first file required in the -r to set the AWS config. 

I thought that if people use pure Ruby class to Shoryuken worker (not Rails), please write `configure_client` in top of the Ruby file. 

```ruby
  config.aws = {
    sqs_endpoint: ENV["SQS_ENDPOINT"],
    secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"],
    access_key_id: ENV["AWS_ACCESS_KEY_ID"],
    region: ENV["AWS_REGION"]
  }
end

class MyWorker
  include Shoryuken::Worker

  shoryuken_options queue: 'default', auto_delete: true
  def perform(sqs_msg, body)
    puts body
  end
end
```

This way is same as Sidekiq, it presented here: https://www.youtube.com/watch?v=bfPb1zD91Rg&index=1&list=PLjeHh2LSCFrWGT5uVjUuFKAcrcj5kSai1

PR #283, opened 2016-12-09
Add explanation about configure_client and configure_server

h3poteto:
`shoryuken v2.1.0` can use `configure_client` and `configure_server` to setup aws.
So, I add explanation that in README.

I will update the wiki later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,6 @@ And setup ```aws``` options to use ```configure_client``` in `config/initializer
 ```ruby
 Shoryuken.configure_client do |config|
   config.aws = {
-    sqs_endpoint:      ...,         # or ENV["AWS_SQS_ENDPOINT"]
-    sns_endpoint:      ...,         # or ENV["AWS_SNS_ENDPOINT"]
     secret_access_key: ...,         # or ENV["AWS_SECRET_ACCESS_KEY"]
     access_key_id:     ...,         # or ENV["AWS_ACCESS_KEY_ID"]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,24 +118,85 @@ queues:
...
-- ```receive_message``` can be used to define the options passed to the http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html#receive_message-instance_method
+```ruby
+Shoryuken.configure_client do |config|
+  config.aws = {
+    sqs_endpoint:      ...,         # or ENV["AWS_SQS_ENDPOINT"]
+    sns_endpoint:      ...,         # or ENV["AWS_SNS_ENDPOINT"]
+    secret_access_key: ...,         # or ENV["AWS_SECRET_ACCESS_KEY"]
+    access_key_id:     ...,         # or ENV["AWS_ACCESS_KEY_ID"]
+    region:            "us-east-1", # or ENV["AWS_REGION"]
+    receive_message: {              # See http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html#receive_message-instance_method
```
phstc: @h3poteto thanks for the update! 🍻 

I think we could remove the `sqs_endpoint` and `sns_endpoint` from the example above, it can mislead people making the believe it's required, usually they just want to copy&paste and try to fill that information, I guess most don't know about these endpoints, WDYT?

phstc:
@h3poteto thanks! Added to master 🍻 

> I will update the wiki later.

❤️

PR #286, opened 2016-12-13
Add example endless workers

mariokostelac:
Very often I need some worker to perform basic E2E test. This two workers should represent a worker that's doing IO (sleep) and workers that are CPU intensive (burning metal).
This should help us to:
1) be confident that at least basic e2e flow works when we release a new version
2) write integration tests if we want them (I'd say we do :))
3) do harness testing by providing different environment variables and running shoryuken for long time.
4) help iterating on issues like https://github.com/phstc/shoryuken/pull/287 quite fast

### DIFF
--- a/test_workers/endless_interruptive_worker.rb
+++ b/test_workers/endless_interruptive_worker.rb
@@ -1,0 +1,41 @@
+class EndlessInterruptiveWorker
+  include Shoryuken::Worker
+
+  # Usage:
+  # QUEUE="super-q"
+  # MAX_EXECUTION_TIME=2000 QUEUE=$QUEUE \
+  # bundle exec ./bin/shoryuken -r ./examples/endless_uninterruptive_worker.rb -q $QUEUE -c 8
+
+  class << self
+    def queue
+      ENV['QUEUE'] || 'default'
+    end
+
+    def max_execution_time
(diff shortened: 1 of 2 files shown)


phstc:
> write integration tests if we want them (I'd say we do :))

We have [one](https://github.com/phstc/shoryuken/blob/master/spec/integration/launcher_spec.rb). You need `SPEC_ALL=true rspec` to run it.

How do you test these workers, you start them, then `rake console` and enqueue messages?

mariokostelac:
Yes, they're self-serving (creating messages for themselves). I mostly use them to be able to run shoryuken as long-running process, test that manager dispatching works properly (or at least does not crash) etc. I do not feel strong about having them in the repo, but would simplify manual long-running testing.

phstc:
@mariokostelac got it! I don't mind having them in the repo if it's part of your testing workflow. But as they are under `examples/` would mind to add a comment on top of them explaining it's for testing purposes? Just to avoid misleading people, otherwise they may think it's an example of implementing workers, WDYT? We could also have a specific folder for that, but I don't have strong opinions on that, up to you.

mariokostelac:
Maybe just putting them under `test_workers/`?

PR #284, opened 2016-12-13
Refactor fetcher, polling strategy and manager

mariokostelac:
Continuation of https://github.com/phstc/shoryuken/pull/236

PR #288, opened 2016-12-13
Strict priority polling w/ backmerge

atyndall:
@phstc This is the strict priority polling implementation from #263 with master backmerged, it didn't make it when you closed #263 in favour of #284.

### DIFF
--- a/lib/shoryuken/polling.rb
+++ b/lib/shoryuken/polling.rb
@@ -21,7 +21,11 @@ module Shoryuken
       alias_method :eql?, :==
 
       def to_s
-        options.empty? ? name : super
+        if options.empty?
+          name
+        else
+          "#<QueueConfiguration #{name} options=#{options.inspect}>"
+        end
       end
     end
 
--- a/lib/shoryuken/polling.rb
+++ b/lib/shoryuken/polling.rb
@@ -132,17 +132,17 @@ module Shoryuken
 
       def initialize(queues)
         # Priority ordering of the queues, highest priority first
-        @initial_order = queues
+        @queues = queues
           .group_by { |q| q }
           .sort_by { |_, qs| -qs.count }
           .map(&:first)
 
-        # Stores the queue ordering with the next queue as first element
-        @queue_order = @initial_order.dup
-
         # Pause status of the queues, default to past time (unpaused)
         @paused_until = queues
           .each_with_object(Hash.new) { |queue, h| h[queue] = Time.at(0) }
+
+        # Start queues at 0
+        reset_next_queue
       end
 
       def next_queue
@@ -152,44 +152,48 @@ module Shoryuken
 
       def messages_found(queue, messages_found)
         if messages_found == 0
-          # If no messages are found, we pause a given queue
           pause(queue)
         else
-          # Reset the queue order to the initial ordering
-          @queue_order = @initial_order.dup
+          reset_next_queue
         end
       end
 
       def active_queues
-        @paused_until
-          .reject { |_, unpause_at| unpause_at > Time.now }
-          .map { |queue, _| [queue, @initial_order.reverse.find_index(queue) + 1] }
+        @queues
+          .reverse
+          .map.with_index(1)
+          .reject { |q, _| queue_paused?(q) }
+          .reverse
       end
 
       private
 
       def next_active_queue
-        now = Time.now
+        reset_next_queue if queues_unpaused_since?
 
-        # Return nil if all queues are paused to prevent infinite loop
-        return nil if @paused_until.values.all? { |t| t > now }
-
-        # If any queues have unpaused since the last time we checked, reset the ordering
-        if @last_check && @paused_until.values.any? { |t| t > @last_check && t <= now }
-          @queue_order = @initial_order.dup
+        size = @queues.length
+        size.times do
+          queue = @queues[@next_queue_index]
+          @next_queue_index = (@next_queue_index + 1) % size
+          return queue unless queue_paused?(queue)
         end
 
-        @last_check = now
+        return nil
+      end
 
-        # `rotate!` through the queue list until we find an unpaused queue
-        begin
-          next_queue = @queue_order.first
-          unpause_at = @paused_until[next_queue]
+      def queues_unpaused_since?
+        last = @last_unpause_check
+        now = @last_unpause_check = Time.now
 
-          @queue_order.rotate!
-        end while unpause_at > now
+        last && @paused_until.values.any? { |t| t > last && t <= now }
+      end
 
-        next_queue
+      def reset_next_queue
+        @next_queue_index = 0
+      end
+
+      def queue_paused?(queue)
+        @paused_until[queue] > Time.now
       end
 
       def pause(queue)
--- a/lib/shoryuken/polling.rb
+++ b/lib/shoryuken/polling.rb
@@ -134,7 +134,7 @@ module Shoryuken
         # Priority ordering of the queues, highest priority first
         @initial_order = queues
           .group_by { |q| q }
-          .sort_by { |q, qs| -qs.count }
+          .sort_by { |_, qs| -qs.count }
           .map(&:first)
 
         # Stores the queue ordering with the next queue as first element
--- a/lib/shoryuken/polling.rb
+++ b/lib/shoryuken/polling.rb
@@ -127,89 +127,71 @@ module Shoryuken
     class StrictPriority < BaseStrategy
 
       def initialize(queues)
-        # Mapping of queues to priority values
-        @queue_priorities = queues
-          .each_with_object(Hash.new(0)) { |queue, h| h[queue] += 1 }
-
-        # Priority ordering of the queues
-        @queue_order = @queue_priorities
-          .to_a
-          .sort_by { |queue, priority| -priority }
+        # Priority ordering of the queues, highest priority first
+        @initial_order = queues
+          .group_by { |q| q }
+          .sort_by { |q, qs| -qs.count }
           .map(&:first)
 
-        # Pause status of the queues
-        @queue_status = @queue_order
-          .each_with_object({}) { |queue, h| h[queue] = [true, nil] }
+        # Stores the queue ordering with the next queue as first element
+        @queue_order = @initial_order.dup
 
-        # Most recently used queue
-        @current_queue = nil
+        # Pause status of the queues, default to past time (unpaused)
+        @paused_until = queues
+          .each_with_object(Hash.new) { |queue, h| h[queue] = Time.at(0) }
       end
 
       def next_queue
-        unpause_queues
-        @current_queue = next_active_queue
-        return nil if @current_queue.nil?
-        QueueConfiguration.new(@current_queue, {})
+        next_queue = next_active_queue
+        next_queue.nil? ? nil : QueueConfiguration.new(next_queue, {})
       end
 
       def messages_found(queue, messages_found)
         if messages_found == 0
           # If no messages are found, we pause a given queue
-          pause(queue) 
+          pause(queue)
         else
-          # Reset the current queue when messages found to cause priorities to re-run
-          @current_queue = nil
+          # Reset the queue order to the initial ordering
+          @queue_order = @initial_order.dup
         end
       end
 
       def active_queues
-        @queue_status
-          .select { |_, status| status.first }
-          .map { |queue, _| [queue, @queue_priorities[queue]] }
+        @paused_until
+          .reject { |_, unpause_at| unpause_at > Time.now }
+          .map { |queue, _| [queue, @initial_order.reverse.find_index(queue) + 1] }
       end
 
       private
 
       def next_active_queue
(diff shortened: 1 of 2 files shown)


phstc:
@mariokostelac would you mind reviewing it?

mariokostelac:
@atyndall could you rebase on master first please? We merged my change in the master.

atyndall:
@mariokostelac Should be done now

lib/shoryuken/polling.rb
```diff
@@ -19,11 +19,49 @@ def ==(other)
       end
 
       alias_method :eql?, :==
+
+      def to_s
+        options.empty? ? name : super
```
mariokostelac: Could you explain this?

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,94 @@ def queue_weight(queues, queue)
...
+          .to_a
+          .sort_by { |queue, priority| -priority }
+          .map(&:first)
+
+        # Pause status of the queues
+        @queue_status = @queue_order
+          .each_with_object({}) { |queue, h| h[queue] = [true, nil] }
+
+        # Most recently used queue
+        @current_queue = nil
```
mariokostelac: From what  I understand, that it actually `last_used_queue`. Can we name it like that? It will make more sense for the loop.

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,94 @@ def queue_weight(queues, queue)
...
+          .each_with_object(Hash.new(0)) { |queue, h| h[queue] += 1 }
+
+        # Priority ordering of the queues
+        @queue_order = @queue_priorities
+          .to_a
+          .sort_by { |queue, priority| -priority }
+          .map(&:first)
+
+        # Pause status of the queues
+        @queue_status = @queue_order
```
mariokostelac: I think we can compress queue_status in just one field, something like "activation time". When `Time.now < activation_time`, queue is not active yet. When `Time.now >= activation_time`, queue is active.

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,94 @@ def queue_weight(queues, queue)
...
+    class StrictPriority < BaseStrategy
+
+      def initialize(queues)
+        # Mapping of queues to priority values
+        @queue_priorities = queues
+          .each_with_object(Hash.new(0)) { |queue, h| h[queue] += 1 }
+
+        # Priority ordering of the queues
+        @queue_order = @queue_priorities
+          .to_a
```
mariokostelac: I think we need to simplify this code or add a comment of the datastructure coming  out :).

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,94 @@ def queue_weight(queues, queue)
...
+
+      def next_active_queue
+        return nil unless @queue_order.length > 0
+
+        start = @current_queue.nil? ? 0 : @queue_order.index(@current_queue) + 1
+        i = 0
+
+        # Loop through the queue order from the current queue until we find a
+        # queue that is next in line and is not paused
+        while true
```
mariokostelac: Can we use a for loop? This external bookkeeping of variable i looks ugly.

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,94 @@ def queue_weight(queues, queue)
...
+          queue = @queue_order[(start + i) % @queue_order.length]
+          active, delay = @queue_status[queue]
+          
+          i += 1
+          return queue if active
+          return nil if i >= @queue_order.length # Prevents infinite looping
```
mariokostelac: Also, using For loop gets rid of this line.

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,94 @@ def queue_weight(queues, queue)
...
+        end
+      end
+
+      def pause(queue)
+        return unless delay > 0
+        @queue_status[queue] = [false, Time.now + delay]
+        logger.debug "Paused '#{queue}'"
+      end
+ 
+      def unpause_queues
```
mariokostelac: Once we compress the state, we do not have to `unpause` queues because we know their order, we just have to check if `activation_time > Time.now` or not :).

spec/shoryuken/polling_spec.rb
```diff
@@ -98,3 +98,142 @@
...
+      queues << queue1
+      queues << queue2
+
+      expect(subject.next_queue).to eq(queue1)
+      expect(subject.next_queue).to eq(queue2)
+      expect(subject.next_queue).to eq(queue1)
+      expect(subject.next_queue).to eq(queue2)
+    end
+
+    it 'cycles when declared asc' do
```
mariokostelac: why looks the same as previous one, while priorities are different?

mariokostelac:
@atyndall could you please describe the algorithm you want to implement? That's the only way we can be sure we're on the same page :).

lib/shoryuken/polling.rb
gshutler:
```ruby
@queue_order = queues.group_by(&:itself).sort_by { |q, qs| -qs.count }.map(&:first)
```

May have to do the longhand of `Object#itself` if you need support for Ruby < 2.2.0

lib/shoryuken/polling.rb
gshutler: It's not drop-in but we have similar code in our custom stuff that works like this:

```ruby
queues = @queue_order.dup

while true
  queue = queues.first
  # process queue
  queues.rotate!
end
```

So rather than maintaining `i` or `@current_queue` you `dup` the ordered queues at some point and then `Array#rotate!` them as needed to move to the next queue.

lib/shoryuken/polling.rb
atyndall: :+1:

atyndall: Nice, will use instead

lib/shoryuken/polling.rb
atyndall: Rework has eliminated variable altogether

lib/shoryuken/polling.rb
atyndall: @gshutler Excellent! I didn't know about `Array#rotate!`, I'll rework the code around it.

lib/shoryuken/polling.rb
atyndall: 👍

lib/shoryuken/polling.rb
atyndall: Ah yes, I'll implement this approach.

spec/shoryuken/polling_spec.rb
atyndall: The priorities aren't different, the priorities are the number. They are declaration order independent.

Therefore
```
[shoryuken, 2]
[uppercut,  1]
```

and

```
[uppercut,  1]
[shoryuken, 2]
```

Have the same meaning, because they have the same priority numbers.

atyndall:
I have made significant changes to the implementation taking into account the feedback, thanks!

@mariokostelac Sure, at a high level each queue is assigned a unique priority, describing how critical its work is.

When shoryuken goes to select a new job, it chooses the highest priority unpaused queue to check first.

If that queue contains no work, it will proceed to check the next highest priority unpaused queue, etc. 

When new work is received or when a queue unpauses, the checking order is reset to check the highest priority queue once more.

lib/shoryuken/polling.rb
mariokostelac: Can we make a case where options does exist a little bit nicer? :)

mariokostelac:
@atyndall so that's priority queue, right? As long as there are jobs in a queue with the highest priority, poll that queue. Once it has no jobs, pause it for `delay` period and continue with the rest of queues?

If yes, we can just use https://ruby-doc.org/stdlib-2.3.1/libdoc/set/rdoc/SortedSet.html.

atyndall:
@mariokostelac Yeah that's correct.
I don't really understand what changes you're envisioning with the use of a `SortedSet`

spec/shoryuken/polling_spec.rb
mariokostelac: Oh, did not see that queues are reversed!

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,76 @@ def queue_weight(queues, queue)
...
+          # If no messages are found, we pause a given queue
+          pause(queue)
+        else
+          # Reset the queue order to the initial ordering
+          @queue_order = @initial_order.dup
+        end
+      end
+
+      def active_queues
+        @paused_until
```
mariokostelac: Also, once we have `queue_paused?`, we can use something simple as 
```ruby
@queues.reject{ |q| queue_paused?(q) }
```

mariokostelac: I think it's far easier to understand

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,76 @@ def queue_weight(queues, queue)
...
+          .map { |queue, _| [queue, @initial_order.reverse.find_index(queue) + 1] }
+      end
+
+      private
+
+      def next_active_queue
+        now = Time.now
+
+        # Return nil if all queues are paused to prevent infinite loop
+        return nil if @paused_until.values.all? { |t| t > now }
```
atyndall: Sounds good, I'll implement that approach.

lib/shoryuken/polling.rb
atyndall: I'll add the method

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,76 @@ def queue_weight(queues, queue)
...
+        next_queue.nil? ? nil : QueueConfiguration.new(next_queue, {})
+      end
+
+      def messages_found(queue, messages_found)
+        if messages_found == 0
+          # If no messages are found, we pause a given queue
+          pause(queue)
+        else
+          # Reset the queue order to the initial ordering
+          @queue_order = @initial_order.dup
```
atyndall: Will add such a method

lib/shoryuken/polling.rb
atyndall: Made some changes

lib/shoryuken/polling.rb
```diff
@@ -102,5 +123,76 @@ def queue_weight(queues, queue)
...
+        # If any queues have unpaused since the last time we checked, reset the ordering
+        if @last_check && @paused_until.values.any? { |t| t > @last_check && t <= now }
+          @queue_order = @initial_order.dup
+        end
+
+        @last_check = now
+
+        # `rotate!` through the queue list until we find an unpaused queue
+        begin
+          next_queue = @queue_order.first
```
atyndall: It's getting refactored out now

phstc:
@mariokostelac I think most of people they don't care about which polling strategy they need to use, the only want to make it work. The shoryuken.yml by default (in the samples/README) should be something like this

```yaml
concurrency: 25  # The number of allocated threads to process messages. Default 25
queues:
  - [high_priority, 6]
  - [default, 2]
  - [low_priority, 1]
```

I would remove the `delay` from the sample, I agree it's confusing. Most people, only want to install, configure the queues and run it. Maybe we could also omit `concurrency` from the samples/README, not sure if it's needed. Most people only need the defaults, these extra options can be documented in the wiki. Including the polling strategy, I'm still seeing it as "edge case".

With that, I wouldn't do any major release, I would just let "choose your polling strategy" something special for now, documented in the wiki. Keeping shoryuken as simple and straightforward as possible, I don't mind about keeping the `delay`, default strategy will be probably the most used one.

WDYT?

Is this change introducing any breaking changes?

atyndall:
@phstc AFAIK there are no breaking changes, merging this in should not have any impact on anyone who doesn't use it.

For your information, this is what our queueing config looks like atm;

```
polling_strategy: !ruby/class 'Shoryuken::Polling::StrictPriority'
timeout: 1140   # Seconds grace to shutdown
concurrency: 1  # The number of allocated threads to process messages. Default 25
delay: 60       # The delay in seconds to pause a queue when it's empty. Default 0
queues:
  - [high_priority,       3]
  - [medium_priority,     2]
  - [low_priority,        1]
```

phstc:
@atyndall nice! `polling_strategy`, `timeout` and `delay` are still optional, right?

atyndall:
@phstc Yep

mariokostelac:
@atyndall @phstc I am fine with having a default polling strategy being set and all associated parameters, but I am not really happy with sharing these parameters between polling strategies. They could have slightly or completely different meanings. That's why I'd like to namespace these parameters in new version :). Also, `delay` is not descriptive at all.

phstc:
> That's why I'd like to namespace these parameters in new version :). Also, delay is not descriptive at all.

@mariokostelac I agree, I just want to make the basic `shoryuken.yml` as trivial as possible. Most of people don't care about these extra settings, as long as we can have a good set of defaults, all is fine 🍻

PR #300, opened 2016-12-30
Ignore USR2 signal

serihiro:
# What will be changed by this PR?
This PR modify signal handling to ignore `USR2`. 

```
2016-12-30T02:00:18Z 33879 TID-oxpjqlcdc INFO: Starting
2016-12-30T02:00:46Z 33879 TID-oxpjbckjw INFO: Got USR2 signal
2016-12-30T02:00:46Z 33879 TID-oxpjbckjw WARN: Received USR2, will do nothing. To execute soft shutdown, please send USR1
```

# What this is for?
When we send `USR2` to shoryuken process, shoryuken execute hard shutdown.
But I think this is implicit and dangerous specification. Other popular daemon programs, e.g. `nginx`, `sidekiq` and `unicorn`, do not execute shutdown by sending `USR2`. So I think it is not common usage of `USR2` to execute hard shutdown.

Perhaps USR 2 was considering other uses ( e.g. reopening log like `sidekiq`).  But for now, no implementation for USR2 exists. So I think that it is safe to keep do nothing.

### DIFF
--- a/lib/shoryuken/cli.rb
+++ b/lib/shoryuken/cli.rb
@@ -205,6 +205,8 @@ module Shoryuken
         queues = launcher.manager.instance_variable_get(:@queues)
 
         logger.info { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{unparse_queues(queues)}" }
+      when 'USR2'
+        logger.warn { "Received #{sig}, will do nothing. To execute soft shutdown, please send USR1" }
       else
         logger.info { "Received #{sig}, will shutdown down" }
 


phstc:
Hi @serihiro 

How are you triggering USR2? I mean, is that part of your flow?

serihiro:
@phstc 
Thank you for your response.

As you said, it is a part of my task that sending kill signal to `Shoryuken` process.
That's are executed in `Capistrano` tasks.

I use `Capistrano` to deploy my application using `Shoryuken`.
So I wrote `Shoryuken` restart task, like the following:

```ruby
namespace :shoryuken do
  task :restart => :environment do
    on roles(:shoryuken) do
      execute :kill, "-s USR1 $(< #{fetch(:shoryuken_pid)}"
      start_shoryuken # this is custom method
    end
  end
end
```

I have other some `Capistrano` tasks on the same rails application. Those use sending kill signal to daemon process(`Sidekiq`, `Unicorn`, and so on) to restart them. 

And, as I wrote in the description of this PR, when other daemon process receives `USR2`, execute shut down. So , I thought this may lead a mistake. 

In short, this PR is only a personal opinion based on my system operation experience,  not necessary feature.
However, I thought that there might be something else that is stuck with the same problem, and I issued this PR.

phstc:
hi @serihiro, thanks for the explanation! 🍻 

@mariokostelac any thoughts on this?

phstc:
@serihiro added to [master](https://github.com/phstc/shoryuken/commit/00a05771a882efbc27e7b03412d8cab4d8c6a5eb) 🍻

serihiro:
Thanks! 🍻

PR #305, opened 2017-01-19
Require Celluloid 17

carlallen:
https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/cli.rb#L72 will fail for celluloid 0.16 as `celluloid/current` does not exist until 0.17


phstc:
Thanks @carlallen! Added to master 🍻

PR #317, opened 2017-02-17
WIP remove AwsConfig

### LINKED ISSUE #293: Remove Aws.config
We should drop any AWS credentials setup from Shoryuken, like this:

```ruby
  config.aws = {
```

In favor of leveraging [`Aws.config`](https://docs.aws.amazon.com/sdkforruby/api/index.html#Configuration). 

aws-sdk is very powerful and supports a few different ways to configure it, we shouldn't duplicate this logic in Shoryuken. This would also "resolve" the problem about people complaining that they've configured the creds in `shoryuken.yml`, but it isn't picked up in client mode (Rails), they would know that they need to configure that by themselves. 

- [x] Drop https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/aws_config.rb
- [x] Update README and Wiki references related to the Aws credentials/config setup. Explaining how to use the `Aws.config`
  - Add a link to [Aws.config](https://docs.aws.amazon.com/sdkforruby/api/index.html#Configuration) for more details
  - Explain that it should be configured using an initializer or ENV variables or IAM roles or any other way described in aws-sdk that's more convenient for them. As long as it's properly configured, Shoryuken will use it.
- [x] Remove SNS references in the code. Shoryuken does not use SNS, people must refer to aws-sdk instead

The maximum we can provide is a setter `sqs=` to allow people to set the  [SQS Client](https://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html), in case the have a specific need, such as `sqs_endpoint` or they want to have different credentials other than the ones in `Aws.config`. I'm not totally convinced about this setter yet, but that is the maximum we can provide.

We currently support `receive_message` in the `shoryuken.yml`, but I think people only use it to for attributes, and because of the FIFO queue, we [ask for all by default](https://github.com/phstc/shoryuken/blob/thread-pool/lib/shoryuken/queue.rb#L52).

phstc:
Fix #293


greggroth:
FYI This removed `Shoryuken::Client.sns` which was a breaking change that wasn't clearly mentioned in the Changelog (aside from a comment about removing AWS configuration).  Also, the [wiki is now out of date](https://github.com/phstc/shoryuken/wiki/SNS-Topics).

phstc:
Hi @greggroth 

Good call. I've just updated the wiki.

Regarding to the breaking change, I assumed it was implied by the major release.  But you are right, it should have been more explicit in the CHANGELOG.md.

greggroth:
Thanks for the quick updates!

PR #322, opened 2017-02-28
Allow dynamic adding queues

### LINKED ISSUE #282: Environment Load Order
We use Rails and our config data is loaded via Figaro into the environment.

In the latest version of Shoryuken, the `shoryuken.yml` file gets parsed/loaded before the Rails environment loads and Figaro has a chance to inject the extra options into `ENV`.

I haven't tested this with Dotenv, but I suspect it would have a similar problem if you were using environment variables in a `shoryuken.yml` file.

Can the `Shoryuken::EnvironmentLoader#load_rails` file be safely moved to the top of `Shoryuken::EnvironmentLoader#setup_options`?

phstc:
Fix #282

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -99,7 +99,7 @@ module Shoryuken
     end
 
     def parse_queue(queue, weight = nil)
-      [weight.to_i, 1].max.times { Shoryuken.queues << queue }
+      Shoryuken.add_queue(queue, [weight.to_i, 1])
     end
 
     def parse_queues
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -145,7 +145,7 @@ module Shoryuken
 
       Shoryuken.queues.uniq.each do |queue|
         begin
-          Shoryuken::Client.queues queue
+          Shoryuken::Client.queues(queue)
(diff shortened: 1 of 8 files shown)


lib/shoryuken/environment_loader.rb
```diff
@@ -99,7 +99,7 @@ def prefix_active_job_queue_names
     end
 
     def parse_queue(queue, weight = nil)
-      [weight.to_i, 1].max.times { Shoryuken.queues << queue }
+      Shoryuken.add_queue(queue, [weight.to_i, 1])
```
h3poteto: You have not take the max of `[weight.to_i, 1]` but is it OK?
I think that `add_queue` takes `weight` which is defined numeric.

phstc: @h3poteto nice! thanks for reviewing it. I meant `[weight.to_i, 1].max`.

This code needs a test `rspec`. Not sure how tests are passing, even the integration.

lib/shoryuken/environment_loader.rb
```diff
@@ -99,7 +99,7 @@ def prefix_active_job_queue_names
     end
 
     def parse_queue(queue, weight = nil)
-      Shoryuken.add_queue(queue, [weight.to_i, 1])
+      Shoryuken.add_queue(queue, [weight.to_i, 1].max)
```
phstc: @h3poteto 🍻

PR #329, opened 2017-03-09
Support retry_intervals as a lambda

### LINKED ISSUE #303: retry_intervals - option to support a lambda?
I noticed that the ExponentialBackoffRetry middleware doesn't add any randomness which is common in backoff algorithms to spread out the retries. With the way the code is structured I you'd just need to do something like: 

```ruby
interval = if attempts < retry_intervals.size
               retry_intervals[attempts]
             else
               retry_intervals.last
             end

# Addition: call if callable
interval = interval.call(attempts) if interval.respond_to?(:call)
```

Note this is speculative and nice to have. Users can swap out the existing middleware for another now. I'm happy to implement it and add tests if it's desirable.

phstc:
Fix #303

### DIFF
--- a/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
+++ b/lib/shoryuken/middleware/server/exponential_backoff_retry.rb
@@ -6,16 +6,16 @@ module Shoryuken
 
         def call(worker, queue, sqs_msg, body)
           if sqs_msg.is_a?(Array)
-            logger.warn { "Exponential backoff isn't supported for batch workers"  }
+            logger.warn { "Exponential backoff isn't supported for batch workers" }
             return yield
           end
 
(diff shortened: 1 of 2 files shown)


lib/shoryuken/middleware/server/exponential_backoff_retry.rb
```diff
@@ -34,7 +34,7 @@ def get_interval(retry_intervals, attempts)
             retry_intervals[attempts]
           else
             retry_intervals.last
-          end
+          end.to_i
```
phstc: @felixbuenemann is that what you were asking for in https://github.com/phstc/shoryuken/issues/303#issuecomment-285438722?

tjsingleton:
<3 lgtm!

PR #335, opened 2017-03-15
Fix initializing sqs client

h3poteto:
I try to set custom sqs client in initializer, like this:

```ruby
  config.sqs_client = Aws::SQS::Client.new(
    endpoint: ENV["SQS_ENDPOINT"],
    secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"],
    access_key_id: ENV["AWS_ACCESS_KEY_ID"],
    region: ENV["AWS_REGION"]
  )
end
```

But, `Shoryuken.sqs_client` was not my custom sqs client.

The class variable `@@sqs_client` is not assigned in `Shoryuken.sqs_client=(sqs_client)`.

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -95,7 +95,7 @@ module Shoryuken
     end
 
     def sqs_client=(sqs_client)
-      @@sqs_client
+      @@sqs_client = sqs_client
     end
 
     def sqs_client_receive_message_opts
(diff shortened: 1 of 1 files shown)


phstc:
Oops... _ has no idea what's doing_

Thanks a ton @h3poteto. Added to master 🍻

PR #343, opened 2017-03-23
Add batch by interval for real

diclophis:


phstc:
Hi @diclophis 

Could you give more context/use case on the batch by interval?

diclophis:
@racerpeter could you comment on what these changes do in general... I would like to get this merged back into upstream so we can close out our fork

racerpeter:
@phstc This branch was actually a WIP-- the use case is a scenario where one would want to accept a stream of events from the queue and organize them into time-based batches. Our use case is related to data warehousing. We have a stream of changes from our primary database cluster that is uploaded to a data warehouse that accepts batched uploads. The time slice consumer would facilitate such an approach.

Ultimately we ended up using a different solution, but by all means feel free to use this spike in any way you see fit (or not! :) )

Peter

phstc:
Hi @racerpeter 

Thanks for explaining it 🍻 

Would [FIFO](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html) or [playing with visibility timeout](https://github.com/phstc/shoryuken/wiki/Sending-a-message#delaying-a-message) work for you use case?

PR #345, opened 2017-03-24
Stop queuing up heartbeat threads

### LINKED ISSUE #338: Thread exhaustion
I'm not sure how to describe this in a replicatable way. :(

But we have a Rails and a non-Rails app that we're using v3 of Shoryuken with. 

After processing a number of requests (on a Mac) we are starting to see `Error fetching message: can't create Thread: Resource temporarily unavailable` errors.

Macs only allow a pretty small number of threads per process (I believe its 2048) and it is usually a pretty hard limit.

However, we only have concurrency set to 10 and no pool size configured for the DB.

I'm not sure how we're exhausting threads. :(

phstc:
Fix #338

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -19,8 +19,11 @@ module Shoryuken
       @polling_strategy = polling_strategy
 
       @heartbeat = Concurrent::TimerTask.new(run_now: true,
-                                             execution_interval: HEARTBEAT_INTERVAL,
-                                             timeout_interval: 60) { dispatch }
+                                             execution_interval: HEARTBEAT_INTERVAL) { @pool.post { dispatch } if @dispatching.false? }
+
+      Concurrent::TimerTask.new(execution_interval: 1) do
+        Shoryuken.logger.info "Threads: #{Thread.list.size}"
+      end.execute
 
       @pool = Concurrent::FixedThreadPool.new(@count, max_queue: @count)
     end
@@ -62,14 +65,16 @@ module Shoryuken
       return if @done.true?
       return unless @dispatching.make_true
 
-      return if ready.zero?
-      return unless (queue = @polling_strategy.next_queue)
+      begin
+        return if ready.zero?
+        return unless (queue = @polling_strategy.next_queue)
 
-      logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }
+        logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }
 
-      batched_queue?(queue) ? dispatch_batch(queue) : dispatch_single_messages(queue)
-    ensure
-      @dispatching.make_false
+        batched_queue?(queue) ? dispatch_batch(queue) : dispatch_single_messages(queue)
+      ensure
+        @dispatching.make_false
+      end
     end
 
     def busy
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -3,7 +3,7 @@ module Shoryuken
     include Util
 
     BATCH_LIMIT        = 10
-    HEARTBEAT_INTERVAL = 0.1
+    HEARTBEAT_INTERVAL = 1
 
     def initialize(fetcher, polling_strategy)
       @count = Shoryuken.options.fetch(:concurrency, 25)
@@ -13,19 +13,18 @@ module Shoryuken
       @queues = Shoryuken.queues.dup.uniq
 
       @done = Concurrent::AtomicBoolean.new(false)
-      @dispatching = Concurrent::AtomicBoolean.new(false)
 
       @fetcher = fetcher
       @polling_strategy = polling_strategy
-
-      @heartbeat = Concurrent::TimerTask.new(run_now: true,
-                                             execution_interval: HEARTBEAT_INTERVAL) { @pool.post { dispatch } if @dispatching.false? }
 
       Concurrent::TimerTask.new(execution_interval: 1) do
         Shoryuken.logger.info "Threads: #{Thread.list.size}"
       end.execute
 
       @pool = Concurrent::FixedThreadPool.new(@count, max_queue: @count)
+      @dispatcher_pool = Concurrent::SingleThreadExecutor.new
+
+      @heartbeat = Concurrent::TimerTask.new(run_now: true, execution_interval: HEARTBEAT_INTERVAL) { dispatch }
     end
 
     def start
@@ -47,6 +46,7 @@ module Shoryuken
       logger.info { 'Shutting down workers' }
 
       @heartbeat.kill
+      @dispatcher_pool.kill
 
       if options[:shutdown]
         hard_shutdown_in(options[:timeout])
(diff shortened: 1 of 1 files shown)


lib/shoryuken/manager.rb
```diff
@@ -19,8 +19,11 @@ def initialize(fetcher, polling_strategy)
       @polling_strategy = polling_strategy
 
       @heartbeat = Concurrent::TimerTask.new(run_now: true,
-                                             execution_interval: HEARTBEAT_INTERVAL,
-                                             timeout_interval: 60) { dispatch }
+                                             execution_interval: HEARTBEAT_INTERVAL) { @pool.post { dispatch } if @dispatching.false? }
```
phstc: @waynerobinson I know it isn't the most beautiful solution in the world. But that fix the thread leaking, I still need to play more with that. But I think the "final" fix will be something like that.

lib/shoryuken/manager.rb
```diff
@@ -19,8 +19,11 @@ def initialize(fetcher, polling_strategy)
...
+
+      Concurrent::TimerTask.new(execution_interval: 1) do
+        Shoryuken.logger.info "Threads: #{Thread.list.size}"
+      end.execute
```
phstc: @waynerobinson With concurrency 10: `bundle exec bin/shoryuken -c 10 -q test -r ./test.rb`, the thread size kept consistent at 16.

lib/shoryuken/manager.rb
waynerobinson: That will probably do it. It reuses the same worker pool for dispatching (which I don't really think is problematic) and the `@pool.post { ... }` will return instantly, meaning the timeout never occurs (because of the implementation, you can't actually turn off the timeout code in `TimerTask`). 

And this still lets you use `TimerTask` as a supervisor-like solution for dispatch, ensuring it stays executing, even if there is an error.

phstc: @waynerobinson thanks for reviewing it 🍻 

I've just changed it to use a separate and single thread pool.

I'm now running some performance tests, let's see how it goes 🙏

waynerobinson: Looks good. 

As long as the block finishes faster than the `execution_timeout` then this leak won't occur, even if the currently broken state of `concurrent-ruby`. 

Now if only the team there could be convinced that there is a bug in the first place. 😜

lib/shoryuken/manager.rb
```diff
@@ -13,16 +13,18 @@ def initialize(fetcher, polling_strategy)
...
 
-      @heartbeat = Concurrent::TimerTask.new(run_now: true,
-                                             execution_interval: HEARTBEAT_INTERVAL,
-                                             timeout_interval: 60) { dispatch }
+      Concurrent::TimerTask.new(execution_interval: 1) do
+        Shoryuken.logger.info "Threads: #{Thread.list.size}"
+      end.execute
 
       @pool = Concurrent::FixedThreadPool.new(@count, max_queue: @count)
+      @dispatcher_pool = Concurrent::SingleThreadExecutor.new
```
phstc: > As long as the block finishes faster than the execution_timeout then this leak won't occur, even if the currently broken state of concurrent-ruby.

@waynerobinson now as I'm using this `SingleThreadExecutor`, it's no longer an issue, it supports only one a thread at time and the fallback policy is discard. If we try to `post` while there's a thread running, it will just discard the `post`.

> Now if only the team there could be convinced that there is a bug in the first place. 😜

I will try to reply on that issue as well. I think they should allow to configure a `max_queue` and `fallback_policy` for the `TimerTask`, too bad they don't allow it, and it ends up having a `max_queue: infinity`.

lib/shoryuken/manager.rb
```diff
@@ -3,7 +3,7 @@ class Manager
     include Util
 
     BATCH_LIMIT        = 10
-    HEARTBEAT_INTERVAL = 0.1
+    HEARTBEAT_INTERVAL = 1
```
phstc: @waynerobinson I'm testing `HEARTBEAT_INTERVAL = 1`, as I'm also calling `dispatch` in the `processor_done` callback.

lib/shoryuken/manager.rb
waynerobinson: Except that `TimerTask` is the thing that's leaking threads because of the ones it creates to do the timeout. But I'm sure the execution task occurs faster than the heartbeat period.

Don't really need a `max_queue` in `TimerTask` because the way the class is designed it should be locking around the actual task and only letting one operate at a time. 

But the lack of a correct implementation for the timeout causes thread leaks in the timeout monitor if the task takes longer than the execution interval (it should be the timeout interval at the very least, but it never refers to this… hence the bug) to complete. 

The current design intention of `TimerTask` seems to have it limited to only ever using 2 threads.

lib/shoryuken/manager.rb
waynerobinson: That's probably going to cause issues with empty queues as it will take a least `HEARTBEAT_INTERVAL` before it will request new messages.

Instead of this you could just have `dispatch` run in a loop and rely on the `TimerTask` to restart it if it ever dies.

waynerobinson: Something like this:

```ruby
def dispatch_now
  while true
    if ready.zero?
      return unless (queue = @polling_strategy.next_queue)
  
      logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }

      batched_queue?(queue) ? dispatch_batch(queue) : dispatch_single_messages(queue)
    else
      sleep MINIMUM_WAIT # 0.05 or something to prevent CPU pegging
    end
  end
end
```

waynerobinson: Although instead of `ready.zero?`, is there just some type of latch on the pool to await a free thread so you don't have to do the sleep?

lib/shoryuken/manager.rb
```diff
@@ -58,24 +54,28 @@ def stop(options = {})
...
+        return if ready.zero?
+        return unless (queue = @polling_strategy.next_queue)
 
-      logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }
+        logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }
 
-      batched_queue?(queue) ? dispatch_batch(queue) : dispatch_single_messages(queue)
+        batched_queue?(queue) ? dispatch_batch(queue) : dispatch_single_messages(queue)
+      ensure
+        dispatch_async
```
phstc: > sleep MINIMUM_WAIT # 0.05 or something to prevent CPU pegging

@waynerobinson I need to keep it running longer, but it seems to be working without a `sleep`. As the `@dispatcher_executor` discards `post` when there's a thread running, the request for fetching is kind of doing `sleep`, it should only be like without any sleep if all processors are busy or non queue is available as you pointed out. I will keep monitoring the CPU, if that does not work, we can def add `sleep 0.05`.

waynerobinson: We always use Receive Wait Time of > 0, so it will always block if the network is up anyway. The sleep is really just for worst case situations.

lib/shoryuken/manager.rb
phstc: > Although instead of ready.zero?, is there just some type of latch on the pool to await a free thread so you don't have to do the sleep?

@waynerobinson not I'm aware of 🐼

lib/shoryuken/manager.rb
```diff
@@ -13,22 +12,22 @@ def initialize(fetcher, polling_strategy)
       @queues = Shoryuken.queues.dup.uniq
 
       @done = Concurrent::AtomicBoolean.new(false)
-      @dispatching = Concurrent::AtomicBoolean.new(false)
 
       @fetcher = fetcher
       @polling_strategy = polling_strategy
 
-      @heartbeat = Concurrent::TimerTask.new(run_now: true,
```
phstc: @waynerobinson I'm doing the `loop` in a `ensure` block, and also calling `dispatch_async` when a processor done, just in case. So I don't think we need to the heartbeat anymore.

waynerobinson: Not sure if the `dispatch_async` in `processor_done` does anything given the normal dispatcher re-runs at the end anyway.

But the ensure and re-run should keep the dispatch loop running I think. 👍

waynerobinson:
Great work! 🍾

lib/shoryuken/manager.rb
phstc: @waynerobinson you are right, maybe I was too overcautious. I added it in there just in case. But I'm considering removing it, `ensure` should work.

PR #349, opened 2017-03-30
Revert shoryuken parameter name to 'config' from 'config_file'

fredsadaghiani:
This [commit](432ca4f0fd03adc2efe4e9aee8135dea35f50d3f) changed the parameter name to the shoryuken script from `config` to `config_file`. Among others things, like scripts that manage the starting/stopping of the process, this change breaks the capistrano-shoryuken gem (See https://github.com/joekhoobyar/capistrano-shoryuken/blob/master/lib/capistrano/tasks/shoryuken.cap#L8)

This change reverts the name of the parameter to the original value.

### DIFF
--- a/bin/shoryuken
+++ b/bin/shoryuken
@@ -21,7 +21,7 @@ module Shoryuken
       method_option :daemon,      aliases: '-d', type: :boolean, desc: 'Daemonize process'
       method_option :queues,      aliases: '-q', type: :array,   desc: 'Queues to process with optional weights'
       method_option :require,     aliases: '-r', type: :string,  desc: 'Dir or path of the workers'
-      method_option :config_file, aliases: '-C', type: :string,  desc: 'Path to config file'
+      method_option :config,      aliases: '-C', type: :string,  desc: 'Path to config file'
       method_option :rails,       aliases: '-R', type: :boolean, desc: 'Load Rails'
       method_option :logfile,     aliases: '-L', type: :string,  desc: 'Path to logfile'
       method_option :pidfile,     aliases: '-P', type: :string,  desc: 'Path to pidfile'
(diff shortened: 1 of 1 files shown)


bin/shoryuken
```diff
@@ -21,7 +21,7 @@ module Shoryuken
       method_option :daemon,      aliases: '-d', type: :boolean, desc: 'Daemonize process'
       method_option :queues,      aliases: '-q', type: :array,   desc: 'Queues to process with optional weights'
       method_option :require,     aliases: '-r', type: :string,  desc: 'Dir or path of the workers'
-      method_option :config_file, aliases: '-C', type: :string,  desc: 'Path to config file'
+      method_option :config,      aliases: '-C', type: :string,  desc: 'Path to config file'
```
phstc: hm @fredsadaghiani that ^ should break [this](https://github.com/phstc/shoryuken/blob/60905beefc0c8d103aac6b0dd22ede7bb4f71d7f/lib/shoryuken/environment_loader.rb#L43), it expects `config_file`.

We would need to change it in both places.  

Or in the `start` body, create an alias `ops[:config_file] = options[:config]`.

Thinking more on it, if we change it now, it would break for people using version 3 already. Version 3 is a [major version with breaking changes](http://semver.org/).

Maybe one possibility would be to use [`aliases`](https://github.com/erikhuda/thor/wiki/Method-Options#available-options) to set `config` as an alias for `config_file`, WDYT?

fredsadaghiani: I'd suggest supporting both (given that `config` was introduced in v3). Specifically, I suggest adding `method_option` for both parameters `config` and `config_file` which are redundant but raise an exception if *both* are specified. Then, in the `Shoryuken#start` method, I'd simply set `options[:config] = options[:config_file]` if it's specified.

fredsadaghiani: See 04c8cd4 for my proposal.

phstc: @fredsadaghiani WDYT about using a thor `aliases`?  Keeping `config_file` and making `config` an alias?

phstc:
@fredsadaghiani added to master with a deprecation warning https://github.com/phstc/shoryuken/commit/37dc67a360827e5843f156621320d67ba1790d11

Thanks for contributing 🍻

PR #362, opened 2017-04-27
Introduce dispatch event

Erol:
This PR introduces a `dispatch` event which fires whenever the dispatcher loop gets called. The possible use case I'm looking at is to monitor uptime and responsiveness of Shoryuken processes by having a callback to ping an instrumentation endpoint.

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -73,6 +73,8 @@ module Shoryuken
           return
         end
 
+        fire_event(:dispatch)
+
         logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }
 
         batched_queue?(queue) ? dispatch_batch(queue) : dispatch_single_messages(queue)
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -35,6 +35,7 @@ module Shoryuken
     timeout: 8,
     lifecycle_events: {
       startup: [],
+      dispatch: [],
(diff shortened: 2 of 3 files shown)


lib/shoryuken/manager.rb
```diff
@@ -73,6 +73,8 @@ def dispatch_now
           return
         end
 
+        fire_event(:dispatch, false)
```
phstc: @Erol `false` is the default value for `reverse`, I guess you can omit it. 

Can you add an expectation in the manager spec for this `fire_event` call?

Erol:
Made the changes requested @phstc. Could you please take a look again?

phstc:
@Erol awesome. Added to master 🍻

Erol:
That was fast, thanks!

Would you be releasing a new version soon? Would like to bump the version we're using so we can listen for the new event. 🙂

phstc:
Hi @Erol 

I don't have an ETA for releasing the new version, I usually wait a little more for adding more stuff. But if nothing shows up soon, I guess I can ship it, but no ETA yet.

You may use it from GitHub:

```ruby
gem 'shoryuken', git: 'https://github.com/phstc/shoryuken.git', ref: 'e331d'
```

Usually only requiring from GitHub is fine, since master is usually stable, but the `ref` above locks with your change.

WDYT?

Erol:
Yeah that's what I opted to do for now until the next release. Thanks for the ref lock tip!

phstc:
hi @Erol I've just released a new version 3.0.7 including your changes!

PR #261, opened 2016-10-24
fixes #249 Fix daemonize when don't set logfile in command line option

h3poteto:
When I didn't give logfile option in command line, shoryuken can't work fine:

```
rails@085a8fd8cf4a:~/app$ bundle exec shoryuken -R -C config/shoryuken.yml -d
You really should set a logfile if you're going to daemonize
/home/rails/app/vendor/bundle/ruby/2.3.0/gems/shoryuken-2.0.11/lib/shoryuken/cli.rb:80:in `daemonize'
/home/rails/app/vendor/bundle/ruby/2.3.0/gems/shoryuken-2.0.11/lib/shoryuken/cli.rb:34:in `run'
/home/rails/app/vendor/bundle/ruby/2.3.0/gems/shoryuken-2.0.11/bin/shoryuken:6:in `<top (required)>'
/home/rails/app/vendor/bundle/ruby/2.3.0/bin/shoryuken:23:in `load'
/home/rails/app/vendor/bundle/ruby/2.3.0/bin/shoryuken:23:in `<main>'
```

In `config/shoryuken.yml`, I wrote logifle option:

``` yaml
verbose: false
concurrency: 20
delay: 25
pidfile: ./tmp/pids/shoryuken.pid
logfile: ./log/shoryuken.log
```

but shoryuken didn't read logfile option.

Current shoryuken source code,
https://github.com/phstc/shoryuken/blob/v2.0.11/lib/shoryuken/cli.rb#L32-L39
call `daemonize` before `EnvironmentLoader.load`, so didn't read logifle option in config file.
That means, it is necessary to call initialize logger before daemonize.

This change intercept old EnvironmentLoader.load, because I reoder method's call to fix daemonize.

PR #368, opened 2017-05-23
Load rails before loading the config file

dotpyfe:
I believe that the relocation of where rails is loaded introduced in PR [#261](https://github.com/phstc/shoryuken/pull/261) breaks some shoryuken config files. In our case - we are using modules defined in our Rails project to dynamically generate queue names in the configuration. This used to work as far as shoryuken version 2.0.11.

The change is small - really just moving the rails load before config file load, like it used to be.

example config file:
```
delay: 25
concurrency: 25
queues:
<% MyApplication::ApplicationConfig.shoryuken_queues.each_value do |q_cfg| %>
  - [ <%= q_cfg[:q_name] %>, <%= q_cfg[:q_weight] %> ]
<% end %>
```

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -18,13 +18,13 @@ module Shoryuken
     end
 
     def setup_options
+      load_rails if options[:rails]
       initialize_options
       initialize_logger
       merge_cli_defined_queues
     end
 
     def load
-      load_rails if options[:rails]
       prefix_active_job_queue_names
       parse_queues
       require_workers


phstc:
Hi @dotpyfe 

Could you have a look at [this thread](https://github.com/phstc/shoryuken/issues/282#issuecomment-282526509). I can cycle through it again, but IIRC we can't change that.

dotpyfe:
ah gotcha - looks like the seperate environment loader solution will work for us, thanks for the pointer! that kind of information would be useful to have as part of the documentation, if you want i could change this pr to be a doc update

phstc:
> if you want i could change this pr to be a doc update

@dotpyfe Maybe a note here: https://github.com/phstc/shoryuken/wiki/Rails-Integration-Active-Job, WDYT?

dotpyfe:
sure that sounds like a great place, ill update the PR

dotpyfe:
i think changing the wiki with this kind of example will be enough:

```ruby
  # Replace Rails logger so messages are logged wherever Shoryuken is logging
  # Note: this entire block is only run by the processor, so we don't overwrite
  #       the logger when the app is running as usual.

  Rails.logger.level = Rails.application.config.log_level

  # config.server_middleware do |chain|
  #  chain.add Shoryuken::MyMiddleware
  # end
  
  vvv ADDED HERE vvv
  # when using Rails resources to configure queues, this would be where to do it
  # MyApplication.shoryuken_queues(Rails.env) do |q_name, q_weight|
  #   Shoryuken.add_queue(q_name, q_weight)
  # end
end
```

phstc:
@dotpyfe great idea 🍻  I added an example: 

```yaml
  # https://github.com/phstc/shoryuken/wiki/Rails-Integration-Active-Job

  # For dynamically adding queues prefixed by Rails.env 
  # %w(queue1 queue2) do |name|
  #   Shoryuken.add_queue("#{Rails.env}_#{name}, 1)
  # end
```

BTW Not sure if it would work for you, but I usually use:

```yaml
# ...
queues:
  - [ <%= ENV['QUEUE_1'] %>, 1]
  - [ <%= ENV['QUEUE_2'] %>, 1]
  - [ <%= ENV['QUEUE_3'] %>, 1]
```

Issue #369, opened 2017-05-23
Rescue in fetcher is too broad, should be given more specific exception classes


raquelxmoss:
Hello!

I use Shoryuken at work to run several workers, and recently the `rescue` [here](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/fetcher.rb#L19) has given us a bit of trouble. We've had a few instances where there have been failures (e.g. `Failed to open TCP connection to sqs ...`) that have been rescued. As such, they're not picked up by Rollbar, and the team is not alerted. We only find out our queue hasn't processed if we remember to log into the server in the morning and read the logs.

I wonder if you had considered being a bit more specific with the exception classes that are being rescued here?

Thanks 😄

phstc:
Hi @raquelxmoss 

hm how did you get `Failed to open TCP connection to sqs`? Was it working, then it lost internet access? I'm wondering if the connection is something we need to check at startup. Actually, Shoryuken does while trying to get the queue URL: https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/queue.rb#L12   

> . As such, they're not picked up by Rollbar

Even if it was raising up the error, I don't think it would be catch by Rollbar or similar trackers, since fetching is core Shoryuken and usually these trackers are triggered by workers, if a worker fails. Most likely raising up would kill the process or make it hang.

What I usually do is to configure CloudWatch alarms based on the queues Messages Available/Inflight, it the size is bigger than what it should for a given period of time, I get notified by email/Slack. Would that help you?

I'm totally open for discussing alternatives.

raquelxmoss:
Hey @phstc, I'm not sure how we managed to get the failed TCP connection, we're still looking into it but haven't made much headway. It appears that yes it was working (on an AWS EC2 instance), but I haven't figured out how it lost the connection. The worker never regained the connection, and had to be manually restarted. Ongoing mystery that we're still looking into 😅.

The CloudWatch alarms might be tricky in our case, as I don't think we can really nail down the business logic of how many messages 'should' be there at a given period of time, I'll have a think about it though.

I'm still curious about not raising up the error
> Most likely raising up would kill the process or make it hang.

In my case, having the error raise and kill the process might be more desirable -- it would trigger a restart, which would re-establish the TCP connection.

phstc:
Hey @raquelxmoss 

> In my case, having the error raise and kill the process might be more desirable -- it would trigger a restart, which would re-establish the TCP connection.

[...]

PR #374, opened 2017-05-31
Add new config option to force the fetcher to raise errors

MichaelDowse:
Depending on your process monitoring setup it may be preferable to have Shoryuken raise errors it encounters while fetching messages (rather than the default behaviour of silently logging these errors).

The fetcher can now be forced to raise errors using the `Shoryuken.sqs_client_receive_message_opts[:raise_errors]` configuration option. Existing behaviour is preserved when this option is not set.

Possibly this option would be better located in `default_worker_options`, I'm happy either way.

This is an opt-in solution to https://github.com/phstc/shoryuken/issues/369 which will not require a major version change.

### DIFF
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -9,18 +9,12 @@ module Shoryuken
 
       logger.debug { "Looking for new messages in '#{queue}'" }
 
-      begin
-        limit = available_processors > FETCH_LIMIT ? FETCH_LIMIT : available_processors
+      limit = available_processors > FETCH_LIMIT ? FETCH_LIMIT : available_processors
 
-        sqs_msgs = Array(receive_messages(queue, limit))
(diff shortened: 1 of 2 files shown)


phstc:
Hi @MichaelDowse 

I tried to turn off my internet connection with `raise_errors` set to true, and it didn't cause the Shoryuken process to be terminated. Is the expectation to kill the process?

phstc:
Hi @raquelxmoss @MichaelDowse

I'm closing this PR in favor of #389. I'm doing a major change in Shoryuken (multiple groups/fetchers), and as a side effect of this change, Shoryuken will also kill itself if the fetcher or manager fails.

Thanks again for raising this problem 🍻

MichaelDowse:
Thanks @phstc, sorry for not getting back to this. #389 looks good 👍

PR #378, opened 2017-06-04
Allow queue URL

### LINKED ISSUE #361: Optionally specify full url instead of just queue name
When we want to interact with different accounts, we have granted permission across them to access each others queues/topics etc, we have to specify the full url or arn of the queue (either one will work, really,) to access them, but shoryuken only exposes the ability to look for queues by name (and thus assumes they're in the same account as the credentials.)  Should the queue: attribute on the worker detect whether or not it's a URL and just use the url, instead of fetching it by name?  What effect would that have on the shoryuken.yml file and the specification of the queues there?  I may even be able to do a PR for this if we can agree on a solution that accommodates both the "queue" attribute of the worker (and thus the initializer of Queue class) and the specification of the queues to listen to in the config file...  Thoughts?

phstc:
Fixes #361

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -53,14 +53,15 @@ module Shoryuken
       if name_or_url.start_with?('https://sqs.')
         self.name = name_or_url.split('/').last
         self.url  = name_or_url
+
+        # Test if the supplied URL is valid
+        fifo?
       else
-        begin
-          self.name = name_or_url
-          self.url  = client.get_queue_url(queue_name: name_or_url).queue_url
-        rescue Aws::SQS::Errors::NonExistentQueue => e
-          raise e, "The specified queue '#{name}' does not exist."
-        end
+        self.name = name_or_url
+        self.url  = client.get_queue_url(queue_name: name_or_url).queue_url
       end
+    rescue Aws::Errors::NoSuchEndpointError, Aws::SQS::Errors::NonExistentQueue => ex
+      raise ex, "The specified queue #{name} does not exist."
     end
 
     def queue_attributes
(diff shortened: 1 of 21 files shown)


lib/shoryuken/queue.rb
```diff
@@ -52,6 +49,21 @@ def fifo?
 
     private
 
+    def set_name_and_url(name_or_url)
+      if name_or_url.start_with?('https://sqs.')
```
phstc: @dolan Are you using Amazon SQS or "SQS compatible queue systems"? If the latter, that `start_with?('https://sqs.')` won't work.

dolan: They're of this form: 'sqs://sqs.*/account_id/queue-name'.

phstc: @dolan so they don't start with `https`? hm

dolan: In amazon's docs I only see references to the ones with the http(s), but I'm using a front end for it called Convox that gives me queues URI's that start with sqs://.  I'm not sure about the implications of that.  I wonder if this is just a convox specific thing...  Interestingly enough, they can be accessed by changing the sqs:// to https:// so what you're proposing would work as long as everyone knows to do that.

phstc: @dolan could you try:

```ruby
client = Aws::SQS::Client.new
client.get_queue_attributes(queue_url: 'your-queue-url-with-protocol-sqs`, attribute_names: ['All'])
```

to see if that works?

phstc: @dolan just tested myself, and the AWS SDK does not support `sqs:` protocol, this a protocol made up by Convox. 

> as long as everyone knows to do that.

as long as Convox users know that the Convox makes up protocols and that's invalid in the AWS SDK context.

dolan: Agreed.

phstc: @dolan thanks for reviewing it 🍻

PR #387, opened 2017-06-22
Load Rails if specified `rails: true` in configuration file

aelesbao:
Currently, you are able to load Rails environment only if explicitly specified on the command line with `-R`. This behavior occurs because `Shoryuken::EnvironmentLoader` checks the `:rails` flag in the current `options` hash of the class, which contains only the path to the `:config_file`.

With this change, it will be possible to enable the Rails flag in the `shoryuken.yml`.


phstc:
@aelesbao thanks! Added to master 🍻

PR #402, opened 2017-07-10
Remove dead code

odk211:
First I found a typo `shoryuken::Logging` => `Shoryuken::Logging` .
But It seems dead code.


phstc:
@odk211 great catch, thanks! Added to master 🍻

PR #405, opened 2017-07-11
Fix active_job_queue_name_prefixing regression with groups

atyndall:
PR makes sure that queues inside groups are prefixed in the same way as ungrouped queues when `active_job_queue_name_prefixing` is enabled.


phstc:
@atyndall awesome! thanks for the fix, I will release a hotfix with it later today 🍻

PR #413, opened 2017-07-20
Fail fast on bad URL

### LINKED ISSUE #412: Minimal verification on URL-defined queue
In the queue initialization there is no verification done on the queue if it is supplied as a URL. Is there some basic verification that could be done here to align it more with the `name` verification?

Perhaps just check a queue attribute? Or some other minimal operation that can verify this is valid?

https://github.com/phstc/shoryuken/blame/master/lib/shoryuken/queue.rb#L61
```ruby
def set_name(name)
  self.name = name
  self.url  = client.get_queue_url(queue_name: name).queue_url
rescue Aws::Errors::NoSuchEndpointError, Aws::SQS::Errors::NonExistentQueue => ex
  raise ex, "The specified queue #{name} does not exist."
end

def set_url(url)
  self.name = url.split('/').last
  self.url  = url
end
```

maschwenk:
Closes #412

`get_queue_url` allows you to optionally specify the account_id of the queue, so this basically programmatically verifies that the queue is correct and verifies that the correct permissions are in place to use the URL in the first place.

http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/Client.html#get_queue_url

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -49,20 +49,21 @@ module Shoryuken
 
     private
 
-    def set_name(name)
-      self.name = name
-      self.url  = client.get_queue_url(queue_name: name).queue_url
-    rescue Aws::Errors::NoSuchEndpointError, Aws::SQS::Errors::NonExistentQueue => ex
-      raise ex, "The specified queue #{name} does not exist."
-    end
-
-    def set_url(url)
-      self.name = url.split('/').last
-      self.url  = url
-    end
-
     def set_name_and_url(name_or_url)
-      name_or_url.start_with?('https://sqs.') ? set_url(name_or_url) : set_name(name_or_url)
+      queue_url_params = if name_or_url.start_with?('https://sqs.')
+                           url = URI.parse(name_or_url).path.split('/')
+
+                           {
+                             queue_name: url.last,
+                             queue_owner_aws_account_id: url.second_to_last
+                           }
+                         else
+                           { queue_name: name_or_url }
+                         end
+      self.name = queue_url_params[:queue_name]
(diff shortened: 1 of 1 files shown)


maschwenk:
Man! `Reek` is no joke. Let me know if you have any suggestions on these

lib/shoryuken/queue.rb
```diff
@@ -49,20 +49,24 @@ def fifo?
...
+      else
+        { queue_name: name_or_url }
+      end
     end
 
     def set_name_and_url(name_or_url)
-      name_or_url.start_with?('https://sqs.') ? set_url(name_or_url) : set_name(name_or_url)
+      queue_url_params = parse_queue_url_params(name_or_url)
+      self.name = queue_url_params[:queue_name]
+      self.url  = client.get_queue_url(queue_url_params)
```
phstc: @maschwenk WDYT about something like [this](https://github.com/phstc/shoryuken/issues/412#issuecomment-316700989) for saving a request, since we need to check `fifo?` anyway, and we don't have to assume URL convention for queues: `*_, account_id, queue_name = URI.parse(name_or_url).path.split('/')`?

Do you mind also adding a test case `queue_spec`?

maschwenk: Ah! forgot you need to check `fifo?` at some point anyway. Will take a look

maschwenk:
Basically just copy-pasted what you had there, changed the names to `set_url`/`set_name` to `set_by_url`/`set_by_name` as that's a little more informative about what it's doing

phstc:
@maschwenk looks good. [I added a test case](https://github.com/phstc/shoryuken/commit/052bdc6f16fe39a84c4ec587664f9f6283eb4901) and merged into master! 🍻

PR #417, opened 2017-07-25
Look for polling strategy inside groups config

atyndall:
Allows group config to specify polling strategy in the .yml like so;

```
timeout: 1140
delay: 60

groups:
  r:
    concurrency: 1
    polling_strategy: StrictPriority
    queues:
      - x
      - y
      - z
  q:
    concurrency: 1
    polling_strategy: StrictPriority
    queues:
      - a
      - b
      - c
```

or like so

```
timeout: 1140
delay: 60
concurrency: 1
polling_strategy: StrictPriority
queues:
  - x
  - y
  - z
```

Note: this change would be backwards incompatible with the current way of setting strategy described [here](https://github.com/phstc/shoryuken/wiki/Polling-strategies). I have a version that also preserves that behaviour, but it's more complicated.

### DIFF
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -57,7 +57,7 @@ module Shoryuken
       end
 
       def polling_strategy(group)
-        option_group = (group == 'default') ? options : options[:groups][group]
+        option_group = group == 'default' ? options : options[:groups][group]
         strategy = option_group.to_h.fetch(:polling_strategy, Polling::WeightedRoundRobin)
         strategy = "Shoryuken::Polling::#{strategy}".constantize if strategy.is_a?(String)
         strategy
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -57,10 +57,18 @@ module Shoryuken
       end
 
       def polling_strategy(group)
-        option_group = group == 'default' ? options : options[:groups][group]
-        strategy = option_group.to_h.fetch(:polling_strategy, Polling::WeightedRoundRobin)
-        strategy = "Shoryuken::Polling::#{strategy}".constantize if strategy.is_a?(String)
-        strategy
+        strategy = (group == 'default' ? options : options[:groups][group]).to_h[:polling_strategy]
+
+        case strategy
+        when 'WeightedRoundRobin', nil # Default case
+          Polling::WeightedRoundRobin
+        when 'StrictPriority'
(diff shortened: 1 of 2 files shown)


lib/shoryuken/options.rb
```diff
@@ -57,7 +57,10 @@ def worker_registry=(worker_registry)
       end
 
       def polling_strategy(group)
-        options[group].to_h.fetch(:polling_strategy, Polling::WeightedRoundRobin)
+        option_group = group == 'default' ? options : options[:groups][group]
+        strategy = option_group.to_h.fetch(:polling_strategy, Polling::WeightedRoundRobin)
+        strategy = "Shoryuken::Polling::#{strategy}".constantize if strategy.is_a?(String)
```
phstc: @atyndall `constantize` depends on `ActiveSupport`.

Maybe the simplest way of doing it would be:

```ruby
if strategy == 'StrictPriority'
  Polling::StrictPriority
else
  Polling::WeightedRoundRobin
end
```

WDYT?

lib/shoryuken/options.rb
```diff
@@ -57,7 +57,10 @@ def worker_registry=(worker_registry)
...
+        strategy
       end
 
       def start_callback
```
phstc: @atyndall WDYT about adding some test cases? 

```ruby
# options_spec.rb

describe '.polling_strategy' do
  context 'when not set' do
    specify do
      expect(Shoryuken.polling_strategy('default')).to eq Shoryuken::Polling::WeightedRoundRobin
    end
  end

  context 'when set' do
    before do
      Shoryuken.options[:groups] = {
        'group1' => {
          polling_strategy: 'StrictPriority'
        }
      }
    end

    specify do
      expect(Shoryuken.polling_strategy('default')).to eq Shoryuken::Polling::WeightedRoundRobin
      expect(Shoryuken.polling_strategy('group1')).to eq Shoryuken::Polling::StrictPriority
    end
  end
end
```

atyndall:
@phstc Made changes as you suggested. I also restructured it a bit to ensure that if a user mistypes a string, it errors instead of silently picking the `WeightedRoundRobin` strategy.

phstc:
@atyndall yay added to master 🍻 I will ship a new release tomorrow including your changes.

phstc:
@atyndall 3.1.7 is out with your change 🍻

PR #426, opened 2017-08-21
Add argument support to #fire_event

hasbrettn:
The main feature of this pull request is that it will add the queue name to the `:dispatch` lifecycle event.

I can think of a couple different method signatures for `#fire_event`.  I'm happy to consider other signatures if you have opinions about it.

What I chose:

```ruby
def fire_event(event, options = {} , *event_args)
```

The other way that I considered:

```ruby
def fire_event( event, reverse = false, *event_args )
```

### DIFF
--- a/lib/shoryuken/util.rb
+++ b/lib/shoryuken/util.rb
@@ -4,13 +4,13 @@ module Shoryuken
       Shoryuken.logger
     end
 
-    def fire_event(event, reverse = false)
+    def fire_event(event, reverse = false, event_options = {})
       logger.debug { "Firing '#{event}' lifecycle event" }
       arr = Shoryuken.options[:lifecycle_events][event]
       arr.reverse! if reverse
       arr.each do |block|
         begin
-          block.call
+          block.call(event_options)
         rescue => ex
           logger.warn(event: event)
           logger.warn "#{ex.class.name}: #{ex.message}"
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -37,7 +37,7 @@ module Shoryuken
         return sleep(MIN_DISPATCH_INTERVAL)
       end
 
-      fire_event(:dispatch)
+      fire_event(:dispatch, false, queue_name: queue.name)
 
       logger.debug { "Ready: #{ready}, Busy: #{busy}, Active Queues: #{@polling_strategy.active_queues}" }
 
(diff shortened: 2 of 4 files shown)


lib/shoryuken/util.rb
```diff
@@ -4,13 +4,13 @@ def logger
       Shoryuken.logger
     end
 
-    def fire_event(event, reverse = false)
+    def fire_event(event, options = {}, *event_args)
```
phstc: Hi @hasbrettn 

I prefer `def fire_event(event, reverse = false, *event_args)` or `def fire_event(event, reverse = false, *event_options)` to avoid introducing breaking changes.

phstc: The latter `event_options` is only for making it more explicit, like `fire_event(:dispatch, false, queue_name: queue.name)`

phstc:
@hasbrettn thanks for the contribution 🍻 

Could you also give a use for passing the queue name?

hasbrettn:
Thanks for the feedback!

Our use case is that we'd like to push a stat to statsd each time the queue is polled to more easily distinguish between low-volume times, and down time.  Because we're getting this up in jruby -> docker -> kubernetes, a lot of the normal process process identification things that you'd normally do to make sure your process is up is more difficult and involves multiple teams.

This change makes it easy to identify that specific queues are actively being polled, rather than resorting to cloudwatch metrics.

lib/shoryuken/util.rb
hasbrettn: Are you thinking:

```ruby
def fire_event(event, reverse = false, event_options = {})
```

Such that event options is meant to be hash?

phstc: > Such that event options is meant to be hash?

@hasbrettn I was thinking on that for making the option more explicit, otherwise it would support `fire_event(..., false, queue_name, something_else)` and `fire_event(..., false, something_else, queue_name)`. If it becomes a hash, `queue_name` will be always `queue_name: `, no matter the order it was passed in. WDYT?

hasbrettn: I think that makes a lot of sense and I'm in support of that. I'll push a change.

hasbrettn:
I _think_ I made the changes we talked about.  Let me know if there's anything else that you want to see in here. 🥂

phstc:
Awesome! Your change is out 3.1.9. Thanks 🍻

PR #429, opened 2017-08-29
Auto retry fetch errors

### LINKED ISSUE #428: Manager occasionally dying
We're running into a problem where ~10 times a month, a shoryuken manager will die. The logs around the shutdown look like this:

```
Manager failed: No Content-Type received. Service returned the HTTP status code: 500

ERROR: /var/app/current/vendor/bundle/gems/aws-sdk-core-2.7.7/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
/var/app/current/vendor/bundle/gems/aws-sdk-core-2.7.7/lib/aws-sdk-core/plugins/idempotency_token.rb:18:in `call'
/var/app/current/vendor/bundle/gems/aws-sdk-core-2.7.7/lib/aws-sdk-core/plugins/param_converter.rb:20:in `call'
/var/app/current/vendor/bundle/gems/aws-sdk-core-2.7.7/lib/seahorse/client/plugins/response_target.rb:21:in `call'
/var/app/current/vendor/bundle/gems/aws-sdk-core-2.7.7/lib/seahorse/client/request.rb:70:in `send_request'
/var/app/current/vendor/bundle/gems/aws-sdk-core-2.7.7/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'
/var/app/current/vendor/bundle/gems/shoryuken-3.1.7/lib/shoryuken/queue.rb:43:in `receive_messages'
/var/app/current/vendor/bundle/gems/shoryuken-3.1.7/lib/shoryuken/fetcher.rb:35:in `receive_messages'
/var/app/current/vendor/bundle/gems/shoryuken-3.1.7/lib/shoryuken/fetcher.rb:16:in `fetch'
/var/app/current/vendor/bundle/gems/shoryuken-3.1.7/lib/shoryuken/manager.rb:82:in `dispatch_single_messages'

Received USR1, will soft shutdown down
```

The error message seems to indicate that SQS returned a 500 when the manager attempted to fetch messages. That exception is caught [here](https://github.com/phstc/shoryuken/blob/b434e1a51798f2f6f71e6349f83753ae687edf35/lib/shoryuken/manager.rb#L46), which results in a shutdown [here](https://github.com/phstc/shoryuken/blob/b434e1a51798f2f6f71e6349f83753ae687edf35/lib/shoryuken/manager.rb#L106).

Does that seem right? Is that the right behavior, or am I missing something?

phstc:
Fixes #428

### DIFF
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -9,7 +9,7 @@ module Shoryuken
     end
 
     def fetch(queue, limit)
-      do_with_retry(10) do
+      fetch_with_auto_retry(3) do
         started_at = Time.now
 
         logger.debug { "Looking for new messages in #{queue}" }
@@ -25,19 +25,20 @@ module Shoryuken
 
     private
 
-    def do_with_retry(max_attempts, &block)
+    def fetch_with_auto_retry(max_attempts, &block)
       attempts = 0
 
       begin
         yield
       rescue => ex
+        # Tries to auto retry connectivity errors
         raise if attempts >= max_attempts
 
         attempts += 1
 
         logger.debug { "Retrying fetch attempt #{attempts} for #{ex.message}" }
 
-        sleep(attempts)
+        sleep((1..5).to_a.sample)
 
         retry
       end
--- a/lib/shoryuken/fetcher.rb
+++ b/lib/shoryuken/fetcher.rb
@@ -9,19 +9,39 @@ module Shoryuken
     end
 
     def fetch(queue, limit)
-      started_at = Time.now
+      do_with_retry(10) do
+        started_at = Time.now
 
-      logger.debug { "Looking for new messages in #{queue}" }
+        logger.debug { "Looking for new messages in #{queue}" }
 
-      sqs_msgs = Array(receive_messages(queue, [FETCH_LIMIT, limit].min))
+        sqs_msgs = Array(receive_messages(queue, [FETCH_LIMIT, limit].min))
 
-      logger.info { "Found #{sqs_msgs.size} messages for #{queue.name}" } unless sqs_msgs.empty?
-      logger.debug { "Fetcher for #{queue} completed in #{elapsed(started_at)} ms" }
+        logger.info { "Found #{sqs_msgs.size} messages for #{queue.name}" } unless sqs_msgs.empty?
+        logger.debug { "Fetcher for #{queue} completed in #{elapsed(started_at)} ms" }
 
-      sqs_msgs
+        sqs_msgs
+      end
     end
 
     private
+
+    def do_with_retry(max_attempts, &block)
+      attempts = 0
+
+      begin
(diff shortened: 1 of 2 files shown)


lib/shoryuken/fetcher.rb
```diff
@@ -9,20 +9,40 @@ def initialize(group)
...
+        raise if attempts >= max_attempts
+
+        attempts += 1
+
+        logger.debug { "Retrying fetch attempt #{attempts} for #{ex.message}" }
+
+        sleep(attempts)
+
+        retry
+      end
```
phstc: @ethangunderson WDYT?

lib/shoryuken/fetcher.rb
```diff
@@ -9,20 +9,40 @@ def initialize(group)
     end
 
     def fetch(queue, limit)
-      started_at = Time.now
+      do_with_retry(10) do
```
ethangunderson: Any thoughts on making this configurable?

lib/shoryuken/fetcher.rb
```diff
@@ -9,20 +9,40 @@ def initialize(group)
...
+      begin
+        yield
+      rescue => ex
+        raise if attempts >= max_attempts
+
+        attempts += 1
+
+        logger.debug { "Retrying fetch attempt #{attempts} for #{ex.message}" }
+
+        sleep(attempts)
```
ethangunderson: This will act as a backoff until max attempts is reached, right? I like that.

phstc: @ethangunderson right!

lib/shoryuken/fetcher.rb
phstc: @ethangunderson I usually make that stuff configurable on demand basis, if someone asks to configure, just to avoid adding extra options.

waynerobinson:
Should this rescue from everything… or just fetch-related exceptions?

phstc:
@waynerobinson it used to rescue everything, then nothing, now it's rescuing everything again, but up to 3 attempts. I tried to find fetch-related exceptions before rescuing all, but it's hard to tell what are they. Is this broad rescue causing you issue?

waynerobinson:
Nothing in particular. But in the past I've had issues catching broad exceptions and things like catching TERM signals. 

Would catching all StandardErrors be enough? It's pretty broad, but doesn't include any of the system level exceptions.

phstc:
@waynerobinson makes sense, but I believe `rescue => ex` rescues from `StandardError` and up, for `Interrupt` and others, it would need to be `rescue Exception => ex`.

> StandardError -- default for rescue
> https://ruby-doc.org/core-2.2.0/Exception.html

waynerobinson:
Ahh… fair enough then! 😀

phstc:
@waynerobinson thanks a lot for checking 🍻

PR #431, opened 2017-08-31
shoryuken_options doesn't change arguments

ota42y:
The `shoryuken_option` method break arguments.
So we can't reuse option value.. :(

This change `stringify_keys` method create new hash and return it.


ota42y:
The test failed.
But it's another cause that has no relation
See https://github.com/phstc/shoryuken/pull/432

phstc:
Thanks @ota42y. Added to master.

PR #432, opened 2017-08-31
Lock aws-sdk

ota42y:
The awd-sdk-core 3.x is major version up and many breaking changes.
So we should lock 2.x until support 3.x version.

### DIFF
--- a/shoryuken.gemspec
+++ b/shoryuken.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency 'dotenv'
 
-  spec.add_dependency 'aws-sdk-core', '> 2'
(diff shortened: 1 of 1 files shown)


shoryuken.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency 'dotenv'
 
-  spec.add_dependency 'aws-sdk-core', '> 2'
+  spec.add_dependency 'aws-sdk-core', '~> 2'
```
phstc: @ota42y how about `=> 2`, so it would support 2 and 3?

http://guides.rubygems.org/patterns/#pessimistic-version-constraint

ota42y: No, shoryuken doesn't support aws-sdk-core 3.x and doesn't work 😢 
So we should lock 2.x.

PR #433, opened 2017-09-02
Make Shoryuken compatibility with aws-sdk-core 2 and 3

### LINKED ISSUE #430: aws-sdk-sqs is required
Hey, seems like for `aws-sdk-core` V3, `aws-sdk-sqs` dependency is required. Probably it worth to add this at least to readme.

phstc:
Fixes #430

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -1,6 +1,11 @@
 require 'yaml'
 require 'json'
 require 'aws-sdk-core'
+begin
+  require 'aws-sdk-sqs' unless defined?(Aws::SQS)
+rescue LoadError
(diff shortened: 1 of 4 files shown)


lib/shoryuken.rb
```diff
@@ -1,6 +1,13 @@
 require 'yaml'
 require 'json'
 require 'aws-sdk-core'
+begin
+  unless defined?(Aws::SQS)
+    require 'aws-sdk-sqs'
+  end
+rescue LoadError
+  fail "AWS SDK 3 requires aws-sdk-sqs to be installed separately. Please add gem 'aws-sdk-sqs' to your Gemfile"
+end
```
phstc: @pyromaniac `Aws::CORE_GEM_VERSION` is the one that returns > 3. But I think only by checking if `Aws::SQS` we can infer if 3 or not.

pyromaniac:
I think it is perfect :shipit:

PR #449, opened 2017-12-28
Add -t option to CLI

maschwenk:
From the wiki:

https://github.com/phstc/shoryuken/wiki/Signals#term

>Tells Shoryuken to hard shutdown within a timeout. Any processors that do not finish within the timeout are hard terminated and their messages are pushed back to AWS SQS. The timeout defaults to 8 seconds since all Heroku processes must exit within 10 seconds.

The described `-t` option was never actually configured in Thor

> The timeout can be defined through the -t option or in the -C shoryuken.yml config by setting the timeout: N value.


phstc:
Thanks @maschwenk. Added to master 🍻

phstc:
@maschwenk 3.2.0 is out with your change! 🍻

Issue #450, opened 2017-12-28
Worker options are not inherited


fabiokr:
Say I have a base worker class that my other workers inherit from, and I have this on it:

```ruby
class BaseWorker
  include Shoryuken::Worker

  shoryuken_options body_parser: :json, auto_delete: true
end

class ChildWorker < BaseWorker
end
```

The child workers are not preserving the config from the parent class. Is that expected behavior?

phstc:
Hi @fabiokr 

Good point. I guess it needs some hackery like [this](https://github.com/mperham/sidekiq/pull/246) ([newer version](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/worker.rb)).

PR #451, opened 2017-12-29
Preserves parent worker class options

fabiokr:
Relates to #450

Uses a `shoryuken_class_attribute` implementation that preserves parent
worker class options and allow child classes to override those.

Example:

```ruby
class ParentWorker
  include ShoryukenWorker

  shoryuken_options queue: "myqueue", auto_delete: true
end

class ChildWorker < ParentWorker
  shoryuken_options auto_delete: false
end
```

In this example, the ChildWorker will carry on the parent options, and
override the `auto_delete` configuration.

### DIFF
--- a/lib/shoryuken/worker.rb
+++ b/lib/shoryuken/worker.rb
@@ -2,6 +2,7 @@ module Shoryuken
   module Worker
     def self.included(base)
       base.extend(ClassMethods)
+      base.shoryuken_class_attribute :shoryuken_options_hash
     end
 
     module ClassMethods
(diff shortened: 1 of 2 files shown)


fabiokr:
cc/ @phstc

fabiokr:
@phstc CodeClimate is complaining about the `shoryuken_class_attribute` implementation, but I tried to avoid changing that as it is basically the same implementation as `sidekiq` and `activesupport`. Thoughts?

phstc:
Muito obrigado @fabiokr e @newx 🍻 

Added to master!

I'm planning to release a new version next week, I will update you both when it's out.

phstc:
@fabiokr @newx 3.2.0 is out with this change! 🍻

fabiokr:
@phstc Awesome, thanks!

PR #460, opened 2018-01-19
reset Client @@queues on .sqs= call

etipton:
I have an interesting use-case where I need to use temporary access credentials that are refreshed periodically. In order to get Shoryuken to play nicely with that paradigm, I needed to make this change.

### DIFF
--- a/lib/shoryuken/client.rb
+++ b/lib/shoryuken/client.rb
@@ -12,6 +12,10 @@ module Shoryuken
       end
 
       def sqs=(sqs)
+        # Since the @@queues values (Shoryuken::Queue objects) are built referencing @@sqs, if it changes, we need to
+        #   re-build them on subsequent calls to `.queues(name)`.
+        @@queues = {}
+
         @@sqs = sqs
       end
     end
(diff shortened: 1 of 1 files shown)


phstc:
hi @etipton 

So every time you refresh the keys you

```ruby
Shoryuken.sqs_client = new_client
Shoryuken::Client.sqs = new_client
```
?

I don't remember the reasons why Shoryuken ended up with `Shoryuken::Client.sqs` and `Shoryuken.sqs_client`, unless I'm missing something it seems misleading. This isn't directly related to your fix, I'm just checking how you refresh keys.

etipton:
Ah, good question; I should've included that. Here's what I'm doing:

```
creds = aws_creds.values_at('access_key_id', 'secret_access_key', 'session_token')
Shoryuken::Client.sqs = Aws::SQS::Client.new(credentials: Aws::Credentials.new(*creds))
```

I tried updating creds on an already-instantiated SQS client (raw AWS SDK) and couldn't get it to work, so it must be doing some memoization or something... that's why I'm re-creating a new Client object every time.

phstc:
@etipton alright, thanks! Added to master 🍻

PR #462, opened 2018-01-22
Exclude job_id from message deduplication when ActiveJob

### LINKED ISSUE #457: Exclude job_id from the message body when calculating message_deduplication_id for ActiveJob
I tested out a FIFO queue with deduplication enabled it didn't prevent duplicate jobs from getting submitted. This is while using ActiveJob which assigns a unique job_id to each message and thus makes the body different. 

Has there been any discussion about removing the `job_id` when processing the `message_deduplication_id` while using ActiveJob?

phstc:
Fix #457

### DIFF
--- a/lib/shoryuken/extensions/active_job_adapter.rb
+++ b/lib/shoryuken/extensions/active_job_adapter.rb
@@ -34,7 +34,7 @@ module ActiveJob
         register_worker!(job)
 
         queue = Shoryuken::Client.queues(job.queue_name)
-        queue.send_message(message(job, options))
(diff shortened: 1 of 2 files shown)


phstc:
Hi @nathanpalmer 

Could you review the changes to remove the `job_id` from the `message_deduplication_id`?

Thanks

phstc:
@nathanpalmer Thanks for reviewing it. I had the same feeling about `except`, but I ended up using it for simplicity purposes, since that code should only run when Rails/`activesupport` is loaded anyways.

PR #478, opened 2018-02-22
Allow Batch Size to be Specified for Requeue

waynerobinson:
Sometimes batches of 10 messages will be larger than the 256KB per-request limit of batch send so allow this to be specified on the command-line during re-queue.

A better (future) option would be to automatically detect the message size and only send as many messages as will fit in a batch send request.


phstc:
@waynerobinson great! Added to master 🍻

Issue #467, opened 2018-01-31
Cannot define multiple 'included' blocks for a Concern, only in development environment


tap87:
The way shoryuken uses eager_load for Rails 4+ is causing ActiveSupport::Concern::MultipleIncludedBlocks error in my development environment. I didn't have this error when I used sidekiq. 
My app uses Rails 5. It seems like eager_loading is causing these errors, when I comment out https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L70, everything works fine. The code should be refactored to eager_load only for rails version 4.0. This is how sidekiq does it now: https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L262-L271.
[This PR fro sidekiq](https://github.com/mperham/sidekiq/pull/2457) might be relevant to this issue.

Here is the stacktrace: 
```
bundler: failed to load command: shoryuken (/Users/tahir/.rbenv/versions/2.3.1/bin/shoryuken)
ActiveSupport::Concern::MultipleIncludedBlocks: Cannot define multiple 'included' blocks for a Concern
  /Users/tahir/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-5.0.6/lib/active_support/concern.rb:126:in `included'
  /Users/tahir/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/hydra-access-controls-10.5.0/app/models/concerns/hydra/access_controls/embargoable.rb:7:in `<module:Embargoable>'
  /Users/tahir/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/hydra-access-controls-10.5.0/app/models/concerns/hydra/access_controls/embargoable.rb:3:in `<module:AccessControls>'
  /Users/tahir/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/hydra-access-controls-10.5.0/app/models/concerns/hydra/access_controls/embargoable.rb:2:in `<module:Hydra>'
  /Users/tahir/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/hydra-access-controls-10.5.0/app/models/concerns/hydra/access_controls/embargoable.rb:1:in `<top (required)>'
  /Users/tahir/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:477:in `load'
[...]

PR #480, opened 2018-03-13
Don't force eager load for Rails 5

tap87:
Addresses #467

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -66,8 +66,10 @@ module Shoryuken
       else
         # Painful contortions, see 1791 for discussion
         require File.expand_path('config/application.rb')
-        ::Rails::Application.initializer 'shoryuken.eager_load' do
-          ::Rails.application.config.eager_load = true
+        if ::Rails::VERSION::MAJOR == 4
+          ::Rails::Application.initializer 'shoryuken.eager_load' do
+            ::Rails.application.config.eager_load = true
+          end
         end
         require 'shoryuken/extensions/active_job_adapter' if Shoryuken.active_job?
         require File.expand_path('config/environment.rb')


phstc:
Hi @tap87 

Thanks for the PR!

I will give it some testing this week.

Have you tested with/without a daemon `-d`?

lib/shoryuken/environment_loader.rb
```diff
@@ -66,8 +66,10 @@ def load_rails
       else
         # Painful contortions, see 1791 for discussion
         require File.expand_path('config/application.rb')
-        ::Rails::Application.initializer 'shoryuken.eager_load' do
-          ::Rails.application.config.eager_load = true
+        if ::Rails::VERSION::MAJOR == 4
```
phstc: @tap87 for keeping compatibility, I'm wondering if instead of checking for Rails 4, we should check for Rails greater than or equals to 5, so all the previous version, would continue to work in the same they are now, given the issue seems to be only with Rails 5 and greater. 

BTW I'm using shoryuken with Rails 5 and it's working fine.

tap87: Yes, we could take that approach. I tried to follow sidekiq here since that is were the existing code was adapted from. https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L261

tap87:
I have tested it both with and without the daemon option, and it works fine

tap87:
@phstc Any updates on this PR. This is how sidekiq does it. [Resque](https://github.com/chanks/que/issues/184) also checks for app config instead of forcing eager load. I also checked [que](https://github.com/chanks/que/issues/184) and they also don't force eager_load in Rails 5

phstc:
Hi @tap87 

Sorry, I've been very busy lately, but I added this to my list to check this week. If everything works well, I will merge and release this week.

phstc:
@tap87 3.2.3 is out with your fix 🎉  Great job, thanks 🍻

tap87:
Thanks @phstc

PR #492, opened 2018-04-13
implement TSTP handling

### LINKED ISSUE #479: Implement TSTP a-la Sidekiq for signaling future deploy
We're using Shoryuken, and would greatly benefit from a similar feature to [Sidekiq's TSTP signal handling](https://github.com/mperham/sidekiq/wiki/Signals#tstp), meaning "stop consuming the queue and finish your work, because the process will die soon".

As I understood, shoryuken was a Sidekiq fork. Was the removal of this signal a decision based on some limitation of the SQS client integration, or could this be done?

sideshowdave7:
Implements TSTP signal handler: Fixes #479 

TSTP tells Shoryuken to "quiet" as it will be shutting down at some point in the near future. It will stop fetching new jobs but continue working on current jobs. Use TSTP + TERM to guarantee shutdown within a time period. Best practice is to send TSTP at the start of a deploy and TERM at the end of a deploy.

Note you still need to send TERM to actually exit the Shoryuken process.
 
Tests:

I tested against a shoryuken process with no active jobs and it behaved as expected.
I also tested sending some jobs to the shoryuken process and verified active ones were allowed to finish, and new jobs stopped being accepted.

I looked around and didn't see any spec tests for signals.

### DIFF
--- a/lib/shoryuken/runner.rb
+++ b/lib/shoryuken/runner.rb
@@ -19,7 +19,7 @@ module Shoryuken
     def run(options)
       self_read, self_write = IO.pipe
 
-      %w(INT TERM USR1 TTIN).each do |sig|
+      %w(INT TERM USR1 TSTP TTIN).each do |sig|
         begin
           trap sig do
             self_write.puts(sig)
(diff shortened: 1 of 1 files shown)


lib/shoryuken/runner.rb
```diff
@@ -123,8 +129,9 @@ def handle_signal(sig)
       case sig
       when 'USR1' then execute_soft_shutdown
       when 'TTIN' then print_threads_backtrace
+      when 'TSTP' then execute_terminal_stop
       when 'TERM', 'INT'
-        logger.info { "Received #{sig}, will shutdown down" }
+        logger.info { "Received #{sig}, will shutdown" }
```
phstc: 🍻

phstc:
@sideshowdave7 thanks! Added to master 🍻

PR #496, opened 2018-04-18
Config by environment

matrinox:
Lets you configure shoryuken by environment, just like `config/database.yml`

```yaml
concurrency: 1
queues:
- [queue1, 1]
- [queue2, 2]

staging:
  concurrency: 5
  queues:
  - [staging-queue1, 1]
  - [staging-queue2, 2]

production:
  concurrency: 10
  queues:
  - [production-queue1, 1]
  - [production-queue2, 2]
```

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -45,10 +45,14 @@ module Shoryuken
       fail ArgumentError, "The supplied config file #{path} does not exist" unless File.exist?(path)
 
       if (result = YAML.load(ERB.new(IO.read(path)).result))
-        result.deep_symbolize_keys
+        merge_config_by_environment(result.deep_symbolize_keys, ENV['RAILS_ENV'].to_sym)
       else
         {}
       end
+    end
+
+    def merge_config_by_environment(config, environment)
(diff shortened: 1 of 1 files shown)


lib/shoryuken/environment_loader.rb
```diff
@@ -45,12 +45,16 @@ def config_file_options
       fail ArgumentError, "The supplied config file #{path} does not exist" unless File.exist?(path)
 
       if (result = YAML.load(ERB.new(IO.read(path)).result))
-        result.deep_symbolize_keys
+        merge_config_by_environment(result.deep_symbolize_keys, ENV['RAILS_ENV'].to_sym)
```
phstc: hi @matrinox 

Shoryuken has also a standalone version, which is not Rails dependent, so `RAILS_ENV` is not an option. 

If you want to make it environment based, I would recommend you to setup queue names via ENV variables.

```yml
  queues:
  - [<%= ENV['QUEUE1'] %>, 1]
  - [<%= ENV['QUEUE2'] %>, 2]
```

matrinox: Ahh, didn't know you could that. Thanks, I'll try it out

PR #507, opened 2018-06-21
Allow supplying an empty list of queues

ocisly:
It can be very useful to start a Shoryuken process without any queues.
For instance, this allows us to easily "pause" Shoryuken.

This currently possible using the config file:
```bash
$ touch no_queues.yml
$ bundle exec shoryuken --config no_queues.yml
WARN: No queues supplied
INFO: Starting
```

However, it's impossible using command line arguments:
```bash
$ bundle exec shoryuken --queues=""
NoMethodError: undefined method `start_with?' for nil:NilClass
```

It would be amazing to make the behaviour consistent ❤️

(Note: of course, it is possible to not specify the `--queues` option at all, but then it becomes hard to script.)

### DIFF
--- a/bin/shoryuken
+++ b/bin/shoryuken
@@ -35,9 +35,9 @@ module Shoryuken
         opts[:config_file] = opts.delete(:config) if opts[:config]
 
         # Keep compatibility with old CLI queue format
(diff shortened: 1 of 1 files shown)


phstc:
Thanks @ocisly. Added to master 🍻

PR #509, opened 2018-07-10
Fixes retained memory due to recursive calls

rubemz:
# Context 

We keep observing the Shoryuken memory growing slowly in production and started debugging to find the root cause.

# The Fix

The previous logic used a recursion for the dispatch loop. The problem with that is that recursive calls only free memory when the recursion resolves.

This PR solves it by using a `while` statement instead of the recursion. 

`rbtrace` and `heapy` were used in order to detect the retained objects, this is one of the head dumps after shoryuken ran for a while:

```
High Ref Counts
==============================

2646  /Users/nakamura/.rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/gems/shoryuken-3.2.3/lib/shoryuken/manager.rb:31

allocated by memory (19576546) (in bytes)
==============================

84552  /Users/nakamura/.rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/gems/shoryuken-3.2.3/lib/shoryuken/manager.rb:31
```

See the [shoryuken manager code](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/manager.rb#L31).

It is saving around 80k from memory, in this case, when a `while` statement is used rather than the recursion.

After this fix the `shoryuken/manager` doesn't even show up as part of the `rbtrace` and `heapy` output

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -13,6 +13,7 @@ module Shoryuken
       @busy_processors  = Concurrent::AtomicFixnum.new(0)
       @executor         = executor
       @running          = Concurrent::AtomicBoolean.new(true)
+      @dispatch_running = Concurrent::AtomicBoolean.new(false)
     end
 
     def start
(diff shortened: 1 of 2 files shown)


rubemz:
btw, I'm still seeing the memory slowly growing and this doesn't solve it completely

phstc:
@rubemz what are the steps to reproduce the memory leak? Are you getting it just by running shoryuken with empty queues? Is there any chance your worker is causing the leakage? If you comment out your worker body, does the memory continue to leak?

phstc:
Closing due inactivity.

PR #512, opened 2018-08-15
Log SQS receipt handle

akirakoyasu:
I'd like to know the receipt handle for deleting a message which cause error.

my case:
Some Active Jobs ran on Shoryuken and one of them always raised an error from its condition. I tried to delete the job but couldn't because of SQS invisibility.

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -40,7 +40,10 @@ module Shoryuken
     end
 
     def receive_messages(options)
-      client.receive_message(options.merge(queue_url: url)).messages.map { |m| Message.new(client, self, m) }
+      client.receive_message(options.merge(queue_url: url)).messages.map do |m|
+        logger.info { "Received message from SQS(#{name}), message_id: #{m.message_id}, receipt_handle: #{m.receipt_handle}" }
+        Message.new(client, self, m)
+      end
     end
 
     def fifo?


phstc:
hi @akirakoyasu 

I understand your use case, but this will drastically increase the log output. You don't have access to the SQS message when using ActiveJob, so you can't get the receipt handle, but you do if you [use Shoryuken workers](https://github.com/phstc/shoryuken/wiki/Worker-options) - this can be an option.

Also, aligned to your suggestion/PR, you could try a custom [middleware](https://github.com/phstc/shoryuken/wiki/Middleware).

Some examples

- https://github.com/honeybadger-io/honeybadger-ruby/blob/9b5c74940cde8c2f971352649c4e3b69217d2508/lib/honeybadger/plugins/shoryuken.rb
- https://github.com/airbrake/airbrake/blob/9419a961eb4db226505881e62ed335e987f47ae1/lib/airbrake/shoryuken.rb

Besides those options above, I would recommend to try to make your workers idempotent, so you would't have problems with re-processing, another options is to use [dead letter queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html).

akirakoyasu:
Hi @phstc,

I understand, and will check options and middlewares that you mentioned.

Thank you!

PR #514, opened 2018-08-24
Allow inline executor to test workers in batch mode

atongen:


phstc:
@atongen LGTM. Added to master 🍻

PR #516, opened 2018-09-04
Make it possible to send options to JSON.parse

tothpeter:
It would be nice to be able to send options to `JSON.parse` through the main options:
 `shoryuken_options body_parser: :json, json_parse: { symbolize_names: true }`


tothpeter:
I was thinking of it, and yes I agree with you.

Issue #517, opened 2018-09-14
Inconsistent behaviour in InlineExecutor vs DefaultExecutor


ocisly:
After several hours of debugging, I've discovered a subtle yet important difference between the two executor implementations.

`Shoryuken::Worker::DefaultExecutor.perform_in` relies on the worker class' implementation of `perform_async`:

```
def perform_in(worker_class, interval, body, options = {})
  # ...
  worker_class.perform_async(body, options.merge(delay_seconds: delay))
end
```

In contrast, `Shoryuken::Worker::InlineExecutor.perform_in` always calls its own implementation of `perform_async`:

```
def perform_in(worker_class, _interval, body, options = {})
  perform_async(worker_class, body, options)
end
```

Now, assume you have a custom `Shoryuken.worker_executor` (e.g. because you want to use `to_json` instead of `JSON.dump`) 

```
class ToJsonExecutor
  class << self
    def perform_async(worker_class, body, options = {})
      Shoryuken::Worker::DefaultExecutor.perform_async(worker_class, body.to_json, options)
    end

    def perform_in(worker_class, interval, body, options = {})
      Shoryuken::Worker::DefaultExecutor.perform_in(worker_class, interval, body.to_json, options)
    end
  end
end
```

This is wrong and will call `body.to_json` twice. The same code will work fine if you replace `DefaultExecutor` with `InlineExecutor`, which is confusing.

Which fix would make more sense?
1) change `DefaultExecutor.perform_in` to be self contained (calling its own implementation of `perform_async` instead of going via the worker class) *or*
2) change `InlineExecutor.perform_in` to delegate to the worker class

phstc:
Hi @ocisly 

Thanks for investigating that and the well detailed explanation.

For a PR, I would go with option 2.

BTW if the body parser wasn't just an example, I would recommend you to use [`body_parser`](https://github.com/phstc/shoryuken/wiki/Worker-options#body_parser) instead.

ocisly:
Thanks! I've opened a PR for option 2.

I don't think `body_parser` is an option here, because I need to control the serialisation of the message body *before it is sent*. I guess client middleware could work?

ocisly:
Just looked into that. client middleware wouldn't work out of the box, because `sanitize_message!` is called before the middleware gets invoked.

PR #518, opened 2018-09-17
InlineExecutor: invoke `perform_async` via the worker class

ocisly:
Fixes phstc/shoryuken#517

### DIFF
--- a/lib/shoryuken/worker/inline_executor.rb
+++ b/lib/shoryuken/worker/inline_executor.rb
@@ -20,7 +20,7 @@ module Shoryuken
         end
 
         def perform_in(worker_class, _interval, body, options = {})
-          perform_async(worker_class, body, options)
+          worker_class.perform_async(body, options)
         end
 
         private


lib/shoryuken/worker/inline_executor.rb
```diff
@@ -20,7 +20,7 @@ def perform_async(worker_class, body, _options = {})
         end
 
         def perform_in(worker_class, _interval, body, options = {})
-          perform_async(worker_class, body, options)
+          worker_class.perform_async(body, options)
```
phstc: @ocisly I couldn't test the code yet, but just eyeballing it, shouldn't this line call `worker_class.new.perform`  like the `perform_async` is calling? Did you have a chance to test this change? Is it fixing what you've reported in #517?

ocisly: Yes, this is the fix. It's modelled after `DefaultExecutor.perform_in`, which also calls `worker_class.perform_async`, which leads to the following sequence of calls:

0. Shoryuken::InlineExecutor.perform_in
1. Shoryuken::Worker.perform_async
2. Shoryuken.worker_executor.perform_async
3. worker_class.new.perform

If we were to call `perform` on the worker instance as you suggest, then the executor won't be invoked at all, leading to the inconsistency described in #517.

ocisly: I've tested this behaviour manually by creating an executor like the one described in the issue and invoking `perform_in` on the `DefaultWorker` from the examples.

phstc:
Thanks @ocisly. It is added to master 🍻

PR #528, opened 2018-10-29
fix fifo

timbreitkreutz:


timbreitkreutz:
Sorry for the noise-- will be submitting a cleaner version.

lib/shoryuken/queue.rb
```diff
@@ -44,7 +44,8 @@ def receive_messages(options)
     end
 
     def fifo?
-      @_fifo ||= queue_attributes.attributes[FIFO_ATTR] == 'true'
+      return @_fifo if defined?(@_fifo)
+      @_fifo = queue_attributes.attributes[FIFO_ATTR] == 'true'
```
phstc: Hi @timbreitkreutz, thanks for the PR.

What is the error this change is fixing? `||=` is a common pattern for instance variables.

timbreitkreutz: This form isn't safe for booleans, especially with side effects.  The ||= will always be re-evaulated if the resulting value is falsy.

Cleaner PR here: #529

PR #531, opened 2018-11-06
Memoize Shoryuken::Queue#visibility_timeout to avoid hitting SQS API multiple times

julien-unitylab:
Add some memoization on Shoryuken::Queue#visibility_timeout to avoid hitting SQS API multiple times.

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -14,7 +14,7 @@ module Shoryuken
     end
 
     def visibility_timeout
-      queue_attributes.attributes[VISIBILITY_TIMEOUT_ATTR].to_i
+      @visibility_timeout ||= queue_attributes.attributes[VISIBILITY_TIMEOUT_ATTR].to_i
     end
 
     def delete_messages(options)


lib/shoryuken/queue.rb
```diff
@@ -14,7 +14,7 @@ def initialize(client, name_or_url)
     end
 
     def visibility_timeout
-      queue_attributes.attributes[VISIBILITY_TIMEOUT_ATTR].to_i
+      @visibility_timeout ||= queue_attributes.attributes[VISIBILITY_TIMEOUT_ATTR].to_i
```
phstc: Hi @julien-unitylab for FIFO it is easy to merge the memoization (#529), since a FIFO queue can't be changed to a standard queue, so that value once it is loaded, won't be changed. 

But for visibility timeout it can change.  If we memoize it, Shoryuken processes will need to be restart to catch up with the changes.

I'm wondering if this could be an option in the `shoryuken.yml#cache_visibility_timeout` (default to false for not introducing break changes). 

WDYT?

phstc: Hey @julien-unitylab 

Did you have a chance to check that ☝️

julien-unitylab: Hello @phstc,
I'm really sorry for my late reply, thanks to have taken care of this :).
Let's continue on #533, I'm closing this PR.

Cheers

PR #532, opened 2018-11-17
Add queue name to inline executor sqs_msg

atongen:
Making `sqs_msg` `OpenStruct` from `InlineExecutor` look more like `Shoryuken::Message` allows easier testing when alternate queue name is passed in as an option.


phstc:
@atongen looks good. Thank you! Added to master 🍻

Issue #454, opened 2018-01-05
Delay cannot be set per processing group


vvondra:
The delay will be ignored in this case
```
groups:
  group_a:
    delay: 10
    queues:
      - queue_a
```

The global delay of 5s will be used in this case:
```
delay: 5
groups:
  group_a:
    delay: 10
    queues:
      - queue_a
```
  

:pushpin: #543

phstc:
Good catch. It was not added in the first version of Processing Groups, it was waiting for a "feature request". [This](https://github.com/phstc/shoryuken/blob/6688d181a36bae98a186e92b398e5546fece7814/lib/shoryuken/polling/base.rb#L63) should be based on the processing group.

lucas-ibotta:
It would be nice if we could configure delay when adding the group.
Something like `Shoryuken.add_group(group, concurrency, delay: 5)`.

I'd be happy to PR that change.

phstc:
@lucas-ibotta 👏 great work on adding the delay support per processing #543! 

Release 4.0.3 is out with delay support per processing group 🎉

PR #543, opened 2018-12-12
Add delay option into Shoryuken.add_group

lucas-ibotta:
* :wave: Hey Pablo! Thanks for all your work on Shoryuken. 😄 
* Fixes https://github.com/phstc/shoryuken/issues/454


phstc:
@lucas-ibotta yay that's great! Thank you. Sorry for the delay to get back to you, I will have a look to this this week and merge it down!

lib/shoryuken/options.rb
```diff
@@ -30,9 +30,10 @@ def active_job?
         defined?(::ActiveJob)
       end
 
-      def add_group(group, concurrency)
+      def add_group(group, concurrency, delay: 0)
         groups[group] ||= {
           concurrency: concurrency,
+          delay: delay,
```
phstc: > Sounds great. Let me know about lib/shoryuken/environment_loader.rb. I think I'm missing where it parses the YML variables to add the delay into the group.

@lucas-ibotta sure. I think [here](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L133), you can:

```ruby
Shoryuken.add_group(group, options.fetch(:concurrency, 25), delay: options.fetch(:delay, 0))
```

phstc: [Here](https://github.com/phstc/shoryuken/blob/bc7938942270a1171d1597064698d88c3f36a4ec/lib/shoryuken/polling/base.rb#L63) you also need to get the delay from the group instead of the global one.

[This](https://github.com/phstc/shoryuken/blob/352b20642b4b8a123b821228231ed5a5c618a9cc/lib/shoryuken/options.rb#L79) is an example how we retrieve `polling_strategy` per group. You can do something similar for `delay`.

Does that make sense?

For testing that locally, you can create two groups with a 25 seconds delay and the other with the default delay. Then play in the console: 

```sh
aws sqs send-message --message-body "Test" --queue-url https://sqs.us-east-1.amazonaws.com/.../...
```

WDYT?

lucas-ibotta: Yup, makes sense. I'll start working on that soon and push the changes. 😄

lib/shoryuken/polling/base.rb
```diff
@@ -60,7 +60,11 @@ def ==(other)
       private
 
       def delay
-        Shoryuken.options[:delay].to_f
+        if name == 'default'
+          Shoryuken.options[:delay].to_f
+        else # allow delay settings per group
+          Shoryuken.options[:groups].to_h[name]).to_h[:delay]
```
phstc: @lucas-ibotta you probably need `.to_f` as well for defaulting to `0.0`, like the main one.

phstc: Maybe:

```ruby
Shoryuken.options[:groups].to_h[name].to_h.fetch(:delay, Shoryuken.options[:delay]).to_f
```

lucas-ibotta: Done. Much cleaner. That code does cause one new code climate failure though.
I'm unsure about the Travis failures too. The tests pass for me locally.

lucas-ibotta:
```
opts = Shoryuken.options
opts[:groups].to_h[name].to_h.fetch(:delay, opts[:delay]).to_f
```
That might pass code climate?

lucas-ibotta:
There's two blockers on this currently: 
* Code climate checks that seems ignorable. What do you think about those?
* The tests are passing for me locally with `rspec` but `bundle exec rspec` raises a `user-defined signal 1` that causes the build to exit with a 1. That's happening on Travis also. Seems like an issue with bundler.

Any insight into these issues?

lucas-ibotta:
Cool. Tests are passing now. Just code climate remains.

phstc:
@lucas-ibotta thanks. I will test it locally to see how it goes. Don't worry about these two feedback from Code Climate - it is fine.

lib/shoryuken/polling/base.rb
```diff
@@ -59,8 +59,8 @@ def ==(other)
 
       private
 
-      def delay
-        Shoryuken.options[:delay].to_f
+      def delay(queue)
+        Shoryuken.options[:groups].to_h[queue].to_h.fetch(:delay, Shoryuken.options[:delay]).to_f
```
phstc: @lucas-ibotta the delay is per group, not per queue, this `Shoryuken.options[:groups].to_h[queue]` is trying to find a delay for a given `queue`. 

Also, for avoiding breaking changes, I would recommend to use `def delay(group = nil)` so if anyone else is using a custom poller and calling `delay` without arguments, it should still work.

.byebug_history
```diff
@@ -0,0 +1,17 @@
...
+exit
+delay
+exit
+exitr
+self.delay
+self
+options
+hash
+@queues
+name
```
phstc: @lucas-ibotta it is probably better to `~/.gitignore` this file. I believe it is more env specific. WDYT?

lib/shoryuken/polling/strict_priority.rb
```diff
@@ -68,8 +68,9 @@ def queue_paused?(queue)
       end
 
       def pause(queue)
-        return unless delay > 0
-        @paused_until[queue] = Time.now + delay
+        delay_time = delay(queue)
```
phstc: @lucas-ibotta delay should be per group, not per queue.

lib/shoryuken/polling/base.rb
lucas-ibotta: Yes, nice catch.

lucas-ibotta:
Yes, nice catch. I have been on holiday, but I'll address those points soon and test things out locally. 👌

lib/shoryuken/polling/strict_priority.rb
lucas-ibotta: Going from the String of the queue name to the group delay required an extra helper... I'd like your thoughts on that implementation. I am testing this locally right now though.

lib/shoryuken/options.rb
```diff
@@ -47,6 +48,16 @@ def add_queue(queue, weight, group)
         end
       end
 
+      def group_for_queue(queue)
+        groups.each do |group|
+          # note, there is a bug here where two queues with the same name
+          # will choose the first queue and group
+          if group[1][:queues].include?(queue)
+            return group.first # the group name
+          end
```
lucas-ibotta: @phstc Going from the queue name string to the group delay configuration here. Is there any existing validation that a queue name must be unique across all groups?

phstc: @lucas-ibotta the same queue can be in multiple groups. We can't get the delay based on the queue name.

What I think we could do, is to change the `PollingStrategy` to receive a second argument `delay = nil` (default to `nil` to avoid breaking changes) then [here](https://github.com/phstc/shoryuken/blob/352b20642b4b8a123b821228231ed5a5c618a9cc/lib/shoryuken/launcher.rb#L75) pass it as the second argument, similar to [this](https://github.com/phstc/shoryuken/blob/352b20642b4b8a123b821228231ed5a5c618a9cc/lib/shoryuken/options.rb#L78-L79).

```ruby
 def create_managers
  Shoryuken.groups.map do |group, options|
    Shoryuken::Manager.new(
      Shoryuken::Fetcher.new(group),
      Shoryuken.polling_strategy(group).new(options[:queues], Shoryuken.delay(group)),
      options[:concurrency],
      executor
    )
  end
end
```


Then in `BaseStrategy#queue` we:

```ruby
def delay
  @delay || Shoryuken.options[:delay].to_f
end
```

WDYT?

lib/shoryuken/polling/base.rb
```diff
@@ -57,10 +57,8 @@ def ==(other)
         end
       end
 
-      private
```
lucas-ibotta: I made `delay` public and wrote some tests for it on each polling strategy.

lucas-ibotta:
Testing this locally and working out some of the bugs.

spec/shoryuken/polling/strict_priority_spec.rb
```diff
@@ -86,6 +86,14 @@
     end
   end
 
+  describe 'delay' do
+    it 'sets delay based on group' do
+      delay_polling = Shoryuken::Polling::StrictPriority.new(queues, 25)
+      expect(delay_polling.send(:delay)).to eq(25.0)
+      expect(subject.send(:delay)).to eq(1.0)
```
lucas-ibotta: I can't track down where this is a `1.0` instead of a `0.0`

lucas-ibotta:
Ok, I tested this locally with
```ruby
# in setup_shoryuken.rb
  Shoryuken.add_group('no-delay', 5)
  Shoryuken.add_queue("https://sqs.us-east-1.amazonaws.com/../normal-queue", 5, 'no-delay')
  Shoryuken.add_group('delay', 5, delay: 25)
  Shoryuken.add_queue("https://sqs.us-east-1.amazonaws.com/../delay-queue", 5, 'delay')
end

class DelayWorker
  include Shoryuken::Worker
    shoryuken_options queue: 'https://sqs.us-east-1.amazonaws.com/../delay-queue'
                  
  def perform(_sqs_message, body)
    puts "DelayWorker Called with #{body}"
  end
end

class NormalWorker
    include Shoryuken::Worker
    shoryuken_options queue: 'https://sqs.us-east-1.amazonaws.com/../normal-queue'
                  
  def perform(_sqs_message, body)
    puts "NormalWorker Called with #{body}"
  end
end
```

Then ran `bundle exec shoryuken --require ./setup_shoryuken.rb` and everything worked!

phstc:
@lucas-ibotta great work, thank you! I will have a look/test and get back to you. I may need a couple of days though, but if all is good I will merge it down and release a new version 🍻

lucas-ibotta:
Excellent. Thanks! Happy New Year to you too.

PR #551, opened 2019-01-30
queue names can be any url

chrisdhanna:
Previously, you could pass a queue name, or a queue url to shoryuken, as long as the url started with `https://sqs.`. This form works for many situations, but not when incorporating shoryuken with localstack. Localstack's sqs urls don't have the leading `sqs.`. You cannot pass the queue name to shoryuken when using localstack, because localstack doesn't implement get_queue_url yet. 

This PR relaxes the url constraint. Instead of having to start with `https://sqs.`, now the name will be treated as a url if it contains `://`.

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -66,7 +66,7 @@ module Shoryuken
     end
 
     def set_name_and_url(name_or_url)
-      if name_or_url.start_with?('https://sqs.')
+      if name_or_url.include?('://')
         set_by_url(name_or_url)
 
         # anticipate the fifo? checker for validating the queue URL


phstc:
Thanks @chrisdhanna, added to master 🍻 

I added a test for the change: https://github.com/phstc/shoryuken/commit/34f8ae64bbec53d7fa388540923eff938a66ff5f#diff-95f741df0c437dc5a0711737d5528d83R30 - Is this `"http://localhost:4576/queue/#{queue_name}"` how a Localstack URL looks like?

chrisdhanna:
yes @phstc , thats the correct format of a localstack queue URL. Thanks for adding the test!

PR #561, opened 2019-05-02
CI: Add 2.6.3

olleolleolle:
This PR only updates the CI matrix to the latest generally available Ruby versions.

Source for version numbers: https://github.com/rvm/rvm/blob/master/config/known


phstc:
yay Thanks @olleolleolle. Added to master.

PR #562, opened 2019-05-15
Load rails before parsing queues

matrinox:
### Context
Configuration loading files like dotenv and figaro need to be loaded before Shoryuken parses the shoryuken config file, which may contain environment variables

### Changes
- Moved `load_rails` method from `EnvironmentLoader` to `Shoryuken::Runner`

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -24,7 +24,6 @@ module Shoryuken
     end
 
     def load
-      load_rails if Shoryuken.options[:rails]
       prefix_active_job_queue_names
       parse_queues
       require_workers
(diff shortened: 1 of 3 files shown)


phstc:
Hi @matrinox - thanks for submitting this PR.

I will have to do some digging for this change. I remember there were some issues trying to load Rails before that call.

Meanwhile, can you have a look at https://github.com/phstc/shoryuken/issues/282#issuecomment-265979943?

matrinox:
@phstc Thanks, I think that's related. I'll take a look later today

matrinox:
@phstc Can you quickly explain to me where you folks have landed on this issue [here](https://github.com/phstc/shoryuken/issues/282#issuecomment-282526509):

> In case it isn't clear, there is a dependency cycle in the boot process:
> 
> Shoryuken must daemonize before Rails is loaded, to avoid clobbering global state (#219)
> Rails must be loaded before the config file is read, so it can reference Rails (#197, this issue)
> The config file must be read before Shoryuken daemonizes, to pick up the logfile option (#249)
> We can't have all three of these things at the same time. No matter how we rearrange the order of operations, we have to give up one of them.

Thanks!

phstc:
@matrinox we load Rails after the config file is read. What's the configuration you need to read from Figaro?

If it is a just a queue name, you could programmatically load them with [`add_group` and `add_queue`](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/options.rb#L33-L48).

matrinox:
It would be nice to define things in a config like with sidekiq, etc. Our current solution is to just load Rails in the shoryuken bin stub. Do you foresee any problems with that?

phstc:
@matrinox unless Sidekiq has changed, you can't access `Rails.env` etc in the `sidekiq.yml`.

> Do you foresee any problems with that?

Changing the load order will require a lot of testing. There were a lot of back-forth with that in the past. 

> It would be nice to define things in a config like with sidekiq, etc

What type of things? `shoryuken.yml` is an ERB, you can `require dotenv` etc.

matrinox:
@phstc I haven't been successful with just requiring `dotenv` in `shoryuken.yml`. So far we have had no issues with requiring rails before starting shoryuken. I'll close this PR then.

LyricL-Gitster:
I finally stumbled upon this after a few hours of trying to determine why my config wasn't being picked up from my `dotenv` files, so I thought I'd offer some help to anyone else that may come along.

https://gist.github.com/LyricL-Gitster/487b31193f8dd23d07bddcbe042a542c

PR #569, opened 2019-06-18
Fix empty_queue_fetch_delay bug in batch processing mode

### LINKED ISSUE #568: Shoryuken does not respect empty_queue_fetch_delay in batch mode
I am implementing a batch processor like this:
```ruby
class DummyBatchProcessor
  include Shoryuken::Worker

  shoryuken_options body_parser: Oj,
                    auto_delete: true,
                    batch: true,
                    queue: "myqueue"
...
```

and setting empty queue fetch delay at application startup like this:
```ruby
      Shoryuken.groups.clear
      Shoryuken.add_group("default",
        configuration.concurrency,
        delay: configuration.empty_queue_fetch_delay)
```

What is strange is that delay works with normal "one-message" processors, but not when I set it to batch mode.

Is it some kind of a bug? If so, where should I look for the code which needs to be fixed?

Thanks!

:pushpin: #569

snoopie:
Fix bug where empty queues were not paused in batch processing mode (fixes #568).

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -74,9 +74,9 @@ module Shoryuken
     end
 
     def dispatch_batch(queue)
-      return if (batch = @fetcher.fetch(queue, BATCH_LIMIT)).none?
+      batch = @fetcher.fetch(queue, BATCH_LIMIT)
       @polling_strategy.messages_found(queue.name, batch.size)
-      assign(queue.name, patch_batch!(batch))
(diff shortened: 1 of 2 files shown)


phstc:
Hi @snoopie 

Thanks for the fix 🎉

I released a new version 5.0.0 with your fix 🚀 

Please let me know if that works.

snoopie:
Thanks! I actually checked the fix on my branch and it worked like charm.

Issue #552, opened 2019-01-31
Queue order is reversed if passed through CLI


Drew-Goddyn:
**How to reproduce:**
1. Pass in 2 queues as a CLI argument: ` shoryuken start -q queue1 queue2 -C ./shoryuken.yml`
2. View the value of `Shoryuken.options[:queues]` in the worker

**Expected output:** 
We expect the order of the queues to match the order they were passed in. I.E. we expect `Shoryuken.options[:queues]` to evaluate to `[["queue1"],["queue2"]]`

**Current output:**
https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L84 deletes and then readds each queue to the Shoryuken options resulting in `Shoryuken.options[:queues]` evaluating to `[["queue2"],["queue1"]]`

**Why is this important?**
If you've also defined `polling_strategy: StrictPriority` within your config file, then the order of the queues passed in through the CLI determines which queue is high and which is low priority. This issue causes the first queue passed in the CLI to be the lowest priority when it should be the highest according to the documentation: https://github.com/phstc/shoryuken/wiki/Polling-strategies

phstc:
Hi @Drew-Goddyn 

Great catch, and thanks for the well detailed explanation.

### why `merge_cli_defined_queues`

The idea behind that method is, if you have a `shoryuken.yml` with:

```yaml
# shoryuken.yml

queues:
  - [queue1, 8]
  - [queue2, 4]
```

And you call `shoryuken` passing in the same queue names with different values:

```sh
shoryuken start -q queue1,10 queue2,20 -C shoryuken.yml
```

#### Expected behavior 

The values passed in the CLI should override the ones configured in `shoryuken.yml`. `queue1` should have priority 10 instead of 8, and `queue2` should have priority 20 instead of 3.

### Quick fix


Add a `.dup` [here](https://github.com/phstc/shoryuken/blob/34f8ae64bbec53d7fa388540923eff938a66ff5f/lib/shoryuken/environment_loader.rb#L80):

```ruby
cli_defined_queues = options[:queues].to_a.dup
```

(would you like to create a pull request for it? 🙏)

### Possible better fix

[...]

PR #583, opened 2019-10-17
Remove EnvironmentLoader#merge_cli_defined_queues

Drew-Goddyn:
### What did you do?

Removed a method from EnvironmentLoader which caused an issue with queue order

### Why?

This method is not needed as CLI options are merged overtop of config options in
EnvironmentLoader#initialize_options

see https://github.com/phstc/shoryuken/issues/552 for further details

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -20,7 +20,6 @@ module Shoryuken
     def setup_options
       initialize_options
       initialize_logger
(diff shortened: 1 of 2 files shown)


Drew-Goddyn:
@phstc better late than never 🤣

phstc:
Looks great! Thank you

phstc:
@Drew-Goddyn 5.0.2 is out with your changes 🤜

PR #585, opened 2019-11-01
Deliver Messages Async

allcentury:
Prior to this change delivering 5000 messages took on average:

    user     system      total        real
    6.351633   0.535708   6.887341 ( 42.842282)

After this change, I am able to delivery 5000 messages in:

    user     system      total        real
    7.160261   1.237327   8.397588 (  7.676820)

In my rough benchmarks I see a 5-7x increase.  This is valuable for us because we enqueue a large number of jobs at once in a loop, something akin to:

```ruby
MyModel.expensive_query.select(:id).each { |obj| MyJob.perform_later(obj) }
```

A couple call outs:

1.  I was unsure if this should be configurable (ie async vs blocking)
2.  Tests either need to call `sleep` to allow the executor to work or I need to mock differently.  I'm open to either but thought I'd check first
3.  In addition to on/off configuration, I'm wondering if options sent to `Concurrent::CachedThreadPool.new` should also be configurable.

### DIFF
--- a/spec/shoryuken/queue_spec.rb
+++ b/spec/shoryuken/queue_spec.rb
@@ -86,12 +86,14 @@ RSpec.describe Shoryuken::Queue do
       expect(sqs).to receive(:send_message).with(hash_including(message_body: 'msg1'))
 
       subject.send_message(message_body: 'msg1')
+      sleep(0.0001)
     end
 
     it 'accepts a string' do
       expect(sqs).to receive(:send_message).with(hash_including(message_body: 'msg1'))
 
       subject.send_message('msg1')
+      sleep(0.0001)
     end
 
     context 'when a client middleware' do
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ If you are using AWS SDK version 3, please also add this line:
 gem 'aws-sdk-sqs'
 ```
 
-The extra gem `aws-sdk-sqs` is required in order to keep Shoryuken compatible with AWS SDK version 2 and 3. 
+The extra gem `aws-sdk-sqs` is required in order to keep Shoryuken compatible with AWS SDK version 2 and 3.
 
 And then execute:
 
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -11,6 +11,7 @@ module Shoryuken
     def initialize(client, name_or_url)
       self.client = client
       set_name_and_url(name_or_url)
+      @thread_pool = Concurrent::CachedThreadPool.new(fallback_policy: :caller_runs, auto_terminate: true)
     end
 
     def visibility_timeout
@@ -34,12 +35,14 @@ module Shoryuken
       options = sanitize_message!(options).merge(queue_url: url)
 
       Shoryuken.client_middleware.invoke(options) do
-        client.send_message(options)
+        send_concurrently { client.send_message(options) }
       end
     end
 
     def send_messages(options)
-      client.send_message_batch(sanitize_messages!(options).merge(queue_url: url))
+      send_concurrently do
+        client.send_message_batch(sanitize_messages!(options).merge(queue_url: url))
+      end
     end
 
     def receive_messages(options)
@@ -54,6 +57,10 @@ module Shoryuken
     end
 
     private
+
+    def send_concurrently(&block)
+      @thread_pool << block
+    end
 
     def set_by_name(name)
       self.name = name


spec/shoryuken/queue_spec.rb
```diff
@@ -86,12 +86,14 @@
       expect(sqs).to receive(:send_message).with(hash_including(message_body: 'msg1'))
 
       subject.send_message(message_body: 'msg1')
+      sleep(0.0001)
```
allcentury: as per my PR message - I need to do this in more specs to get them all of them to pass

phstc:
hi @allcentury 

That's interesting and I understand the use case. It reminds me [sucker_punch](https://github.com/brandonhilkert/sucker_punch). But I'm wondering if this need is too specific to be in Shoryuken itself. 

For example, if you "enqueue" a lot of threads and SQS starts to throttle or something similar, how would you recover that? 

You can also implement this outside of Shoryuken like this 👇 

```ruby
send_concurrently do
  MyWorker.perform_async(params)
end
```

Or that

```ruby
class MyWorker
  include Shoryuken::Worker
  include SendConcurrently 
end

# then

MyWorker.send_concurrently(params)
```

I need to think more about it, but if we add this to Shoryuken, I believe it will need to be "on a call basis" like `MyWorker.concurrent.perform_async(params)` so it is not all or nothing.

🤔 that might be tricky with Active Job, maybe `MyJob.set(concurrent: true).perform_later(params)`.

WDYT?

allcentury:
@phstc - thanks for the thoughtful reply.  

> You can also implement this outside of Shoryuken

We are in fact doing exactly as you described, we have a shim that allows us to use a thread pool.  It has reduced expensive cron tasks from being long running, enabling us to run them more frequently without overlap.  We see SQS delivery performance issues throughout our day, sometimes our 99 percentile is over 11 seconds.  In the example above - we have far more workers consuming messages than producing so that's where our scalability issue resides.

> That's interesting and I understand the use case. It reminds me sucker_punch. But I'm wondering if this need is too specific to be in Shoryuken itself.

That's a really good question to ask - I don't have a perfect answer.  From my point of view, the gains are free and the trade off feels slightly invisible to the users of shoryuken but I could be missing something.  

> For example, if you "enqueue" a lot of threads and SQS starts to throttle or something similar, how would you recover that?

That's an interesting point.  We ensure all jobs are idempotent on our end, so if a thread is producing a lot of jobs and gets killed or raises an unrecoverable exception, our k8s policy will restart that task.  Since our jobs are idempotent we recover just fine - but I can see your point here.  I do think though that I need to pick a better concurrent ruby object here as [thread pools lack robust exception handling](https://github.com/ruby-concurrency/concurrent-ruby/issues/616).

As far as the API goes, I feel like `perform` and `perform_later` should be left as-is.  I think it's better if a user configures a `BlockingDelivery` or a `ConcurrentDelivery` object.  Assuming I update with a better abstraction, WDYT of that idea?

phstc:
> That's a really good question to ask - I don't have a perfect answer. From my point of view, the gains are free and the trade off feels slightly invisible to the users of shoryuken but I could be missing something.

If we add this to Shoryuken, users must explicitly set it. There are a lot of issues that may happen with threads like that, and some may not be easily detected. Like a ton of messages in the thread pool, then process crash - we can't assume that all workers are idempotent 😢 


> As far as the API goes, I feel like perform and perform_later should be left as-is. I think it's better if a user configures a BlockingDelivery or a ConcurrentDelivery object. Assuming I update with a better abstraction, WDYT of that idea?

So you are using Active Job, not standard Shoryuken workers, right?

AJ interface is very limited, so we can't extend much. But there's an interesting example for this [Async adapter](https://github.com/rails/rails/blob/9bfb73a5f672520e792df8f2e87e2e6a977f3db3/activejob/lib/active_job/queue_adapters/async_adapter.rb#L25-L28).

Maybe you could create a  `ShoryukenConcurrentSendAdapter` that just wraps the [Shoryuken adapter](https://github.com/phstc/shoryuken/blob/9d20e85539756cab45890003df60ec083f769f83/lib/shoryuken/extensions/active_job_adapter.rb) adding the thread support.

```ruby
config.active_job.queue_adapter = \ 
  ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter.new \
    ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter
```
That could be a totally adapter too, like `AJConcurrentSendAdapter`, that wraps any AJ adapter. 

WDYT?

phstc:
Just found this: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-throughput-horizontal-scaling-and-batching.html#horizontal-scaling it is in Java, but it is an interesting idea.

allcentury:
@phstc - I've tried an alternative approach in https://github.com/allcentury/shoryuken/commit/5a84a257fe9ad4a68dd4e6029ec42c9a9a14b6f5 ; This seems to be working fine and gives the callers the ability to add their own success/failure handlers.  Take a look and let me know what you think.

phstc:
@allcentury great! should we close this PR in favor of one with https://github.com/allcentury/shoryuken/commit/5a84a257fe9ad4a68dd4e6029ec42c9a9a14b6f5?

allcentury:
closed in favor of #588

PR #588, opened 2019-11-22
Deliver Messages Async

allcentury:
Prior to this change delivering 5000 messages took on average:

    user     system      total        real
    6.351633   0.535708   6.887341 ( 42.842282)

After this change, I am able to delivery 5000 messages in:

    user     system      total        real
    7.160261   1.237327   8.397588 (  7.676820)

In my rough benchmarks I see a 5-7x increase.  This is valuable for us because we enqueue a large number of jobs at once in a loop, something akin to:

```ruby
MyModel.expensive_query.each { |obj| MyJob.perform_later(obj) }
```

Usable by setting:

```ruby
config.active_job.queue_adapter = ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter.new
```

Handlers are available for both success and failure and must respond to
call.

Example:

```ruby
adapter = ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter.new
adapter.success_handler = ->(job, options) { StatsD.increment(job.class.name + "success") }
adapter.error_handler = ->(err, (job, options)) { StatsD.increment(job.class.name + "failure") }

config.active_job.queue_adapter = adapter
```

### DIFF
--- a/lib/shoryuken/default_worker_registry.rb
+++ b/lib/shoryuken/default_worker_registry.rb
@@ -1,7 +1,7 @@
 module Shoryuken
   class DefaultWorkerRegistry < WorkerRegistry
     def initialize
-      @workers = {}
+      @workers = Concurrent::Hash.new
     end
 
     def batch_receive_messages?(queue)
--- a/lib/shoryuken/extensions/active_job_concurrent_send_adapter.rb
+++ b/lib/shoryuken/extensions/active_job_concurrent_send_adapter.rb
@@ -1,43 +1,42 @@
 # ActiveJob docs: http://edgeguides.rubyonrails.org/active_job_basics.html
 # Example adapters ref: https://github.com/rails/rails/tree/master/activejob/lib/active_job/queue_adapters
- module ActiveJob
-   module QueueAdapters
-     # == Shoryuken concurrent adapter for Active Job
-     #
-     # This adapter sends messages asynchronously (ie non-blocking) and allows
-     # the caller to set up handlers for both success and failure
-     #
-     # To use this adapter, set up as:
-     #
-     # adapter = ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter.new
-		 # adapter.success_handler = ->(job, options) { StatsD.increment(job.class.name + "success") }
-		 # adapter.error_handler = ->(err, (job, options)) { StatsD.increment(job.class.name + "failure") }
-		 #
-		 # config.active_job.queue_adapter = adapter
-     class ShoryukenConcurrentSendAdapter < ShoryukenAdapter
+module ActiveJob
+  module QueueAdapters
+    # == Shoryuken concurrent adapter for Active Job
+    #
+    # This adapter sends messages asynchronously (ie non-blocking) and allows
+    # the caller to set up handlers for both success and failure
+    #
+    # To use this adapter, set up as:
+    #
+    # adapter = ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter.new
+    # adapter.success_handler = ->(job, options) { StatsD.increment(job.class.name + "success") }
+    # adapter.error_handler = ->(err, (job, options)) { StatsD.increment(job.class.name + "failure") }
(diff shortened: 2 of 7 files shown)


lib/shoryuken/default_worker_registry.rb
```diff
@@ -1,7 +1,7 @@
 module Shoryuken
   class DefaultWorkerRegistry < WorkerRegistry
     def initialize
-      @workers = {}
+      @workers = Concurrent::Hash.new
```
allcentury: I changed this because I'm worried https://github.com/allcentury/shoryuken/blob/712b35c4dd253ef275d7756760b604cd184d45bd/lib/shoryuken/extensions/active_job_adapter.rb#L34 is not thread safe.

lib/shoryuken/extensions/active_job_concurrent_send_adapter.rb
```diff
@@ -0,0 +1,42 @@
...
+    #
+    # To use this adapter, set up as:
+    #
+    # adapter = ActiveJob::QueueAdapters::ShoryukenConcurrentSendAdapter.new
+    # adapter.success_handler = ->(job, options) { StatsD.increment(job.class.name + "success") }
+    # adapter.error_handler = ->(err, (job, options)) { StatsD.increment(job.class.name + "failure") }
+    #
+    # config.active_job.queue_adapter = adapter
+    class ShoryukenConcurrentSendAdapter < ShoryukenAdapter
+      attr_accessor :error_handler
```
allcentury: I'm open to moving this as constructor requirement but the advantage to leaving this as an `attr_accessor` is callers can switch the handlers for different tasks.  For instance, we might care about an error handler for a mission critical code path and not in others.  Allowing callers to change this at run time has value for us but I'm open to alternatives.

phstc:
@allcentury This looks great 🎉  I will QA it locally and then :shipit: 

We should eventually combine some code for the active job adapters with the standard workers, but I believe this is outside of the scope of this pull request. 

* https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/worker.rb
* https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/worker/default_executor.rb

lib/shoryuken/extensions/active_job_concurrent_send_adapter.rb
phstc: > Allowing callers to change this at run time has value for us but I'm open to alternatives.

@allcentury Would change at runtime be thread-safe?

phstc:
Hi @allcentury 

Awesome work on this change 🎉  I did some minor changes in #589 and merged it into master.

lib/shoryuken/extensions/active_job_concurrent_send_adapter.rb
allcentury: @phstc - probably not, I'm also naively thinking of my own use case, where a k8s cron task is run.  I can change this in the task though so I think using dependency injection (like your added commit does) is the way to go.

allcentury:
@phstc - wonderful!  Thanks again

phstc:
@allcentury 5.0.3 is out with your changes, can you test it out?

Thank you 🎉

phstc:
Hi @allcentury 

Do you want to add a section in [Sending a message
](https://github.com/phstc/shoryuken/wiki/Sending-a-message) explaining the `ShoryukenConcurrentSendAdapter` so other users can also know about it?

allcentury:
@phstc - do you think it should go there or [here](https://github.com/phstc/shoryuken/wiki/Rails-Integration-Active-Job)?

phstc:
@allcentury good call. Since it is only supported with AJ, [the wiki you mentioned](https://github.com/phstc/shoryuken/wiki/Rails-Integration-Active-Job) makes more sense! thanks for checking 🎉

allcentury:
I updated the docs here: https://github.com/phstc/shoryuken/wiki/Rails-Integration-Active-Job thanks @phstc !

Issue #591, opened 2020-01-05
Respect message specific VisibilityTimeout


moyuanhuang:
Happy new year!

According to [AWS documenation][doc], it's possible to set the `VisibilityTimeout` for a specific message regardless of the `VisibilityTimeout` value in the queue level. This is useful since in our application, messages in one queue can be processed by different shoyuken workers, which takes up very different amount of time. We want to set the visibility timout to a bigger value for those messages that takes longer time, so that it won't get processed multiple times.

Fortunately, we were able to use the `Message#visibility_timeout=` setter to change the message's visibility timeout (from 60 seconds to 2 hours in this case). However, shoryuken would log warnings like this

```
started at 2020-01-04 08:01:09 +0000
...
exceeded the queue visibility timeout by 705623.076762 ms
completed in: 765623.076762 ms
```

...which is misleading.

I think this is caused by shoryuken always uses the queue level visibility timeout, and not respecting the visibility timeout in the message level

https://github.com/phstc/shoryuken/blob/af6ba758767aa7e541cae0ad3e4b146ef8eb7792/lib/shoryuken/middleware/server/timing.rb#L16

I'm happy to open a PR if this is a legit issue or otherwise I might be missing something here?

Thanks!

[doc]: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibility.html

phstc:
Hi @moyuanhuang 

Happy new year!

I will be happy to review a PR changing that code to verify the `sqs_msg` visibility timeout if any otherwise fallback to the queue visibility timeout for a less misleading message. I don't remember if the `sqs_msg` returns the queue visibility timeout if not set at the message level you may need a fallback to the queue one, but it is worth checking before adding the fallback. 

BTW there's this option [`auto_visibility_timeout`](https://github.com/phstc/shoryuken/wiki/Worker-options#auto_visibility_timeout) that could help your use case as well.

PR #592, opened 2020-01-07
Message level visibility timeout

moyuanhuang:
This PR tries to address #591, it

- Adds the ability to query `visibility_timeout` from message level
- Changes the logic inside `Timing` middleware to print "exceeds visibility timeout" warning according to message level `visibility_timeout`

### DIFF
--- a/lib/shoryuken/message.rb
+++ b/lib/shoryuken/message.rb
@@ -1,12 +1,16 @@
 module Shoryuken
   class Message
     attr_accessor :client, :queue_url, :queue_name, :data
+    attr_reader :become_available_at, :visibility_timeout
 
     def initialize(client, queue, data)
       self.client     = client
       self.data       = data
       self.queue_url  = queue.url
       self.queue_name = queue.name
+
+      @visibility_timeout = Shoryuken::Client.queues(queue).visibility_timeout
+      self.become_available_at = Time.now + @visibility_timeout
     end
 
     def delete
@@ -28,6 +32,8 @@ module Shoryuken
         receipt_handle: data.receipt_handle,
         visibility_timeout: timeout
       )
+      @visibility_timeout = timeout
+      self.become_available_at = Time.now + timeout
     end
 
     def message_id
--- a/lib/shoryuken/middleware/server/timing.rb
+++ b/lib/shoryuken/middleware/server/timing.rb
@@ -4,17 +4,20 @@ module Shoryuken
       class Timing
         include Util
 
-        def call(_worker, queue, _sqs_msg, _body)
+        def call(_worker, _queue, sqs_msg, _body)
           started_at = Time.now
 
           logger.info { "started at #{started_at}" }
 
           yield
 
-          total_time = elapsed(started_at)
+          ended_at = Time.now
+          total_time = (ended_at - started_at) * 1000
 
-          if (total_time / 1000.0) > (timeout = Shoryuken::Client.queues(queue).visibility_timeout)
-            logger.warn { "exceeded the queue visibility timeout by #{total_time - (timeout * 1000)} ms" }
+          if ended_at > sqs_msg.become_available_at
+            log_msg = 'exceeded the message visibility timeout by ' \
+              "#{(ended_at - sqs_msg.become_available_at) * 1000} ms"
+            logger.warn { log_msg }
           end
 
           logger.info { "completed in: #{total_time} ms" }
--- a/spec/shoryuken/middleware/server/timing_spec.rb
+++ b/spec/shoryuken/middleware/server/timing_spec.rb
@@ -18,6 +18,8 @@ RSpec.describe Shoryuken::Middleware::Server::Timing do
   end
 
   it 'logs timing' do
+    expect(sqs_msg).to receive(:become_available_at).and_return(Time.now + 1_000)
+
     expect(Shoryuken.logger).to receive(:info) do |&block|
       expect(block.call).to match(/started at/)
     end
@@ -29,8 +31,14 @@ RSpec.describe Shoryuken::Middleware::Server::Timing do
   end
 
   context 'when exceeded the `visibility_timeout`' do
+    before do
+      @started_at = Time.now 
+      @ended_at = @started_at + 120
+    end
+
     it 'logs exceeded' do
-      allow(subject).to receive(:elapsed).and_return(120_000)
+      expect(Time).to receive(:now).and_return(@started_at, @ended_at)
+      expect(sqs_msg).to receive(:become_available_at).and_return(@ended_at - 30).twice
 
       expect(Shoryuken.logger).to receive(:info) do |&block|
         expect(block.call).to match(/started at/)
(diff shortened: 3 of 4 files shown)


moyuanhuang:
@phstc Feel free to take a look!

The implementation is a little weird since I had to let `Shoryuken::Message` to manage the `visibility_timeout` attribute since it's not available on `Aws::SQS::Message` (I'm not sure why they to provide such an interface).

I also made the `Timing` middleware to use `sqs_msg.become_available_at` which will be always populated. This eliminates the need of falling back to queue level `visibility_timeout`.

lib/shoryuken/middleware/server/timing.rb
```diff
@@ -4,17 +4,20 @@ module Server
...
 
           yield
 
-          total_time = elapsed(started_at)
+          ended_at = Time.now
+          total_time = (ended_at - started_at) * 1000
 
-          if (total_time / 1000.0) > (timeout = Shoryuken::Client.queues(queue).visibility_timeout)
-            logger.warn { "exceeded the queue visibility timeout by #{total_time - (timeout * 1000)} ms" }
+          if ended_at > sqs_msg.become_available_at
```
phstc: @moyuanhuang isn't this still using the queue visibility instead of the message specific visibility timeout? 

This `Shoryuken::Client.queues(queue).visibility_timeout` will give the queue visibility, not the visibility for the specific message.

moyuanhuang: The `sqs_msg.become_available_at` is initialized using that value, and thus will remain the same with the queue level visibility timeout until `sqs_msg.visibility_timeout(timeout)` is called. Within that method call, the `become_available_at` will be set to `Time.now + timeout`. ([code])

[code]: https://github.com/phstc/shoryuken/pull/592/files/57c6640f2ff7e03b0a69ea25aaaebd59b13904fe#diff-eb88953e947b8431ea88b452a873eac8R36

moyuanhuang: @phstc wanna take another look?

phstc: Sure! Sorry for taking longer on this, I will 👀 over the weekend. But makes sense.

I missed the setting part in the code diff, thanks for pointing that out.

moyuanhuang:
It just occurred to me that `AutoExtendVisibility` is not working correctly too once a message level visibility timeout is set. I'm happy to open another issue to talk more details into that and discuss how we can resolve it!

phstc:
Hi @moyuanhuang 

Could you clarify the issue with `AutoExtendVisibility`?

lib/shoryuken/message.rb
```diff
@@ -28,6 +32,8 @@ def visibility_timeout=(timeout)
         receipt_handle: data.receipt_handle,
         visibility_timeout: timeout
       )
+      @visibility_timeout = timeout
+      self.become_available_at = Time.now + timeout
     end
```
phstc: @moyuanhuang WDYT about just having a getter for `visibility_timeout` instead of introducing a new accessor  `become_available_at`?


```ruby
def visibility_timeout
  @visibility_timeout ||= Shoryuken::Client.queues(queue).visibility_timeout
end

def visibility_timeout=(timeout)
  client.change_message_visibility(
    queue_url: queue_url,
    receipt_handle: data.receipt_handle,
    visibility_timeout: timeout
  )
  @visibility_timeout = timeout
end
```

In that way, `visibility_timeout` would also be lazy-loaded. So we wouldn't call `Shoryuken::Client.queues(queue).visibility_timeout` unless `Message#visibility_timeout` gets called.

lib/shoryuken/middleware/server/timing.rb
phstc: If we go with the suggested change of only introducing a getter `Message#visibility_timeout` the only diff will be:

```ruby
# from this
# (timeout = Shoryuken::Client.queues(queue).visibility_timeout)
# to

(timeout = sqs_msg.visibility_timeout)
```

Right?

lib/shoryuken/message.rb
moyuanhuang: Sorry for the late reply! Got distracted by some other stuffs.

According to the [aws-doc], the visibility timeout, whenever set, is relevant to the time the `ChangeMessageVisibility` request is issued.

Here, the `become_available_at` stands for the **timestamp** when the message becomes available, it is computed according to

```
Time.now + timeout
```

...where `timeout` is the visibility timeout **duration** passed into this method.

This being said, I think in the `Timing` class we should either compare
- `ended_at` (a timestamp) with `sqs_msg.become_available_at` (also a timestamp), or
- `total_time / 1000.0` (a duration) with `sqs_msg.visibility_timeout - sqs_msg.visibility_timeout_reset_at` (also a duration), where `visibility_timeout_reset_at` would be an extra attribute on `Message` that we keep track of.

[aws-doc]:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#changing-message-visibility-timeout

moyuanhuang:
> It warns that the message is exceeding the queue visibility timeout by X amount the time.
It tells the amount exceeded based on the queue visibility, not the amount exceeded since the last visibility bump.

I think this is true, but from my own developing experience, I was confused since after resetting the visibility timeout for a message, I was still given the warning, which made me thinking that the message was retried.

My expectation is that this warning's purpose is to remind developer that the message might be retried. And hence I'd love to see it showing up depending on the messages' actual visibility timeout.

I think the confusion is caused by the word "**queue** level" is easy to be ignored by users. So taking a step back, maybe an alternative is to be more explicit in the message, by adding something like "Message might not be retried if its visibility timeout has been manually bumped". WDYT?

lib/shoryuken/message.rb
phstc: Right, what I was trying to suggest is a code change to minimize the diff in this pull request and not introducing a new accessor that is not used in SQS itself. 

The change I suggested would introduce minimum changes in `message.rb` and also in `timing.rb` as commented in [here](https://github.com/phstc/shoryuken/pull/592/files#r365613536).

phstc:
> So taking a step back, maybe an alternative is to be more explicit in the message, by adding something like "Message might not be retried if its visibility timeout has been manually bumped". WDYT?

[The auto visibility timeout also prints out debug](https://github.com/phstc/shoryuken/blob/af6ba758767aa7e541cae0ad3e4b146ef8eb7792/lib/shoryuken/middleware/server/auto_extend_visibility.rb#L33-L35), so it probably does not help much on giving the full picture of what happened.  

Just to explain it better, I love getting contributions ❤️ But since shoryuken is being on for a while, we need to be careful with changes. I understand your explanation, but so far, no one else has raised this concern. 

That's the reason I'm trying to think more about it. /me is still thinking

lib/shoryuken/message.rb
phstc: reading better what you said, I missed `Time.now + timeout`, we need to keep accumulating the total visibility timeout.

phstc: Is there a way we can take this `@visibility_timeout = Shoryuken::Client.queues(queue).visibility_timeout` from the initializer and lazy-load it?

moyuanhuang:
@phstc Sure! Take your time, if we decided that this is not very helpful to developers we can close the PR and reopen it if people ever find it  useful. I'll try to find sometime this week to cleanup the code change regardless. Thanks ❤️ ❤️ ❤️

phstc:
>  I'll try to find sometime this week to cleanup the code change regardless

Thank you ❤️ and sorry for the back-and-forth. 

Have a great week!

PR #597, opened 2020-02-29
Allow custom loggers to use with_context

allcentury:
Prior to this change anyone who used a custom logger likely missed out
on anything set in Shoryuken::Logging.with_context.

This PR allows loggers who implement:

- #add_context
- #clear_context!

to take advantage of any context set via .with_context.

This PR also adds the ability to set nested context's, prior to this
only one context could be set.

Lastly, I added the exception class name raised in Shoryuken::Processor#process, we were
expecting to see logs for `ActiveJob::DeserializationError` but because
of some configuration issues this didn't happen.  I thought others would
also benefit from this since this is default behavior with ruby's built
in logger https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html#method-i-add

### DIFF
--- a/lib/shoryuken/logging.rb
+++ b/lib/shoryuken/logging.rb
@@ -16,10 +16,19 @@ module Shoryuken
     end
 
     def self.with_context(msg)
-      Thread.current[:shoryuken_context] = msg
+      if logger.respond_to?(:add_context)
+        logger.add_context(msg)
+      else
+        Thread.current[:shoryuken_context] ||= []
+        Thread.current[:shoryuken_context] << msg
+      end
       yield
     ensure
-      Thread.current[:shoryuken_context] = nil
+      if logger.respond_to?(:clear_context!)
+        logger.clear_context!
+      else
+        Thread.current[:shoryuken_context] = nil
+      end
     end
 
     def self.initialize_logger(log_target = STDOUT)
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -22,7 +22,7 @@ module Shoryuken
         end
       end
     rescue Exception => ex
-      logger.error { "Processor failed: Exception: #{ex.name} with message:#{ex.message}" }
+      logger.error { "Processor failed: Exception: #{ex.class.name} with message:#{ex.message}" }
       logger.error { ex.backtrace.join("\n") } unless ex.backtrace.nil?
 
       raise
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -22,7 +22,7 @@ module Shoryuken
         end
       end
     rescue Exception => ex
-      logger.error { "Processor failed: #{ex.message}" }
+      logger.error { "Processor failed: Exception: #{ex.name} with message:#{ex.message}" }
       logger.error { ex.backtrace.join("\n") } unless ex.backtrace.nil?
 
       raise
--- a/spec/shoryuken/logging_spec.rb
+++ b/spec/shoryuken/logging_spec.rb
@@ -1,0 +1,98 @@
+require "spec_helper"
+
+RSpec.describe Shoryuken::Logging do
+  class self::MyLogger < Logger
+    def initialize(output)
+      super(output)
+      @context = []
+    end
+
+    def info(msg)
+      super(@context.join(" ") + " " + msg)
+    end
+
+    def add_context(msg)
+      @context << msg
+    end
+
+    def clear_context
+      @context = nil
+    end
+  end
+  let(:output) do
+    StringIO.new
+  end
+  context ".with_context" do
+    context "with customer logger" do
+      let(:formatter) do
+        double('my_formatter')
+      end
+      let(:custom_logger) do
+        self.class::MyLogger.new(output)
+      end
+      before(:each) do
+        Shoryuken::Logging.logger = custom_logger
+      end
+      it "can attach context if logger implements #update_log_content" do
+        my_context = "my_context"
+        my_msg = "My log msg"
+        Shoryuken::Logging.with_context(my_context) do
+          custom_logger.info my_msg
+        end
+
+        output.rewind
+        msg = output.read
+        expect(msg).to include(my_context)
+        expect(my_msg).to include(my_msg)
+      end
+
+      it "cleans up via #clear_context" do
+        expect(custom_logger).to receive(:clear_context!)
+        Shoryuken::Logging.with_context("context") do
+          custom_logger.info "some msg"
+        end
+      end
+    end
+
+    context "built in logger" do
+      it "appends the context" do
+        my_context = "my_context"
+        my_msg = "My log msg"
+        Shoryuken::Logging.initialize_logger(output)
+        Shoryuken::Logging.with_context(my_context) do
+          Shoryuken::Logging.logger.info my_msg
+        end
+
+        output.rewind
+        msg = output.read
+        expect(msg).to include(my_context)
+        expect(my_msg).to include(my_msg)
+      end
+
+      it "lets you nest context" do
+        my_context = "my_context"
+        my_context2 = "aother context"
+        my_msg = "My log msg"
+        Shoryuken::Logging.initialize_logger(output)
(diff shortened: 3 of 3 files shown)


allcentury:
What do you think about this @phstc ?

lib/shoryuken/logging.rb
```diff
@@ -16,10 +16,19 @@ def context
     end
 
     def self.with_context(msg)
-      Thread.current[:shoryuken_context] = msg
+      if logger.respond_to?(:add_context)
+        logger.add_context(msg)
```
phstc, line 20: 👋 

What's the logger impl that supports `add_context`?

allcentury: We have an inherited `ActiveSupport::TaggedLogging` logger that we will add the custom method to.

phstc: I'm wondering if that's specific for your custom method. Therefore it may not be generic enough to have it in Shoryuken.  WDYT?

If that's the case (too specific), you can extend the current logger, adding your specific code, then change the Shoryuken default logger. It is an option [here](https://github.com/phstc/shoryuken/blob/15923f8b02793c412dc430e16f1f4742d9e1f471/lib/shoryuken/options.rb#L89).

allcentury: we set a logger in our initializer, `Shoryuken::Logging.logger = Rails.logger` - this logger adds a lot of observability goodies for us and I imagine other companies who have multiple apps running ruby do the same.  

I think either shoryuken should establish an interface (like this PR suggests) for this added context or we remove it in favor of in-lining the context somehow. 

As it stands, we didn't even know this was available and while we'd prefer not to reach into `Thread.current`, I don't see a good way to do this without an interface.  I'm open to other suggestions if you have 'em

phstc: Hi @allcentury 

How about nesting `with_context`? so we don't need to introduce a new interface?

```ruby
Shoryuken::Logging.with_context(msg_a) do 
  # ...
  Shoryuken::Logging.with_context(msg_b) do 
    # ...
  end
end
```

```ruby
def self.with_context(msg)
  Thread.current[:shoryuken_context] ||= []
  Thread.current[:shoryuken_context] << msg

  yield
ensure
  Thread.current[:shoryuken_context].pop
end
```

I have mixed feelings with `add_context`. I'm wondering if this is for error tracking if so, I believe that maybe error trackers could be more appropriated. [Honeybadger](https://github.com/honeybadger-io/honeybadger-ruby/blob/master/lib/honeybadger/plugins/shoryuken.rb) and [Rollbar](https://docs.rollbar.com/docs/shoryuken-integration) have built-in integrations with Shoryuken and both have methods for adding contexts. 

Just so I can get more context on the use-case, how are you using this logging?

lib/shoryuken/processor.rb
```diff
@@ -22,7 +22,7 @@ def process
         end
       end
     rescue Exception => ex
-      logger.error { "Processor failed: #{ex.message}" }
+      logger.error { "Processor failed: Exception: #{ex.class.name} with message:#{ex.message}" }
```
phstc, line 25: That might be related to the error tracking I mentioned above, I would strongly recommend an error tracking for tracking errors. The `raise` below would eventually be logged in an error tracking, along with the error message and message body. 

I'm not against adding `Exception:`  though, but if we do it in here, we may want to add it [here](https://github.com/phstc/shoryuken/blob/c7a0984fbb6811a307b5c062047ad1cacbfd41b9/lib/shoryuken/manager.rb#L104) and [here](https://github.com/phstc/shoryuken/blob/bc7938942270a1171d1597064698d88c3f36a4ec/lib/shoryuken/middleware/server/exponential_backoff_retry.rb#L26) too - all the places that `ex.backtrace.join("\n")`.

phstc:
> also benefit from this since this is default behavior with ruby's built
in logger https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html#method-i-add

@allcentury Could you give me an example of that? Is `Logger` adding the exception class name?

lib/shoryuken/logging.rb
allcentury: I understand your mixed feelings, is there a way to remove `Shoryuken::Logging.with_context` so there is no special ❄️ for `Shoryuken::Logging` vs using a different logger?

The use case is we don't want to miss out on valuable log lines because of which logger Shoryken is using.

phstc: There's only one reference in the code for [`with_context`](https://github.com/phstc/shoryuken/blob/bc7938942270a1171d1597064698d88c3f36a4ec/lib/shoryuken/processor.rb#L19).

We have a few options:

1. In your workers add a `rescue` and use your own logging.

```ruby
def perform
rescue => e
  # your own logging
  raise
end
```

2. Create a middleware to ensure all your worker failures will be logged in the same way

```ruby
def call(_worker, queue, _sqs_msg, _body)
  custom_with_context(...) do
    yield
  end
rescue
  # moar logging
  raise
end
```

3. Use [my example](https://github.com/phstc/shoryuken/pull/597#discussion_r407112579) for nesting `with_context`. Wouldn't this work?

4. Make `Logging` a Shoryuken `Option`.

allcentury: 1.  That could work but there are obvious tradeoffs there
2.  This is what we're doing today
3.  I don't grok the 3rd option, the nesting doesn't really change that the log formatter has this special method called `context` that knows to reach into a thread local variable https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/logging.rb#L13 .  I'm trying to come up with a way to make that generic so any logger could store or pass that valuable context information as needed.  
4. Not sure what you're thinking here

phstc: 3. 

With my change suggestion, this 👇 would possible.

```ruby
Shoryuken::Logging.with_context(...) do # internal from the `processor.rb`

   # Somewhere in your  code
   Shoryuken::Logging.with_context(...) do
      # this now will have the context added by the `process.rb` + the added context above
   end
end
```
Acting similarly to your `add_context`.

```ruby
Shoryuken::Logging.with_context(...) do # internal from the `processor.rb`

   # Somewhere in your  code
   Shoryuken::Logging.add_context(...)
   # this now will have the context added by the `process.rb` + the added context above
end
```

4. 

Another idea is to make `Shorykuen::Logging` customizable. So you can add your implementation that calls `add_context` or any other specifics to your use-case. 

You could:

<details>
  <summary>Sample code</summary>

```ruby
module Shoryuken
  module Logging
    def self.with_context(*args)
      Shoryuken.logging_handler.with_context(*args)
    end

    def self.initialize_logger(*args)
      Shoryuken.logging_handler.initialize_logger(*args)
    end

    def self.logger
      Shoryuken.logging_handler.logger
    end

    def self.logger=(log)
      Shoryuken.logging_handler.logger = log
    end
  end
end

module Shoryuken
  module DefaultLoggingHandler
    class Pretty < Logger::Formatter
      # Provide a call() method that returns the formatted message.
      def call(severity, time, _program_name, message)
        "#{time.utc.iso8601} #{Process.pid} TID-#{Thread.current.object_id.to_s(36)}#{context} #{severity}: #{message}\n"
      end

      def context
        c = Thread.current[:shoryuken_context]
        c ? " #{c}" : ''
      end
    end

    def self.with_context(msg)
      Thread.current[:shoryuken_context] = msg
      yield
    ensure
      Thread.current[:shoryuken_context] = nil
    end

    def self.initialize_logger(log_target = STDOUT)
      @logger = Logger.new(log_target)
      @logger.level = Logger::INFO
      @logger.formatter = Pretty.new
      @logger
    end

    def self.logger
      @logger || initialize_logger
    end

    def self.logger=(log)
      @logger = (log ? log : Logger.new('/dev/null'))
    end
  end
end

module Shoryuken
  class Options
    def logging_handler
      @logging_handler || DefaultLoggingHandler
    end
  end
end
```
</details>

☝️ with that, you could `Shoryuken.logging_handler = YourLoggingHandlerWithAddContext`.

PR #598, opened 2020-03-17
Add a Fake Executor

colemanja91:
This is parallel with `Sidekiq::Testing.fake!', and is especially helpful if you are testing job logic
separately from where the job is enqueued.

### DIFF
--- a/lib/shoryuken/worker/fake_executor.rb
+++ b/lib/shoryuken/worker/fake_executor.rb
@@ -1,0 +1,28 @@
+module Shoryuken
+  module Worker
+    class FakeExecutor
+      class << self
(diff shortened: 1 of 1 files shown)


lib/shoryuken/worker/fake_executor.rb
```diff
@@ -0,0 +1,28 @@
...
+          )
+        end
+
+        def perform_in(worker_class, _interval, body, options = {})
+          worker_class.perform_async(body, options)
+        end
+      end
+    end
+  end
+end
```
phstc, line 28: Hi @colemanja91 

How would `FakeExecutor` be used? Do you have a code snippet example?

Issue #602, opened 2020-05-22
request: allow queues in shoryuken.yml to be provided as ARN values


nippysaurus:
I'm writing a CDK template for deploying an application with a lot of Shoryuken queues. Everything in AWS is using the ARN, but Shoryuken only works with URL and not ARN. These values are easy translated from one format to the other, but with so many queues it would still be a lot of mapping values in my code to do the translation. It would be super convenient if Shoryuken would do this for me.

[Here you can see how the AWS CDK template does the mapping](https://github.com/aws/aws-cdk/blob/master/packages/%40aws-cdk/aws-sqs/lib/queue.ts#L200)

phstc:
Hi @nippysaurus 

[Shoryuken can work with queue name or URL](https://github.com/phstc/shoryuken/blob/3ed8eb10ff59cc0fa81389890d13c826d45ae3a4/lib/shoryuken/queue.rb#L69).

[You can get `queueUrl` and `queueName` with CDK](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-sqs.Queue.html#queuename-1).

You could also output then.

```typescript
new cdk.CfnOutput(this, "QUEUE_URL", { value: queue.queueURL });
new cdk.CfnOutput(this, "QUEUE_NAME", { value: queue.queueName });

```

[The AWS SDK Ruby expects `queue_url` for interacting with a queue](https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/SQS/Client.html#send_message-instance_method), so the conversion is required. 


There are also some alternatives, for example, converting from ARN to URL or Name in your `shoryuken.yml`, it is an ERB. You can add code in there. Also, instead of adding queues via the `shoryuken.yml` you could add them programmatically [`Shoryuken.add_queue(...)`](https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/options.rb#L48) in an initializer. 

I don't recall previous requests for that, but I'm open to pull requests if the logic makes sense in case you are interested in submitting one.

nippysaurus:
Yeah it's easy to do the mapping, but with the URL and ARN being so easily translated it would be nice to be able to supply either and have Shoryuken figure out what it is.

I took your advice and did the mapping in the config.

```ruby
# config/initializers/shoryuken.rb
def SqsArnToUrl(arn_str)
  return unless arn_str

  arn = Aws::ARNParser.parse(arn_str)

[...]

PR #603, opened 2020-05-26
add ability to configure queue by ARN

nippysaurus:
This is my first pass at adding the ability to set queue location as an ARN value.

I removed use of the AWS parser as it violates the `'aws-sdk-core', '>= 2'` dependency (it was only added recently) and I don't think it was intended for outside use anyway.

Related: #602

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -65,12 +65,28 @@ module Shoryuken
       self.url  = url
     end
 
+    def arn_to_url(arn_str)
+      _,_,_,region,account_id,resource = arn_str.split(":")
+
+      required = [region, account_id, resource]
+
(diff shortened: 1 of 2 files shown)


nippysaurus:
I'm not entirely sure how you would like me to resolve this codeclimate feedback.

IMHO the concept/logic for processing the config input has probably out-grown the Queue class an needs to be split out. That would allow for more thorough testing of that code as well because it can be done in isolation.

phstc:
Hi @nippysaurus 

Thank you. Shoryuken 5.0.5 is out including your changes.

* https://github.com/phstc/shoryuken/wiki/Shoryuken-options#queues
* https://github.com/phstc/shoryuken/wiki/Multiple-ways-of-configuring-queues-in-Shoryuken

PR #607, opened 2020-06-06
Add behaviour to register_worker

andreleoni:
# Problem

We will have some Jobs on our application that have the same queue.

Today, if no one queue come on sqs message, the application will get the last job "registered". I think that it behaviour should be more intelligent, so, I changed to register the last dispatcher job.

For example:

### Job A:

```
class AJob
  include Shoryuken::Worker

  shoryuken_options queue: :development_default, auto_delete: true

  def perform(_, args)
    puts "executed_job_a"
  end
end
```

### Job B:

```
class BJob
  include Shoryuken::Worker

  shoryuken_options queue: :development_default, auto_delete: true

  def perform(_, args)
    puts "executed_job_b"
  end
end
```

### So, my dispatcher should be

```
class DispatcherJob
  inclsude Shoryuken::Worker

  shoryuken_options queue: :development_default, auto_delete: true, dispatcher: true, body_parser: true

  def perform(_, sqs_msg)
    if sqs_msg['action'] == 'create'
      JobA.perform_async(sqs_msg)
    elsif sqs_msg['action'] == 'delete'
      JobB.perform_async(sqs_msg)
    end
  end
end
```

So, I can receive, for example, a message from SNS to my queue, and threat in a dispatcher job to choose the destination.

I think that is a generic and useful way to solve our problem.

# Proposed changes

When have a dispatcher worker, he will have priority on register. If a worker already registered, only will be overrided if another worker be a dispatcher.

### DIFF
--- a/lib/shoryuken/default_worker_registry.rb
+++ b/lib/shoryuken/default_worker_registry.rb
@@ -40,7 +40,7 @@ module Shoryuken
         end
       end
 
-      @workers[queue] = clazz
+      @workers[queue] = clazz if !@workers[queue] || clazz.get_shoryuken_options['dispatcher']
     end
 
     def workers(queue)
--- a/lib/shoryuken/default_worker_registry.rb
+++ b/lib/shoryuken/default_worker_registry.rb
@@ -32,15 +32,20 @@ module Shoryuken
     end
 
     def register_worker(queue, clazz)
-      if (worker_class = @workers[queue])
+      worker_class = @workers[queue]
+
+      if worker_class
         if worker_class.get_shoryuken_options['batch'] == true || clazz.get_shoryuken_options['batch'] == true
           fail ArgumentError, "Could not register #{clazz} for #{queue}, "\
             "because #{worker_class} is already registered for this queue, "\
             "and Shoryuken doesn't support a batchable worker for a queue with multiple workers"
+        elsif worker_class.get_shoryuken_options['dispatcher'] == true && clazz.get_shoryuken_options['dispatcher'] == true
+          fail ArgumentError, "Could not register #{clazz} for #{queue}, "\
+            "because #{queue} queue can have only one dispatcher job for queue and #{worker_class} was already registered."
         end
       end
 
-      @workers[queue] = clazz if !@workers[queue] || clazz.get_shoryuken_options['dispatcher']
+      @workers[queue] = clazz if !worker_class || clazz.get_shoryuken_options['dispatcher']
     end
 
     def workers(queue)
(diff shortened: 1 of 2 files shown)


phstc:
hi @andreleoni 

Thanks for your contribution 🎉 

For this approach, I would recommend you to use two queues. I believe that's more the SQS way.

For the fanout you are doing, you could also use [event bridge](https://aws.amazon.com/eventbridge/), so you don't need to worry about priority or even implementing the dispatcher yourself in case of it is only checking `sqs_msg['action']`.

andreleoni:
hello @phstc, the pleasure to contribute is mine. thank you for the excellently created gem and your time to maintain it.

I think that create another queue for each SNS, should solve our problem for now, but, only some considerations about our structure...

We are thinking in the "sidekiq" mindset usage, for example, have queues named like `critical, low, default`... And another specific queue to payment, shipment monitoring, and another major feature... So, I will have many jobs in the same queue... (Like the example)

If I have a class only for each one SNS job, and some problem occurs in that queue, we should create a validation on our RailsApp to reschedule to correct queue, because that queue is only to see the SNS jobs. In the example, I used only the example of `perform_async`, so, to not reschedule the job, like is the same queue, should be a `.new.perform`...

With this concept, I should have a job as a fallback, to avoid the problem with a service send a job without a `shoryuken_class` attribute on SQS message. (a message sent from another source for example)

However, my usage will be out from "Rails app" (systems integration), but, I believe that, as I have an "automatic queue fallback register" with the last file that loaded if I choose which file will be, is a good feature (like my dispatcher example)

Does should be that "dispatcher" is a bad option name? What do you think about it? 

I will talk with my company team to talk about a "single job queue" solution.

Thank you again for your time!

phstc:
> I think that create another queue for each SNS, should solve our problem for now, but, only some considerations about our structure...

You can also create a single queue for each SNS or a single SNS for multiple queues. If you use Event Bridge instead of SNS, you can also filter based on the message content, as you did programmatically. 

> We are thinking in the "sidekiq" mindset usage, for example, have queues named like critical, low, default... And another specific queue to payment, shipment monitoring, and another major feature... So, I will have many jobs in the same queue... (Like the example)

Shoryuken also supports [queues with different priorities](https://github.com/phstc/shoryuken/wiki/Shoryuken-options#queues), also [processing groups](https://github.com/phstc/shoryuken/wiki/Processing-Groups). You could also check [custom worker registry](https://github.com/phstc/shoryuken/wiki/Worker-registry).

> Does should be that "dispatcher" is a bad option name? What do you think about it?

It is not about the naming. It is more about having other ways of achieving that, which I believe could be better. Also, I would rather take advantage of the other ways of doing that than adding more. Let's say specific-complexity to Shoryuken, which, once added, needs to be maintained. 

I will be happy to give more guidance on the above, as is, I don't think it should be merged.

andreleoni:
Thank you for the answer!

PR #611, opened 2020-06-10
Allow middleware to modify arguments passed to worker

mfly:
I really like the idea of middlewares as they make extending `shoryuken` pretty straightforward. Unfortunately there are still some limitations.
I'm working on a middleware that implements the SQS Extended Client which solves the problem of messages bigger than 256KB (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-s3-messages.html).
The main problem is that the current implementation does not allow for a middleware to modify body of the message that is passed to worker.
It is possible to use a custom `body_parser` for this purpose but in my opinion message retrieval and deserialisation are two different things. The message body stored in S3 may also be encoded as json or any other format, so it seems justified to leave the deserialisation logic as-is.

This PR allows server middlewares to alter the arguments passed to the next processing step which solves my use case. I included some additional logic to make sure that it does not break existing middlewares. If you find it useful, I'll prepare a similar PR for the client middleware implementation.


hex337:
Hello! I am working on a project that's using Shoryuken under the hood, and we are adding our own extensions on top, but as part of that, we want to handle the de-serialization in middleware so that when someone extends the worker class we have defined, the `.body` contains the deserialized structure (we may also want to remove metadata that we use for stats tracking).

We can't rely on a wrapper in this case, since that pushes the responsibility to the end user, who shouldn't care what transport protocol we use behind the scenes.

Right now, the only way I can do that is by creating my own worker executor, which seems overkill.

cjlarose:
@hex337 I think for your use case, you might want to consider doing something like the ActiveJob adapter. https://github.com/ruby-shoryuken/shoryuken/blob/17ff4459facf8b7398dfadd2195ac25936255327/lib/shoryuken/extensions/active_job_adapter.rb#L80-L90

The ActiveJob adapter for shoryuken has basically the same problem you have: users are using a particular interface and don't care about the transport mechanism. The way the ActiveJob adapter for shoryuken works is by defining a custom worker `ActiveJob::QueueAdapters::ShoryukenAdapter::JobWrapper`, which deserializes the actual message body and passed the deserialized version onto the user's worker code.

lib/shoryuken/middleware/chain.rb
```diff
@@ -93,11 +93,13 @@ def clear
 
       def invoke(*args, &final_action)
         chain = retrieve.dup
-        traverse_chain = lambda do
+        traverse_chain = lambda do |*modified_args|
+          next_args = modified_args.any? ? modified_args : args
```
cjlarose, line 97: I like the overall idea here and I like especially how this is backwards-compatible with existing middleware that just calls `yield`. The implementation, however, has an interesting quirk that I'm not sure was intended: Suppose you have three middleware A, B, and C. A modifes the args, but B does not. When it's time to invoke `C`, `modified_args` is empty (B didn't yield anything), so `C` receives the original top-level `args`. Instead, it probably makes more sense for `C` to receive the args as they were passed to B (the same args that A `yield`ed).

cjlarose (changes requested):

PR #617, opened 2020-07-27
Update logging.rb

### LINKED ISSUE #616: warning: instance variable @logger not initialized
This message is printed when running with warnings.

```
shoryuken-5.0.4/lib/shoryuken/logging.rb:33: warning: instance variable @logger not initialized
```

https://github.com/phstc/shoryuken/blob/7ce3e7540375798866d3bde482f3869bb6eee6df/lib/shoryuken/logging.rb#L33

:pushpin: #617

ioquatix:
Fixes #616.

### DIFF
--- a/lib/shoryuken/logging.rb
+++ b/lib/shoryuken/logging.rb
@@ -30,7 +30,11 @@ module Shoryuken
     end
 
     def self.logger
-      @logger || initialize_logger
+      unless defined? @logger
+        @logger = initialize_logger
+      end
+      
+      @logger
     end
 
     def self.logger=(log)


lib/shoryuken/logging.rb
```diff
@@ -30,7 +30,11 @@ def self.initialize_logger(log_target = STDOUT)
     end
 
     def self.logger
-      @logger || initialize_logger
+      unless defined? @logger
+        @logger = initialize_logger
+      end
```
phstc, line 35: In which version of Ruby do you see this warning?

In that case, if it is defined, but `nil`, it won't `initialize_logger `. That may change the functionality, we may need to check defined and nil (presence) to keep it as is. 

WDYT?

ioquatix: All versions of Ruby 2.x - yes if someone may initialise it to nil you may need extra checks.

ioquatix:
Was this issue ever addressed?

phstc:
@ioquatix sorry, this got auto closed when `master` was "renamed" to `main`. After deleting `master`, GitHub automatically closed opened pull requests targeting `master`.

How can I reproduce https://github.com/ruby-shoryuken/shoryuken/issues/616 so I can test this pull request?

Thanks for checking!

PR #619, opened 2020-07-28
feat: Modify add_groups to accept polling_strategy

hkrdrm:
- modify add_groups add polling_strategy param
  - modify polling_strategy to look use groups if options[:groups].nil?

### DIFF
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -34,14 +34,16 @@ module Shoryuken
       defined?(::ActiveJob)
     end
 
-    def add_group(group, concurrency = nil, delay: nil)
+    def add_group(group, concurrency = nil, delay: nil, polling_strategy: 'WeightedRoundRobin')
       concurrency ||= options[:concurrency]
       delay ||= options[:delay]
+      polling_strategy ||= polling_strategy
 
       groups[group] ||= {
         concurrency: concurrency,
         delay: delay,
-        queues: []
+        queues: [],
+        polling_strategy: polling_strategy
       }
     end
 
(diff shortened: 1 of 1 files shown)


lib/shoryuken/options.rb
```diff
@@ -56,7 +58,12 @@ def ungrouped_queues
     end
 
     def polling_strategy(group)
-      strategy = (group == 'default' ? options : options[:groups].to_h[group]).to_h[:polling_strategy]
+      if options[:groups].nil?
+        strategy = (group == 'default' ? options : groups.to_h[group]).to_h[:polling_strategy]
+      else
+        strategy = (group == 'default' ? options : options[:groups].to_h[group]).to_h[:polling_strategy]
+      end
```
phstc, line 65: hi @hkrdrm, thanks for the contribution. That's very welcome. 

WDYT about having an `if` for `default`? 

```ruby
strategy = \
  if group == 'default'
    options.to_h[:polling_strategy]
  elsif options[:groups].nil?
    # uses custom polling_strategy for the given group
    groups.to_h[group].to_h[:polling_strategy]
  else
    # uses default strategy
    options[:groups].to_h[group].to_h[:polling_strategy]
  end
```

Can you check if the comment makes sense? Late hours, I'm not sure if I got them right.

hkrdrm: Yeah that simplifies things and looks good to me.  On the comments for the else block I'd probably say its checking the option hash for polling strategy since its not necessarily default just whatever is coming in on the options hash.  I'll amend the commit and push it up.

PR #627, opened 2020-08-22
Add Rails reloading support

cjlarose:
This change adds explicit integration for Rails autoloading. Rails autoloading already partially worked in Shoryuken in that constants that weren't eager loaded were automatically autoloaded. Shoryuken, however, didn't respect `config.cache_classes`. This meant that once a constant was loaded by the Shoryuken process, it would never be unloaded and reloaded. If you changed the code in a worker class after Shoryuken had loaded it, the change would have no effect until you restarted the Shoryuken process.

The proposed change uses `Rails.application.reloader`, which takes care of safely reloading constants. This is the same technique employed by [active_job][0] and [sidekiq][1] and even explicitly documented in the [Rails guides][2]. The reloader takes the appropriate locks to ensure that no two threads ever simultaneously unload or reload constants. If the application developer modifies code in an autoloaded directory (including the implementation of workers themselves in `app/workers`), the changes will take effect the next time a job is processed.

[0]: https://github.com/rails/rails/blob/ab4ab090667e96c0e8a2623e33573ae06be14b54/activejob/lib/active_job/railtie.rb#L42-L44
[1]: https://github.com/mperham/sidekiq/blob/460ce532714444e186a60d49c9251b93909d4a58/lib/sidekiq/rails.rb#L13-L15
[2]: https://github.com/rails/rails/blob/ab4ab090667e96c0e8a2623e33573ae06be14b54/guides/source/threading_and_code_execution.md#reloader

## Breaking changes

This change probably warrants a major version bump. The reason for that is that if Shoryuken library code has any reference to objects that belong to autoloaded modules, it's quite easy to run into autoloading errors. If other people's apps are like my own, this is most likely to happen with custom middleware. This is an example from a Rails initializer:

```ruby
    chain.add MyCustomMiddleware
  end
end
```

Assume `MyCustomMiddleware` is defined in an autoloaded directory (basically anything under `app`, in addition to directories explicitly added to `config.autoload_paths`), and that the execution of `MyCustomMiddleware#call` has unqualified references to autoloaded constants. In this case, if Shoryuken tries to invoke that middleware after the application developer has triggered reloading (by modifying files in autoloaded directories), then autoloading will fail with `A copy of MyCustomMiddleware has been removed from the module tree but is still active!`. 

The fix is to ensure that Shoryuken library code never retains a reference to any object defined in autoloaded modules and whose code has references to autoloaded modules. In the example of custom middleware, the easiest solution is to just attach the middleware to workers instead of the global `server_middleware`. You can even do this in your base worker class.

```ruby
class ApplicationWorker
  include Shoryuken::Worker

  def self.inherited(klass)
    klass.server_middleware do |chain|
      chain.add MyCustomMiddleware
    end
  end
end
```

This works because the worker-specific middleware chain has a lifetime associated with the worker class. Once unloading happens, the old references to `MyCustomMiddleware` are unreachable by the Shoryuken process. When a new job comes in, Shoryuken autoloads the worker class, and the `inherited` block will run, constructing a new instance of the `MyCustomMiddleware` class through autoloading.

Users upgrading to a version of Shoryuken that includes this change can, in theory, alternatively just turn on `config.cache_classes` in their Shoryuken server process (even in development). The experience will be the same as before in that classes won't be reloaded, but it might be an easier upgrade path for some.

```ruby
Shoryuken.configure_server do
  Rails.application.configure do
    config.cache_classes = true
  end
end
```

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -69,6 +69,8 @@ module Shoryuken
     :sqs_client=,
     :sqs_client_receive_message_opts,
     :sqs_client_receive_message_opts=,
+    :reloader,
+    :reloader=,
     :options,
     :logger,
     :register_worker,
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -70,6 +70,13 @@ module Shoryuken
             ::Rails.application.config.eager_load = true
           end
         end
+        ::Rails::Application.initializer 'shoryuken.set_reloader_hook' do |app|
+          Shoryuken.reloader = proc do |&block|
+            app.reloader.wrap do
+              block.call
+            end
+          end
+        end
         require 'shoryuken/extensions/active_job_adapter' if Shoryuken.active_job?
         require File.expand_path('config/environment.rb')
       end
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -15,7 +15,7 @@ module Shoryuken
     }.freeze
 
     attr_accessor :active_job_queue_name_prefixing, :cache_visibility_timeout, :groups,
-                  :launcher_executor,
+                  :launcher_executor, :reloader,
                   :start_callback, :stop_callback, :worker_executor, :worker_registry
     attr_writer :default_worker_options, :sqs_client
     attr_reader :sqs_client_receive_message_opts
@@ -26,6 +26,7 @@ module Shoryuken
       self.active_job_queue_name_prefixing = false
       self.worker_executor = Worker::DefaultExecutor
       self.cache_visibility_timeout = false
+      self.reloader = proc { |&block| block.call }
       # this is needed for keeping backward compatibility
       @sqs_client_receive_message_opts ||= {}
     end
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -14,11 +14,13 @@ module Shoryuken
     end
 
     def process
-      return logger.error { "No worker found for #{queue}" } unless worker
+      Shoryuken.reloader.call do
+        return logger.error { "No worker found for #{queue}" } unless worker
 
-      Shoryuken::Logging.with_context("#{worker_name(worker.class, sqs_msg, body)}/#{queue}/#{sqs_msg.message_id}") do
-        worker.class.server_middleware.invoke(worker, queue, sqs_msg, body) do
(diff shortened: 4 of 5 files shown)


cjlarose:
@phstc Anything I can do to help move this PR along?

phstc:
@cjlarose thank you for this very detailed PR. 

What would be the use case for auto-reloading? How would you use it on your project?

cjlarose:
Auto-reloading is a nice quality-of-life improvement when I'm making changes to the codebase of my application locally. Whenever I modify a Rails controller or model or something, that change is reflected the next time I make a web request to the server (for most changes, I don't need to restart the web server). This happens by default automatically for Rails apps. But since shoryuken doesn't have explicit support for autoloading, it basically means that whatever version of a module or class that shoryuken loaded into memory first, that's the version it'll have for the lifetime of the process. If I want to modify a background job, it means that I have to kill the shoryuken process and restart it. Being able to unload and reload classes when I change them means that I wouldn't have to restart shoryuken and I'd have a tighter feedback loop--it's just one less thing I have to remember to do.

We onboard new engineers a bunch at my company it's always something we have to explain to them: whenever you make changes to the application, those changes are reflected immediately in the web app, but if you want your changes to take effect in the background jobs, you have to restart shoryuken. It's always been a little awkward to work around, so a colleague and I looked into make it work the way we'd like. We started with a monkey-patch for Shoryuken in our own application to do it, and the response was really positive among our devs. We put this PR together so others can enjoy the same benefits.

The other reason why this change makes sense for the application that I work on is that Rails autoloading, by itself, isn't thread-safe. While it's not considered best practice for modern Rails apps to do any autoloading in production, we still have some modules in autoloaded (but not eager loaded) directories. This means that it's possible for two shoryuken threads to simultaneously discover that a constant is missing, then try to autoload the corresponding module, but do so in a way that everything is double-loaded. This can cause errors in production, though they're somewhat mitigated by the fact that jobs will retry automatically if they fail. We've fixed most of these problems by moving more stuff into autoloaded directories, but the more robust fix is to use Rails' `reloader` since that'll make threads all take the appropriate exclusive locks when they're trying to autoload, preventing these kinds of race conditions.

Hope that makes sense! Lemme know if you have other questions!

cjlarose:
My team deployed a patch of Shoryuken that includes this change and we saw some unexpected problems related to memory usage. We're not sure yet if the patch was the cause, but we'll investigate.

PR #628, opened 2020-08-24
Improvements to FIFO queue processing

cstyles:
This PR proposes a way to improve the processing of messages in FIFO queues by removing the "only fetch 1 message at a time" limit while still ensuring that we process messages with a common message group ID in-order. See the commit messages for details.

A couple of notes:
- I'm not very familiar with rspec. I tried to cobble together some basic tests but please let me know if I need to make any changes.
- I originally implemented the "ensure we process messages with a common message group ID in-order" logic using a chain of `Promise`s but later opted to just use a single `Promise` with a loop inside it. I left the original implementation in the git history so feel free to take a look and decide which version you prefer.


spec/shoryuken/fetcher_spec.rb
```diff
@@ -105,12 +105,12 @@
       let(:limit) { 10 }
       let(:queue) { instance_double('Shoryuken::Queue', fifo?: true, name: queue_name) }
 
-      it 'polls one message at a time' do
+      it 'polls the provided limit' do
         # see https://github.com/phstc/shoryuken/pull/530
```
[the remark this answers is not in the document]
cstyles: Good catch! I pushed d6b7d35b0220e1b44e8a8fa85be0813f70fc8dcb which removes the comment. Thanks for the review, kind stranger!

lib/shoryuken/manager.rb
```diff
@@ -61,29 +68,40 @@ def processor_done
...
-      logger.debug { "Assigning #{sqs_msg.message_id}" }
+      logger.debug { "Assigning #{sqs_msgs.map(&:message_id)}" }
 
       @busy_processors.increment
 
-      Concurrent::Promise.execute(
-        executor: @executor
-      ) { Processor.process(queue_name, sqs_msg) }.then { processor_done }.rescue { processor_done }
+      promise = Concurrent::Promise.execute(executor: @executor) do
+        sqs_msgs.each { |sqs_msg| Processor.process(queue_name, sqs_msg) }
```
phstc, line 79: We can't keep the messages in memory to process one at a time. Because the time spent with one message would count against all the other pending messages.

phstc (commented):
@cstyles this is an interesting approach, but I'm worried about keeping messages in memory while one message is being processed. That can cause unexpected side effects, for example, if the pending messages exceed their visibility timeout, then there's a redrive policy and send them to a DL queue.

lib/shoryuken/manager.rb
cstyles: That's a good point. For my particular use case (where the messages are all processed fairly quickly) it's not a concern but I can see how this solution isn't acceptable for all use cases.

Would you be open to making the message fetch limit configurable for FIFO queues so that users can opt in to the behavior implemented in this PR if they're OK with the risk?

cjlarose: If I understand this change correctly, I think users can _already_ opt into very similar behavior by using the `batch` option.

Assuming you have a FIFO queue where every message in the queue uses the same `message_group_id`:

```ruby
class AddWorker
  include Shoryuken::Worker

  shoryuken_options queue: 'cjlarose_shoryuken_test.fifo',
                    batch: true,
                    body_parser: :json,
                    auto_delete: true

  def perform(sqs_msgs, bodies)
    bodies = Array.wrap bodies
    bodies.each do |body|
      puts body['id']
    end
  end
end

50.times { |i| AddWorker.perform_async foo: SecureRandom.uuid, id: i }
```

```sh
bundle exec shoryuken -v -R -c 1 --delay 5 -q cjlarose_shoryuken_test.fifo
```

In this configuration, Shoryuken will fetch several messages from the queue at a time. The visibility timeout clocks for each message begin to tick. All of the messages fetched are sent to a single worker instance, which can process the messages in order.

I think one difference between the changes proposed in this PR and what I've suggested here is that with the proposed changes, assuming you have a FIFO queue with messages with several distinct `message_group_id`s, it's possible that Shoryuken could process messages received in a single batch, but belonging to distinct `message_group_id`s, concurrently. This could, in theory, reduce message processing latencies.

I think that this is largely already addressed to some extent, though, by SQS's behavior when receiving multiple messages from a FIFO queue. From [the docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html#FIFO-queues-understanding-logic):

> When receiving messages from a FIFO queue with multiple message group IDs, Amazon SQS first attempts to return as many messages with the same message group ID as possible. This allows other consumers to process messages with a different message group ID. 

This behavior means that, as long as Shoryuken has available threads, it can continue to perform batch `ReceiveMessage` calls on the same queue, receive messages with different `message_group_ids` not already being processed, and process those batches concurrently.

I think if users want to minimize latency when using a FIFO queue, their best bet is to use the existing non-batch processing mechanism. If they instead want to reduce API calls at the potential cost of increased latencies, they'd use the existing batch mechanism.

If there's a use case for which the proposed changes are a measurably better solution than anything that Shoryuken already offers, I'd be glad to hear it, but I'm so far I don't know what that use case would be.

PR #632, opened 2020-10-11
Support use custom middleware for InlineExecutor

kamillle:
I'd like to test my custom middleware that is for server with InlineExecutor, but InlineExecutor evaluate default_server_middleware without evaluate custom middleware. It's caused by [InlineExecutor#call](https://github.com/phstc/shoryuken/blob/v5.0.5/lib/shoryuken/worker/inline_executor.rb#L30-L35).
When we use DefaultExecutor, Shoryuken::Worker#perform is called by [Processor#process](https://github.com/phstc/shoryuken/blob/bc7938942270a1171d1597064698d88c3f36a4ec/lib/shoryuken/processor.rb#L21). At that time middleware#call is evaluated before call Shoryuken::Worker#perform. However InlineExecutor#call doesn't evaluate middleware#call before call Shoryuken::Worker#perform.

Why don't we evaluate middleware for InlineExecutor??😄

### DIFF
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -94,7 +94,7 @@ module Shoryuken
     end
 
     def configure_server
-      yield self if server?
+      yield self if server? || inline_executor?
     end
 
     def server_middleware
@@ -159,9 +159,15 @@ module Shoryuken
       @active_job_queue_name_prefixing
     end
 
+    def inline_executor?
+      worker_executor == Worker::InlineExecutor
+    end
+
     private
 
     def default_server_middleware
+      return Middleware::Chain.new if inline_executor?
+
       Middleware::Chain.new do |m|
         m.add Middleware::Server::Timing
         m.add Middleware::Server::ExponentialBackoffRetry
--- a/spec/shoryuken/worker/inline_executor_spec.rb
+++ b/spec/shoryuken/worker/inline_executor_spec.rb
@@ -2,7 +2,17 @@ require 'spec_helper'
 
 RSpec.describe Shoryuken::Worker::InlineExecutor do
   before do
+    # Reset values on memory to set up right condition for InlineExecutor
+    TestWorker.instance_variable_set(:@_server_chain, nil)
+    Shoryuken.shoryuken_options.instance_variable_set(:@_server_chain, nil)
     Shoryuken.worker_executor = described_class
+  end
+
+  after do
+    # Reset values on memory to not affect other examples
+    TestWorker.instance_variable_set(:@_server_chain, nil)
+    Shoryuken.shoryuken_options.instance_variable_set(:@_server_chain, nil)
+    Shoryuken.worker_executor = Shoryuken::Worker::DefaultExecutor
   end
 
   describe '.perform_async' do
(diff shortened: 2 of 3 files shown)


lib/shoryuken/options.rb
```diff
@@ -148,7 +148,7 @@ def on(event, &block)
     end
 
     def server?
-      defined?(Shoryuken::CLI)
+      defined?(Shoryuken::CLI) || inline_executor?
```
kamillle, line 151: Allow yielding with InlineExecutor
https://github.com/phstc/shoryuken/blob/be59f61df31565c63b0840a4655d03ab849fcc81/lib/shoryuken/options.rb#L96-L98

Should I rewrite below??

```rb
def configure_server
  yield self if server? || inline_executor?
end
```

spec/shoryuken/worker/inline_executor_spec.rb
```diff
@@ -46,4 +46,54 @@
       end
     end
   end
+
+  describe 'InlineExecutor with custom middleware' do
```
kamillle, line 50: @phstc 
When I run rspec with only this file(I mean spec/shoryuken/worker/inline_executor_spec.rb), nothing fail. However running rspec with all files, this file examples are failed all. Could u help me with this situation?

kamillle:
@phstc Are you still there?

lib/shoryuken/options.rb
```diff
@@ -159,9 +159,15 @@ def active_job_queue_name_prefixing?
...
     end
 
+    def inline_executor?
+      worker_executor == Worker::InlineExecutor
+    end
+
     private
 
     def default_server_middleware
+      return Middleware::Chain.new if inline_executor?
```
phstc, line 169: I'm wondering if we should create an `inline_middleware` and `configure_inline` instead of patching `server`. 

cc/ @cjlarose

phstc:
Hi @kamillle 

Thanks for your contribution. Can you expand the use case you want a middleware for inline execution? Is it only for "functional" testing your middleware?

lib/shoryuken/options.rb
cjlarose: Yeah checking for `worker_executor == Worker::InlineExecutor` is a definite code smell: it's a violation of the tell-don't-ask principle.

I'd like to know more about the motivation for this change: it's clear that this implementation is trying to make it so that when using the `InlineExecutor`, worker-specific middleware is executed, but we don't execute any of the `default_server_middleware`. I could see that behavior as being unexpected for some users--the least surprising options are either

1) Just execute _all_ server middleware when using the InlineExecutor or
2) Make it clear in the API which middleware is executed when using different executors

I _think_ @phstc is suggesting something like

```ruby
Shoryuken.configure_inline do |config|
  config.inline_middleware do |chain|
    ...
  end
end
```

which mirrors

```ruby
  end
end
```

But I think if the main motivation for changing the default middleware stack is for testing purposes, then this is more simply delegated to the responsibility of the user. Then they can _choose_ whether or not they want to disable the default middleware stack when testing.

```ruby
    ... # configure middleware for default executor
  end

  config.server_middleware.clear if Rails.env.test? # let users decide what do to with default middleware
end
```

cjlarose: The other thing to consider, too, is that as written, this change introduces what I think is a breaking API change since if you were using the `InlineExecutor` before, you might have used it in a way that you didn't expect worker-specific middleware to be executed.

One option would be to introduce a new executor (`InlineExecutorWithMiddleware` or something) so that folks can opt-in to the new behavior without a breaking change.

frsantos:
Sorry to comment on a closed PR. I just noticed that the `InlineExecutor` does not execute any middlewares and found this PR.

Our use case is one in which we send some default data to the workers in a client middleware, and it is recovered and prepared in a server middleware. That default data is taken from thread variables (i.e., some `current_user` and such), so we don't have to pass it around on every worker invocation, and recover it on every worker class.

We use the inline executor sometimes for working on development, or on tests, but they need to work in both environments the same as it works on production.

I think the inline middleware that @phstc suggested, defaulted to empty, would be a great addition for our use case. Would you accept a PR?

PR #635, opened 2020-11-02
Allow message_attributes to be set from ActiveJob clients

timbreitkreutz:
We'd like to be able to access the SQS message attributes from the Shoryuken client code and middleware.  Previously, a memoized copy of the needed attributes were used which interfered with middleware when jobs were nested (i.e. ActiveJob clients calling other Active Jobs).  In addition, we had to work implement workarounds for metrics and tracing attributes at that level.

The suggestion here is to allow these to be passed in as options which then get used as the starting point.  The necessary active job attributes are then overwritten on top of any that are passed in to be sure they function the same as before.

We are open to any suggestions on style, thanks!

Co-authored by @tamouse

### DIFF
--- a/lib/shoryuken/extensions/active_job_adapter.rb
+++ b/lib/shoryuken/extensions/active_job_adapter.rb
@@ -53,30 +53,23 @@ module ActiveJob
       def message(queue, job, options = {})
         body = job.serialize
 
-        msg = {}
+        attributes = options.delete(:message_attributes) || {}
+
+        msg = {
+          message_body: body,
+          message_attributes: attributes.merge(MESSAGE_ATTRIBUTES)
+        }
 
         if queue.fifo?
           # See https://github.com/phstc/shoryuken/issues/457
           msg[:message_deduplication_id] = Digest::SHA256.hexdigest(JSON.dump(body.except('job_id')))
         end
 
-        msg[:message_body] = body
-        msg[:message_attributes] = message_attributes
-
         msg.merge(options)
       end
 
       def register_worker!(job)
         Shoryuken.register_worker(job.queue_name, JobWrapper)
-      end
-
-      def message_attributes
-        @message_attributes ||= {
-          'shoryuken_class' => {
-            string_value: JobWrapper.to_s,
-            data_type: 'String'
-          }
-        }
       end
 
       class JobWrapper #:nodoc:
@@ -88,6 +81,13 @@ module ActiveJob
           Base.execute hash
         end
       end
+
+      MESSAGE_ATTRIBUTES = {
+        'shoryuken_class' => {
+          string_value: JobWrapper.to_s,
+          data_type: 'String'
+        }
+      }.freeze
     end
   end
 end
--- a/.reek.yml
+++ b/.reek.yml
@@ -1,0 +1,5 @@
+---
+detectors:
+
+  UtilityFunction:
+    public_methods_only: true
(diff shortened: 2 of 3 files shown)


timbreitkreutz:
@phstc Here's the nicer version of the PR we deleted a couple weeks ago. Is the utility function in code climate a blocker?  Is there anything else you'd like us to change or adjust on this one?

timbreitkreutz:
Also how do we get Travis running?

timbreitkreutz:
Update: We have been running this version in our main app for a couple weeks now and have no reported issues.

phstc:
Hi @timbreitkreutz 

I can't really review this now. But I added a form for "recruiting" Shoryuken maintainers. Let me know if you are interested https://github.com/phstc/shoryuken/commit/4a03b64a4c48275d83c69b82c60c1379d9a350dc

lib/shoryuken/extensions/active_job_adapter.rb
```diff
@@ -53,32 +53,25 @@ def calculate_delay(timestamp)
       def message(queue, job, options = {})
         body = job.serialize
 
-        msg = {}
+        attributes = options.delete(:message_attributes) || {}
+
+        msg = {
+          message_body: body,
+          message_attributes: attributes.merge(MESSAGE_ATTRIBUTES)
+        }
```
phstc, line 61: @timbreitkreutz Sorry for taking long to get back on this. 

Is it being used when enqueuing a job? `MyJob.perform_later(params)` 

I do miss the ability to send Shoryuken specific params (for example, FIFO params) when using Active Job is this change adding this support?

timbreitkreutz: Sorry was on vacation.  Will try to get an answer soon :)

timbreitkreutz: Sadly I don't think there is a way to inject this via `perform_later`--unless I'm missing something.  This PR was actually aimed more at the middleware layer.

cjlarose: @phstc I was able to test this locally. The proposed changes do not permit specification of `message_attributes` when using `perform_later` (I've documented that as a separate issue in #646).

Instead, the following can be used to enqueue a message with additional `message_attributes`, without overriding the `shoryuken_class` attribute.

```ruby
class AddJob < ActiveJob::Base
  def perform(a, b)
    puts a + b
  end
end

job = AddJob.new 1, 2
AddJob.queue_adapter.enqueue(job, message_attributes: { 'attr' => { string_value: 'myval', data_type: 'String' } })
```

Everything looks good from a testing perspective, so I'll merge it. Although it doesn't solve #646, it doesn't hinder it and might actually be an important prerequisite.

@timbreitkreutz Thanks so much for your patience on this one!

cjlarose: @timbreitkreutz If you could rebase on `master`, I'd appreciate it. Travis should run correctly after pushing up the rebased version. 🤞

timbreitkreutz: Thanks @cjlarose !

timbreitkreutz:
Fork master has been updated.  Merge with squash I believe will result in a single commit to shoryuken repo?

cjlarose:
Thanks! Tests are green. I'll squash merge.

cjlarose:
Whoops. I said I'd squash merge and then immediately forgot. The `master` branch is protected from force-pushes, so I think I'll just leave it. An extra merge commit won't hurt anyone.

PR #644, opened 2021-01-19
Unpause FIFO queues on worker completion

### LINKED ISSUE #643: FIFO Queues with delays
When using FIFO queues & configuring a delay, there will only be 1 message processed per delay period. As an example I have 2 messages in the SQS queue. With a 5 second shoryuken `delay` & the job taking 2 seconds to complete, the process to fetch will wait the 5 seconds to go and fetch the next message from the queue. This proposal would allow for the fetching of the FIFO queue which is paused after a worker has completed the work so the delay would kick in only after fetching with 0 messages being returned.

Configuration:
```
concurrency: 5
delay: 5
queues:
  - ordered.fifo
```

shoryuken.rb
```
  config.on(:dispatch) do
    # called every time it tries to fetch messages from SQS
    puts "#{Time.now}: Fetching Message..."
  end
end
```

Job.rb
```
class Job < ApplicationJob
  self.queue_adapter = :shoryuken
  include Shoryuken::Worker
  queue_as "ordered.fifo"

  def perform(message)
    puts message
    sleep 2
  end
end
```

Logs from running shoryuken with a 5 second delay:

```
2021-01-18T16:39:40Z 6597 TID-owbyxurl8 INFO: Starting
2021-01-18 11:39:40 -0500: Fetching Message...
2021-01-18T16:39:40Z 6597 TID-owbzp8mn0 ActiveJob/Job/development-ordered.fifo/5142b4e9-e146-4e26-b20a-bc44522449d1 INFO: started at 2021-01-18 11:39:40 -0500
2021-01-18 11:39:40 -0500: Fetching Message...
2021-01-18T16:39:42Z 6597 TID-owbzp8mn0 ActiveJob/Job/development-ordered.fifo/5142b4e9-e146-4e26-b20a-bc44522449d1 INFO: completed in: 2127.785 ms
2021-01-18 11:39:45 -0500: Fetching Message...
2021-01-18T16:39:45Z 6597 TID-owbzp8mn0 ActiveJob/Job/development-ordered.fifo/7ec83d58-3eef-485e-b0f7-a349df522117 INFO: started at 2021-01-18 11:39:45 -0500
2021-01-18 11:39:45 -0500: Fetching Message...
2021-01-18T16:39:47Z 6597 TID-owbzp8mn0 ActiveJob/Job/development-ordered.fifo/7ec83d58-3eef-485e-b0f7-a349df522117 INFO: completed in: 2033.8250000000003 ms
```

:pushpin: #644

davidrichey:
* Updated Manager#processor_done to check for FIFO queue & unpause
  * Added specs
* Added unpause method on strategies
  * Added specs around #unpause

closes #643


lib/shoryuken/polling/weighted_round_robin.rb
```diff
@@ -35,17 +35,28 @@ def active_queues
         unparse_queues(@queues)
       end
 
+      def unpause_queue(queue)
+        return if @paused_queues.empty?
+
+        logger.debug "Unapusing #{queue}"
```
cjlarose, line 41:
```suggestion
        logger.debug "Unpausing #{queue}"
```

lib/shoryuken/polling/strict_priority.rb
```diff
@@ -38,6 +38,10 @@ def active_queues
           .reverse
       end
 
+      def unpause_queue(queue)
+        @paused_until[queue] = Time.now
```
cjlarose, line 42: I like the log message in `WeightedRoundRobin#unpause_queue`. It'd be good to have one here, too.

lib/shoryuken/polling/strict_priority.rb
```diff
@@ -38,6 +38,11 @@ def active_queues
           .reverse
       end
 
+      def unpause_queue(queue)
```
cjlarose, line 41: For symmetry with the method `messages_found`, let's consider renaming this method `message_processed`.

The reason is that the interaction between the `Manager` class and the polling strategy is that the manager is just informing the polling strategy of certain events, allowing the polling strategy to respond as it sees fit. In this case, it's not the responsibility of the `Manager` to tell the polling strategy to _pause_ a queue, it's just the responsibility for the `Manager` to let the polling strategy know that it's done processing the message: the polling strategy is welcome to do whatever it wants with that knowledge.

cjlarose: Let's also add `message_processed` to `Shoryuken::Polling::BaseStrategy`. Its implementation can be empty (for backward compatibility--users might have subclassed the base class themselves).

lib/shoryuken/manager.rb
```diff
@@ -57,8 +57,12 @@ def ready
       @max_processors - busy
     end
 
-    def processor_done
+    def processor_done(queue)
       @busy_processors.decrement
+      client_queue = Shoryuken::Client.queues(queue)
+      return unless client_queue.fifo?
+
+      @polling_strategy.unpause_queue(queue)
```
cjlarose, line 65: Since [users can specify their own polling strategy class][0], we want to make it optional to implement the new method. Let's invoke this method only if `@polling_strategy` responds to it.

[0]: https://github.com/cjlarose/shoryuken/blob/1a261d66a678a41266e389642c54bbf07819ee21/lib/shoryuken/options.rb#L65-L66

cjlarose (changes requested):
Overall approach seems solid! I took another pass after reviewing some of the surrounding context.

cjlarose:
Thanks for being so responsive to feedback! I'll do some testing on some real queues, but overall, this is looking great!

cjlarose:
I tested this locally on some FIFO queues. On `master`, I observed that if `concurrency` is more than `1` and you have a `delay` configured, FIFO message throughput is very poor because the "fetcher" pauses every time it gets an empty receive.

On this branch, I observed that the fetcher resumed whenever a job is finished, improving throughput, especially when the number of distinct `message_group_id`s is small.

PR #654, opened 2021-02-10
Fix keyword argument error

daikimiura:
This PR fixes keyword arguments error in Active Job.
In `shoryuken v5.1.0` with `ruby 3.0.0p0`, `perform` method with keyword arguments causes `ArgumentError` as below.
```
class KwargsJob < ActiveJob::Base
  def perform(argument:)
    puts argument
  end
end

KwargsJob.perform_now(argument: 2) # => ArgumentError: wrong number of arguments (given 1, expected 0)
```
I suggest the same approach as Rails. ( https://github.com/rails/rails/pull/38267 )

@phstc @cjlarose

### DIFF
--- a/lib/shoryuken/extensions/active_job_extensions.rb
+++ b/lib/shoryuken/extensions/active_job_extensions.rb
@@ -19,6 +19,7 @@ module Shoryuken
         super(*arguments)
         self.sqs_send_message_parameters = {}
       end
+      ruby2_keywords(:initialize) if respond_to?(:ruby2_keywords, true)
 
       def enqueue(options = {})
         sqs_options = options.extract! :message_attributes,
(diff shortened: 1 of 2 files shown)


cjlarose:
This looks like a regression in shoryuken 5.1.0. The fix looks correct since we want source compatibility with all supported Ruby versions. I'll do some testing locally later today before I merge this and push a bugfix release.

Thanks for your contribution @daikimiura!

cjlarose:
I was able to add a test that fails under Ruby 3.0 without the proposed fix, and passes with the fix in place. I'll merge and create a bugfix release.

cjlarose:
Shoryuken 5.1.1 is released with this fix in place. Thanks again, @daikimiura !

PR #664, opened 2021-04-02
Adds a check to automatically adjust for batch size limit enforced by aws

correnarsprowls:
This addresses https://github.com/ruby-shoryuken/shoryuken/issues/663

The change to `batch_send` will ensure that a batch is never larger than the 256kb limit set by SQS. If the batch size is too large then it will recurse until the batch size is small enough.  If a single message is larger than the batch size limit then it will display the original error message but the rest of the batch will succeed.

These changes were tested using LocalStack.

### DIFF
--- a/bin/cli/sqs.rb
+++ b/bin/cli/sqs.rb
@@ -4,6 +4,9 @@ require 'date'
 module Shoryuken
   module CLI
     class SQS < Base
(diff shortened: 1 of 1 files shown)


cjlarose:
This looks great! I'll do some testing locally, but the code looks pretty good.

cjlarose:
I verified locally that this will correctly subdivide message batches that exceed the SQS batch message payload limit.

cjlarose:
Thanks for your contribution, @correnarsprowls!

PR #671, opened 2021-07-09
Update `dispatch` hook to have utilization metrics

rbroemeling:
I'd like to export metrics from Shoryuken so that we have metrics available for the number of worker threads busy at any given time. The logical thing to do is to put the metrics in the `dispatch` lifecycle hook, but right now they don't have that information available.

I'm suggesting that we add that information to the arguments that are passed to the hook, so that it is available within the hook.

### DIFF
--- a/lib/shoryuken/manager.rb
+++ b/lib/shoryuken/manager.rb
@@ -38,7 +38,7 @@ module Shoryuken
         return sleep(MIN_DISPATCH_INTERVAL)
       end
 
-      fire_event(:dispatch, false, queue_name: queue.name)
+      fire_event(:dispatch, false, busy: busy, queue_name: queue.name, ready: ready)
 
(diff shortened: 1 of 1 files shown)


cjlarose:
I'm hesitant to add `busy` and `ready` to the `dispatch` event because it's not really clear how to make this work correctly in configurations that use multiple [processing groups](https://github.com/ruby-shoryuken/shoryuken/wiki/Processing-Groups) (and therefore, have many instances of `Shoryuken::Manager`, which each have different values for `busy` and `ready`). Additionally, `dispatch` is fired off before `busy` is actually decremented, and that's a little unexpected.

I do, however, recognize the value in being able to extract the current state of the Shoryuken process as far as understanding its utilization. I've documented that as a separate issue in #672

PR #681, opened 2021-10-15
Use Forwardable within Message to avoid method boilerplate

GabrielMalakias:
First of all, thank you for maintaining this gem! I really appreciate your efforts. 

While diving into the code I just noticed that a bunch of methods were defined just as a proxy to the data structure. 

As I have seen here: https://github.com/ruby-shoryuken/shoryuken/blob/master/lib/shoryuken.rb#L48 I decided to use the SingleForwardable to delegate to the structure directly.

It's a small thing for now but might the beginning of some contributions

### DIFF
--- a/lib/shoryuken/message.rb
+++ b/lib/shoryuken/message.rb
@@ -1,6 +1,6 @@
 module Shoryuken
   class Message
-    extend SingleForwardable
+    extend Forwardable
 
     def_delegators(:data,
                    :message_id,


cjlarose:
Thanks for your contribution! Tests are failing. I think you might want to look into `Forwardable` instead of `SingleForwardable`.

https://ruby-doc.org/stdlib-2.6.2/libdoc/forwardable/rdoc/Forwardable.html

lib/shoryuken/message.rb
```diff
@@ -1,5 +1,16 @@
...
+
+    def_delegators(:data,
+                   :message_id,
+                   :receipt_handle,
+                   :md5_of_body,
+                   :body,
+                   :attributes,
+                   :md5_of_message_attributes,
+                   :message_attributes)
+
```
januszm, line 13: A minor thing but skipping the parentheses in this case would match the macro-style methods better and also sorting the lines ascending would make it easier to read, e.g.:

```ruby
def_delegators :attributes,
               :body,
               :data,
               :md5_of_body,
               :md5_of_message_attributes,
               :message_attributes,
               :message_id,
               :receipt_handle
```

Issue #670, opened 2021-07-08
Shoryuken worker claims queue does not exist


coding-bunny:
Hello,

I hope I can explain this properly, because I'm at a loss as to why the worker in our K8s pod keeps reporting that the queue does not exist. The credentials are set up correctly, as we use the IAM Profile approach in our containers to assign the required role to a pod for AWS access.

This is what's inside the `config/shoryuken.yml`:

```yaml
---
concurrency: 2
pool: 1
delay: 1
queues:
  - <%= ::ENV['SQS_QUEUE_NAME'] %>
```

We've created a Worker inside our `lib/` folder of the Rails application, but we're not using ActiveJob.
Rails has been given an initializer to load the Worker after Rails has been loaded, which should be the case if I understand the documentation correctly, given that it works locally and no errors are thrown:

```
bundle exec shoryuken -R -C config/shoryuken.yml

bundler: failed to load command: shoryuken (/app/vendor/bundle/ruby/2.7.0/bin/shoryuken)
ArgumentError: The specified queue(s) deployment-history-queue do not exist.
Try 'shoryuken sqs create QUEUE-NAME' for creating a queue with default settings
 app/vendor/bundle/ruby/2.7.0/gems/shoryuken-5.2.2/lib/shoryuken/environment_loader.rb:162:in `validate_queues'
/app/vendor/bundle/ruby/2.7.0/gems/shoryuken-5.2.2/lib/shoryuken/environment_loader.rb:30:in `load'
/app/vendor/bundle/ruby/2.7.0/gems/shoryuken-5.2.2/lib/shoryuken/runner.rb:39:in `run'
/app/vendor/bundle/ruby/2.7.0/gems/shoryuken-5.2.2/bin/shoryuken:42:in `start'
/app/vendor/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:27:in `run'
/app/vendor/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor/invocation.rb:127:in `invoke_command'
/app/vendor/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor.rb:392:in `dispatch'
/app/vendor/bundle/ruby/2.7.0/gems/thor-1.1.0/lib/thor/base.rb:485:in `start'
/app/vendor/bundle/ruby/2.7.0/gems/shoryuken-5.2.2/bin/shoryuken:53:in `<top (required)>'
[...]

Issue #511, opened 2018-07-18
Rails is loaded after config file is parsed


adaminglehart:
Up until recently I was using an older version of Shoryuken (2.0.11). I'm currently trying to upgrade to the latest version (3.2.3), but I've found that Rails is now being loaded **after** setting up options from a config file (when using the `--rails` and `--config=<config_file_path>` CLI options). This was not the case before, so you were able to reference things like environment variables if using a library like dotenv or Rails config from application.yml, e.g.
```
queues:
  - <%= CONFIG[:sqs_queue] %>
  - my-<%= Rails.env %>-queue
```
Looks like https://github.com/phstc/shoryuken/pull/261 is the PR where this changed. Based on https://github.com/phstc/shoryuken/pull/197 it seems like, at least at some point, Rails was intended to be loaded first, perhaps this is no longer the case? If not, any recommendations for how to accomplish this a different way? Thanks!

phstc:
Hi @adaminglehart 

There are some problems with loading Rails first, if we do it first, there will be some issues in Shoryuken, like if we fix one, we break the other.

Some alternatives:

- a) Replace `Rails.env` with `ENV['RAILS_ENV']` in case you are setting `RAILS_ENV`
- b) Create two configs for Shoryuken, shoryuken.development.yml and shoryuken.yml

WDYT?

adaminglehart:
@phstc Thanks for the prompt response! 

I see, that's too bad. Using `ENV['RAILS_ENV']` will work for that particular case, but for any other environment-specific configuration I'll have to go with using different Shoryuken configs per-environment. I was hoping to avoid this because I'd prefer to store things like environment-specific queues in a centralized place like application.yml, since that info also needs to be accessed in the worker class and potentially other places in the codebase, but I'll figure out a way to make it work since it sounds like that won't be an option at the moment. 

This is a total shot in the dark, but an interesting option would be if the name of the worker's queue could somehow be specified separately from the actual SQS endpoint. So, if your shoryuken.yml was just
```
queues:
  - high-priority
```
and then the worker class had `shoryuken_options queue: 'high-priority', sqs_url: CONFIG[:high_priority_queue_url]`, or if in your shoryuken initializer you could specify mappings from queue names to SQS endpoints. This would be similar to how Sidekiq works, where the name of your queues is separate from which Redis endpoint you're connecting to. I'm not familiar enough with Shoryuken's codebase to know if this is even remotely an option (I imagine it's not) but just an idea

[...]

PR #686, opened 2021-11-02
Initialize Rails before parsing Config File

csebryam:
### Description

When you specify to load rails in the `shoryuken.yml` file ( i.e `rails: true`), `Shoryuken::EnvironmentLoader#setup_options` will read the `config_file` before rails has had a chance to be initialized. Any declared rails environment variables will not be present at this time. 

In the example below: 
```ruby
# config/shoryuken.yml

------------
delay: 25
queues:
  - <%= ENV['AWS_SQS_QUEUE'] %>
```
`AWS_SQS_QUEUE` will not be present in the rails environment variables as it is currently. As a result a log output will show this message: `WARN: No queues supplied`

### Fixes:
Initialize rails as the first step in the `#setup_options` 

### Description

Currently if you set the `:rails` flag in the CLI with the `-R` option, the `Shoryuken::Runner` will override options in the config files with the CLI options
```ruby
# lib/shoryuken/runner.rb

loader = EnvironmentLoader.setup_options(options)

# When cli args exist, override options in config file
Shoryuken.options.merge!(options)

loader.load
```

When `loader.load` gets called, we run into the same issue as Part 1. The `config_file` has already been read-in and any rails environment variables were not present at that time. Again, this is because loading rails happens too late in the process, in the `#load` method.

### Fixes
`Shoryuken::EnvironmentLoader` will initialize rails looking for flags from either `Shoryuken.options` or CLI `options` hash. See `load_rails?` method.

## Other

Fixes issues similar to these #670
> Because even though I use the dotenv gem, the environment variables are not loaded properly by Shoryuken.
It does work when they exist on the OS level

Similar Issues #511

### DIFF
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -80,7 +80,7 @@ module Shoryuken
     end
 
     def load_rails?
-      options[:rails] || Shoryuken.options[:rails]
+      options[:rails]
     end
 
     def prefix_active_job_queue_name(queue_name, weight)
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -18,12 +18,12 @@ module Shoryuken
     end
 
     def setup_options
+      load_rails if Shoryuken.options[:rails]
       initialize_options
       initialize_logger
     end
 
     def load
-      load_rails if Shoryuken.options[:rails]
       prefix_active_job_queue_names
       parse_queues
       require_workers
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -18,7 +18,7 @@ module Shoryuken
     end
 
     def setup_options
-      load_rails if Shoryuken.options[:rails]
+      load_rails if options[:rails] || Shoryuken.options[:rails]
       initialize_options
       initialize_logger
     end
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -18,7 +18,7 @@ module Shoryuken
     end
 
     def setup_options
-      load_rails if options[:rails] || Shoryuken.options[:rails]
+      initialize_rails if load_rails?
       initialize_options
       initialize_logger
     end
@@ -55,7 +55,7 @@ module Shoryuken
       Shoryuken.logger.level = Logger::DEBUG if Shoryuken.options[:verbose]
     end
 
-    def load_rails
+    def initialize_rails
       # Adapted from: https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb
 
       require 'rails'
@@ -77,6 +77,10 @@ module Shoryuken
         end
         require File.expand_path('config/environment.rb')
       end
+    end
+
+    def load_rails?
+      options[:rails] || Shoryuken.options[:rails]
     end
 
     def prefix_active_job_queue_name(queue_name, weight)
--- a/lib/shoryuken/runner.rb
+++ b/lib/shoryuken/runner.rb
@@ -30,9 +30,6 @@ module Shoryuken
 
       loader = EnvironmentLoader.setup_options(options)
 
-      # When cli args exist, override options in config file
-      Shoryuken.options.merge!(options)
-
       daemonize(Shoryuken.options)
       write_pid(Shoryuken.options)
 
--- a/spec/shoryuken/environment_loader_spec.rb
+++ b/spec/shoryuken/environment_loader_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Shoryuken::EnvironmentLoader do
 
   describe '#parse_queues loads default queues' do
     before do
-      allow(subject).to receive(:load_rails)
+      allow(subject).to receive(:initialize_rails)
       allow(subject).to receive(:prefix_active_job_queue_names)
       allow(subject).to receive(:require_workers)
       allow(subject).to receive(:validate_queues)
@@ -24,7 +24,7 @@ RSpec.describe Shoryuken::EnvironmentLoader do
 
   describe '#parse_queues includes delay per groups' do
     before do
-      allow(subject).to receive(:load_rails)
+      allow(subject).to receive(:initialize_rails)
       allow(subject).to receive(:prefix_active_job_queue_names)
       allow(subject).to receive(:require_workers)
       allow(subject).to receive(:validate_queues)
@@ -47,7 +47,7 @@ RSpec.describe Shoryuken::EnvironmentLoader do
 
   describe '#prefix_active_job_queue_names' do
     before do
-      allow(subject).to receive(:load_rails)
+      allow(subject).to receive(:initialize_rails)
       allow(subject).to receive(:require_workers)
       allow(subject).to receive(:validate_queues)
       allow(subject).to receive(:validate_workers)


cjlarose:
Hey @csebryam, thanks for opening this PR and sorry I've been slow to give feedback.

In general, I'm open to this kind of change, but I'm hesitant because I think it might be backwards-incompatible. In some of the applications I maintain, we reference, for example, the Shoryuken logger instance in a Rails initializer like so

```ruby
```

I think this change might actually not break my applications in this case, but it's difficult to say that this change won't break _anyone_'s current configuration.

What would help me as a maintainer would be to know a little bit more about your use case. What are you trying to accomplish with this change? Do you expect to have to export environment variables, for example, in a Rails initializer, in order to read them from `shoryuken.yml`?

If so, one workaround is to use Shoryuken's configuration API instead of the YAML file. An example is in https://github.com/ruby-shoryuken/shoryuken/wiki/Rails-Integration-Active-Job

```ruby
# config/initializers/shoryuken.rb

  # Replace Rails logger so messages are logged wherever Shoryuken is logging
  # Note: this entire block is only run by the processor, so we don't overwrite
  #       the logger when the app is running as usual.

  Rails.logger.level = Rails.application.config.log_level

  # config.server_middleware do |chain|
  #  chain.add Shoryuken::MyMiddleware
  # end

  # For dynamically adding queues prefixed by Rails.env 
  # Shoryuken.add_group('default', 25)
  # %w(queue1 queue2).each do |name|
  #   Shoryuken.add_queue("#{Rails.env}_#{name}", 1, 'default')
  # end
end
```

csebryam:
@cjlarose Thanks for your response and for being open to adding these changes. Also, thanks for sharing your `shoryuken.rb` initializer file as it was helpful in understanding how you're initializing the Rails logger. 

To avoid any logger issues, we could move the `initialize_logger` as the first setup step. 
```ruby
    def setup_options
      initialize_logger
      initialize_rails if load_rails?
      initialize_options
    end
```

###  Use case:
1. To help new Shoryuken gem users avoid Rails specific environment loading issues. I've seen many _Issues_ and _StackOverflow_ questions regarding this (some of which I looked at to get additional help). It would definitely streamline the integration.
2. Without this change, it makes it difficult to use Dotenv or Figaro to setup environment variables as stated in the [Wiki](https://github.com/ruby-shoryuken/shoryuken/wiki/Complete-Shoryuken-Setup-for-Rails-5-with-ActiveJob#configure-shoryukenyml
): 
> You can use dotenv to setup your environment variables.
3. This is basic functionality that should just work without having to find workarounds.

In the `EnvironmentLoader` file, there is an inline comment that states it was [adapted from Sidekiq cli ](https://github.com/ruby-shoryuken/shoryuken/blob/master/lib/shoryuken/environment_loader.rb#L59). Just going off of that, Sidekiq loads the environment as the first step in the `setup_options` method, and then loads the `config_file` as seen here: https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/cli.rb#L225-L241. We would essentially be doing the same thing in this PR. 

Hope to hear your thoughts.

cjlarose:
Ok this definitely feels like the right change to make. I don't think it's necessary to move the `initialize_logger` step up. We can start with `initialize_rails`

I just want to clarify what the expected behavior is in a bunch of cases so that we can clearly communicate this change in the changelog.

If a user, for example, was using `Shoryuken.configure_server` to, for example, set `delay` and `timeout` values _and_ has a `config/shoryuken.yml` file _and_ set some options on the CLI, what order is everything interpreted? Does that order change with this PR?

csebryam:
@cjlarose Sorry for my delay in responding, this took some digging.

Bumping to a major version makes sense to me that way it doesn't break anyones expected behavior.

> what order is everything interpreted? Does that order change with this PR?

After some investigation, this PR does change the order things get interpreted.

options found in `Shoryuken.configure_server` will take precedence because of this code here:
```ruby
    def add_group(group, concurrency = nil, delay: nil)
      concurrency ||= options[:concurrency]
      delay ||= options[:delay]
      ...
    end
```

The order things get loaded in this PR:
1. Because rails is initialized as the first step, `Shoryuken.configure_server` options will be loaded first.
    - It's the first time the method `add_group` gets called, setting any default options mentioned above as the overall defaults.

2. Other options get merged in when `Shoryuken::Runner` runs:
```ruby
loader = EnvironmentLoader.setup_options(options)
loader.load
```

Again, keeping any defaults configured in the initializer.

### Previous Version Below

**Whats more, loading rails before or after doesn't change much in terms of how defaults get interpreted.**

```ruby
    def load
      load_rails
      prefix_active_job_queue_names
      parse_queues
      ...
    end
```

This is because in both cases rails gets loaded/initialized before `parse_queues`. This results in loading options from the rails initialized file options found in `Shoryuken.configure_server` first. 

In your example above:
```ruby
  Shoryuken.add_group('default', 25)
    %w(queue1 queue2).each do |name|
      Shoryuken.add_queue("#{Rails.env}_#{name}", 1, 'default')
  end
end
```

This code is essentially what `parse_queues` runs (i.e `Shoryuken.add_group`) and it happens first in both cases, keeping the code interpretation order basically the same.

csebryam:
@cjlarose Not sure if I was able to answer your questions fully but I'm happy to look into anything else. Do you think this PR is ready to merge?

cjlarose:
Hey @csebryam! I appreciate you figuring out the details on how everything gets initialized. Sorry I haven't been responsive--I took some vacation time. This is still on my radar, though. I'll take a deeper look at it and get you some feedback!

lib/shoryuken/environment_loader.rb
```diff
@@ -18,12 +18,12 @@ def initialize(options)
     end
 
     def setup_options
+      initialize_rails if load_rails?
       initialize_options
```
cjlarose, line 22: Since `EnvironmentLoader#intialize_options` will read in the config file first, then the CLI args, it seems like we can safely remove the call to merge in the CLI args from  `Shoryuken::Runner#run`. What do you think?

https://github.com/ruby-shoryuken/shoryuken/blob/b0dc0c505885be5fcd8c1e31a6a544b0ebc9800d/lib/shoryuken/runner.rb#L33-L34

lib/shoryuken/environment_loader.rb
```diff
@@ -79,6 +79,10 @@ def load_rails
       end
     end
 
+    def load_rails?
+      options[:rails] || Shoryuken.options[:rails]
```
cjlarose, line 83: Is it possible to rely solely on the CLI arg `options[:rails]` here? It's not clear to me in what situation `options[:rails]` would be falsey here, but `Shoryuken.options[:rails]` would be truthy here, especially since by the time this method is invoked, `initialize_options` has not yet run.

cjlarose (changes requested):
Hey @csebryam ! Sorry it took so long to get you some feedback. I was able to take a deeper look and left some comments, but this change looks overall pretty great! Thanks again for taking this on.

lib/shoryuken/environment_loader.rb
csebryam: I would be ok with just having `options[:rails]`, that is how I'm using it.  Though, I wanted to keep the existing functionality in case someone was enabling it through the `config/shoryuken.yml` file.

cjlarose: Correct me if I'm wrong, but `Shoryuken.options[:rails]` won't be populated by the value in `config/shoryuken.yml` until `initialize_options` runs. So if the intent is to preserve functionality of setting `rails: true` in the config file, I don't think this code accomplishes that.

Either way, I don't think it's necessary to preserve the behavior of allowing `rails: true` to be set in the config file because it's a bit circular: if we have to read the config file to know whether or not to load rails, but we also don't want to read the config file until _after_ initializing Rails, that seems messy.

csebryam: That's right, by the time `initialize_options` runs, it has already passed the option to `initialize_rails`.

> if we have to read the config file to know whether or not to load rails, but we also don't want to read the config file until after initializing Rails, that seems messy.

That is a very good point!

Moving forward with this change to only allowing CLI arg to load rails.

lib/shoryuken/environment_loader.rb
csebryam: That makes sense. Also, removing this line will clear things up as to when/how the options get loaded and avoids further confusion.

cjlarose:
Changes look perfect. I'll have to take another pass later and do some testing on some of my own applications, but I think this is just about good to go.

cjlarose:
Tested on my own applications. Looks great! Thanks for much for this!

gondalez:
I just wanted to say a belated thank you for this change, it helped to dry up our app's config nicely 🙏

PR #687, opened 2021-11-03
Preserve queue weights when unpausing

cjlarose:
#644 introduced a change that unpauses FIFO queues when a message is processed from that queue. This prevents unnecessary delays, especially when using FIFO queues that contain messages with just a single unique `message_group_id`.

One side-effect of its implementation, however, reset queue weights in the `WeightedRoundRobin` polling strategy. In that class, `@queues` is an array of queue names, but some of the queue names are duplicated intentionally in order to represent weights. The change in #644 called `uniq!` on `@queues`, effectively resetting queue weights back down to 1 for all active (unpaused) queues. Since queue weights grow and shrink dynamically over time to adjust to the number of messages fetched from those queues, this is undesirable.

cc @davidrichey

### DIFF
--- a/lib/shoryuken/polling/weighted_round_robin.rb
+++ b/lib/shoryuken/polling/weighted_round_robin.rb
@@ -36,12 +36,10 @@ module Shoryuken
       end
 
       def message_processed(queue)
-        return if @paused_queues.empty?
+        paused_queue = @paused_queues.find { |_time, name| name == queue }
+        return unless paused_queue
 
-        logger.debug "Unpausing #{queue}"
-        @paused_queues.reject! { |_time, name| name == queue }
-        @queues << queue
-        @queues.uniq!
+        paused_queue[0] = Time.at 0
       end
 
       private
--- a/spec/shoryuken/polling/weighted_round_robin_spec.rb
+++ b/spec/shoryuken/polling/weighted_round_robin_spec.rb
@@ -106,12 +106,30 @@ RSpec.describe Shoryuken::Polling::WeightedRoundRobin do
   end
 
   describe '#message_processed' do
-    it 'removes paused queue, adds to active queues' do
-      strategy = Shoryuken::Polling::WeightedRoundRobin.new([queue1, queue2])
-      strategy.send(:pause, queue1)
-      expect(strategy.active_queues).to eq([[queue2, 1]])
-      strategy.message_processed(queue1)
-      expect(strategy.active_queues).to eq([[queue2, 1], [queue1, 1]])
+    it 'removes delay from paused queue' do
+      queues << queue1
+      queues << queue2
+      subject.send(:pause, queue1)
+
+      expect(subject.active_queues).to eq([[queue2, 1]])
+      subject.message_processed(queue1)
+      expect(subject.active_queues).to eq([[queue2, 1]])
+      subject.send(:unpause_queues)
+      expect(subject.active_queues).to eq([[queue2, 1], [queue1, 1]])
+    end
+
+    it 'preserves weight of queues when unpausing' do
+      queues << queue1
+      queues << queue1
+      queues << queue2
+      subject.messages_found(queue1, 1)
+      subject.send(:pause, queue2)
+
+      expect(subject.active_queues).to eq([[queue1, 2]])
+      subject.message_processed(queue2)
+      expect(subject.active_queues).to eq([[queue1, 2]])
+      subject.send(:unpause_queues)
+      expect(subject.active_queues).to eq([[queue1, 2], [queue2, 1]])
     end
   end
 end
--- a/spec/shoryuken/polling/weighted_round_robin_spec.rb
+++ b/spec/shoryuken/polling/weighted_round_robin_spec.rb
@@ -109,26 +109,33 @@ RSpec.describe Shoryuken::Polling::WeightedRoundRobin do
     it 'removes delay from paused queue' do
       queues << queue1
       queues << queue2
-      subject.send(:pause, queue1)
+
+      expect(subject.next_queue).to eq(queue1)
+      subject.messages_found(queue1, 0) # pauses queue1
 
       expect(subject.active_queues).to eq([[queue2, 1]])
-      subject.message_processed(queue1)
-      expect(subject.active_queues).to eq([[queue2, 1]])
-      subject.send(:unpause_queues)
-      expect(subject.active_queues).to eq([[queue2, 1], [queue1, 1]])
+
+      subject.message_processed(queue1) # marks queue1 to be unpaused
+
+      expect(subject.next_queue).to eq(queue2) # implicitly unpauses queue1
+      expect(subject.active_queues).to eq([[queue1, 1], [queue2, 1]])
     end
 
     it 'preserves weight of queues when unpausing' do
       queues << queue1
       queues << queue1
       queues << queue2
+
+      expect(subject.next_queue).to eq(queue1)
       subject.messages_found(queue1, 1)
-      subject.send(:pause, queue2)
+
+      expect(subject.next_queue).to eq(queue2)
+      subject.messages_found(queue2, 0) # pauses queue2
 
       expect(subject.active_queues).to eq([[queue1, 2]])
-      subject.message_processed(queue2)
-      expect(subject.active_queues).to eq([[queue1, 2]])
-      subject.send(:unpause_queues)
+      subject.message_processed(queue2) # marks queue2 to be unpaused
+
+      expect(subject.next_queue).to eq(queue1) # implicitly unpauses queue2
       expect(subject.active_queues).to eq([[queue1, 2], [queue2, 1]])
     end
   end
--- a/lib/shoryuken/polling/strict_priority.rb
+++ b/lib/shoryuken/polling/strict_priority.rb
@@ -39,8 +39,10 @@ module Shoryuken
       end
 
       def message_processed(queue)
-        logger.debug "Unpausing #{queue}"
-        @paused_until[queue] = Time.at 0
+        if queue_paused?(queue)
+          logger.debug "Unpausing #{queue}"
+          @paused_until[queue] = Time.at 0
+        end
       end
 
       private
--- a/lib/shoryuken/polling/strict_priority.rb
+++ b/lib/shoryuken/polling/strict_priority.rb
@@ -40,7 +40,7 @@ module Shoryuken
 
       def message_processed(queue)
         logger.debug "Unpausing #{queue}"
-        @paused_until[queue] = Time.now
+        @paused_until[queue] = Time.at 0
       end
 
       private


lib/shoryuken/polling/weighted_round_robin.rb
```diff
@@ -36,12 +36,10 @@ def active_queues
       end
 
       def message_processed(queue)
-        return if @paused_queues.empty?
+        paused_queue = @paused_queues.find { |_time, name| name == queue }
```
davidrichey, line 39: @cjlarose Is there a case where multiple of the same queues will be paused? If so, would the desired behavior be to unpause all queues matching the name?

The current implementation removes all paused queues by the same name (`@paused_queues.reject! { |_time, name| name == queue }`)

lib/shoryuken/polling/weighted_round_robin.rb
```diff
@@ -36,12 +36,10 @@ def active_queues
...
+        return unless paused_queue
 
-        logger.debug "Unpausing #{queue}"
-        @paused_queues.reject! { |_time, name| name == queue }
-        @queues << queue
```
davidrichey, line 43: When a queue gets [paused](https://github.com/ruby-shoryuken/shoryuken/pull/687/files#diff-4ecb88f1d35234e6a67d51dce18553f9915646a21cd18f43ecdaa01e4102948bR48) the queue is remove from `@queues`. This seems that it would break the activation of paused queues on `#message_processed`

spec/shoryuken/polling/weighted_round_robin_spec.rb
```diff
@@ -106,12 +106,30 @@
...
-      strategy.message_processed(queue1)
-      expect(strategy.active_queues).to eq([[queue2, 1], [queue1, 1]])
+    it 'removes delay from paused queue' do
+      queues << queue1
+      queues << queue2
+      subject.send(:pause, queue1)
+
+      expect(subject.active_queues).to eq([[queue2, 1]])
+      subject.message_processed(queue1)
+      expect(subject.active_queues).to eq([[queue2, 1]])
```
davidrichey, line 116: I believe to keep the intent on #644 the test should stay the same here:

```
expect(strategy.active_queues).to eq([[queue2, 1], [queue1, 1]])
```

spec/shoryuken/polling/weighted_round_robin_spec.rb
```diff
@@ -106,12 +106,30 @@
...
+    it 'preserves weight of queues when unpausing' do
+      queues << queue1
+      queues << queue1
+      queues << queue2
+      subject.messages_found(queue1, 1)
+      subject.send(:pause, queue2)
+
+      expect(subject.active_queues).to eq([[queue1, 2]])
+      subject.message_processed(queue2)
+      expect(subject.active_queues).to eq([[queue1, 2]])
```
davidrichey, line 130: Same concept as the comment above:
```
expect(subject.active_queues).to eq([[queue1, 2], [queue2, 1]])
```

davidrichey (commented):
@cjlarose thanks for information. I agree that the changes made in #644 would break the weighting of the polling strategy when a queue is unpaused. 

I've added some comments as I have a couple concerns around the unpausing of a queue in the `message_processed` method. It has been some time since I look at this code so please let me know if I missed something.

lib/shoryuken/polling/weighted_round_robin.rb
cjlarose: This is a good catch! I'll look more into this and see what's going on.

spec/shoryuken/polling/weighted_round_robin_spec.rb
cjlarose: So the behavior of `message_processed` for `WeightedRoundRobin` changed slightly that it doesn't actually unpause the queue immediately, it now just changes the "paused until" timestamp such that when the next queue is requested, `unpause_queues` will actually do the unpausing. I think the intend of #644 is still preserved, it's just that the internal data structures are modified at slightly different times.

Let me know if I'm misunderstanding it.

I think the new proposed change to `WeightedRoundRobin#message_processed` simplifies things a little since now only `unpause_queues` is responsible for manipulating the right data structures for unpausing.

lib/shoryuken/polling/weighted_round_robin.rb
cjlarose: It's not clear from this diff alone, but I think `@queues` is now correctly updated in `unpause_queues`

spec/shoryuken/polling/weighted_round_robin_spec.rb
davidrichey: Got it. I think this is a good implementation. It does simplify the responsibility for unpausing queues 💯

davidrichey:
@cjlarose I was able to take a spin with these updates locally making sure the changes unpause the queue as expected when a new message is on the queue at the time the message processes. Once [this comment](https://github.com/ruby-shoryuken/shoryuken/pull/687#discussion_r743074208) is confirmed/resolved I believe this will be good to go.

lib/shoryuken/polling/weighted_round_robin.rb
cjlarose: > Is there a case where multiple of the same queues will be paused?

Yes, this can happen for any queue whose maximum weight is more than 1. 

> If so, would the desired behavior be to unpause all queues matching the name?

This is probably best illustrated by an example:

Let's start off by assuming the `WeightedRoundRobin` polling strategy has got up to a weight of 3 for `queue1` (and that the maximum weight for `queue1` is 3).

```ruby
queues = %w(queue1 queue1 queue1)
paused_queues = []
```

For context, each iteration of the `Manager`'s `dispatch` loop executes `#next_queue` on the `polling_strategy`, then `#messages_found`. Any time a job completes, `#message_processed` is invoked. (There's no mechanism AFAIK to prevent `#message_processed` from being invoked concurrently with `#next_queue` or `#messages_found`, but let's put aside any weird data races for now to keep things simple.)

Assume we get an "empty receive" from SQS for `queue1`. In `messages_found`, we end up putting a single instance of the queue into `paused_queues`. Assuming our start time is `t0` and `delay` is 5 seconds,

```ruby
queues = %w(queue1 queue1)
paused_queues = [[t0 + 5, 'queue1']]
```

Assuming the next few `#next_queue` operations complete before `delay`, all instances of the queue end up paused. For simplicity, we'll assume the "paused until" times are one second apart each:

```ruby
queues = []
paused_queues = [
  [t0 + 5, 'queue1'],
  [t0 + 6, 'queue1'],
  [t0 + 7, 'queue1'],
]
```

When `#next_queue` is called after `t0 + 5` has passed, then exactly one instance of the queue will be unpaused by `#unpause_queues`.

```ruby
queues = %w(queue1)
paused_queues = [
  [t0 + 6, 'queue1'],
  [t0 + 7, 'queue1'],
]
```

If there are still no messages in the queue, `#messages_found` will return the queue back to `paused_queues`.

If, however, there are messages in the queue this time, `#messages_found` will actually increase the weight of the queue in `queues`.

```ruby
queues = %w(queue1 queue1)
paused_queues = [
  [t0 + 6, 'queue1'],
  [t0 + 7, 'queue1'],
]
```

Assuming the job takes a while to run, and no new messages appear on the queue, the queues will all become paused again eventually:

```ruby
queues = []
paused_queues = [
  [t0 + 8, 'queue1'],
  [t0 + 9, 'queue1'],
  [t0 + 10, 'queue1'],
  [t0 + 11, 'queue1'],
]
```

Then, assuming our job completes, `#message_processed` will be called. We have two options: we can either modify the "paused until" timestamp of just one occurrence of the queue in `paused_queues`, or we can update all of them.

Suppose we modify the "paused until" timestamp for just one instance of the queue in `paused_queues` (as we do in 02b340acaf88aabbcde4719d68d8616b60797634), specifically the first that we find:

```ruby
queues = []
paused_queues = [
  [Time.at 0, 'queue1'],
  [t0 + 9, 'queue1'],
  [t0 + 10, 'queue1'],
  [t0 + 11, 'queue1'],
]
```

On the next call to `#next_queue`, that first instance will be unpaused

```ruby
queues = %w(queue1)
paused_queues = [
  [t0 + 9, 'queue1'],
  [t0 + 10, 'queue1'],
  [t0 + 11, 'queue1'],
]
```

The weight of the queue will ramp up on each iteration of the dispatch loop as long as there are messages in the queue. (Once a single instance of the queue is in `queues`, it's possible to increase the weight in this way, due the behavior of `#messages_found`).

If instead we modify the "paused until" timestamp for all instances of the queue in `paused_queues` (maybe like e140f17d5a6567ca7e3ecbbc7c57cab5813183e5):

```ruby
queues = []
paused_queues = [
  [Time.at 0, 'queue1'],
  [Time.at 0, 'queue1'],
  [Time.at 0, 'queue1'],
  [Time.at 0, 'queue1'],
]
```

Then the next call to `#next_queue` will still unpause just one instance:

```ruby
queues = %w(queue1)
paused_queues = [
  [Time.at 0, 'queue1'],
  [Time.at 0, 'queue1'],
  [Time.at 0, 'queue1'],
]
```

The only difference is that all entries in `paused_queues` can be returned back into `queues` before waiting for the original full delay time, since each iteration of the `dispatch` loop will move one instance over.

In practice, I don't think there's really a difference: in the case where we had a queue with a maximum weight greater than 1, and there was a period of time where the queue was empty, then suddenly a bunch of work appears on the queue, I think both strategies will allow consumption to pick up quickly again.

If anything, doing this deep dive reveals an opportunity for future improvement:

Maybe `paused_queues` shouldn't actually contain multiple instances of the same queue. This behavior doesn't really seem useful and it seems like it might actually make it possible to increase the queue weight to be greater than the user-configured maximum weight. Also, it seems to violate the expectation of the `delay` property if the weight is greater than 1. Specifically, if there are no messages in the queue, as a user, I expect no more than one call to fetch messages from that queue every `delay` seconds, but in practice, each instance in `paused_queues` has its own `delay` timer.

cjlarose: @davidrichey I tried writing a test to verify that it's possible for the same queue to appear multiple times in `@paused_queues`, but I think I discovered that it's actually not possible; I misunderstood what was happening. `#pause` is the only method that adds entries to `@paused_queues`. It is defined like so:

```ruby
      def pause(queue)
        return unless @queues.delete(queue)

        @paused_queues << [Time.now + delay, queue]
        logger.debug "Paused #{queue}"
      end
```

It starts by deleting _all_ instance of the queue from `@queues`, then adding a _single_ entry to `@paused_queues`. Because there are no longer any references to the queue in `@queues`, it doesn't seem possible for `#messages_found` to be invoked for the paused queue (so it doesn't seem paused to call `#pause` on an already-paused queue). I think the code is good to go as-is.

PR #689, opened 2021-12-01
Add throttling support to StrictPriority

### LINKED ISSUE #535: Throttling
Hello there,

Very quick question here: is there a way to throttle fetching of messages from SQS queues ? I'm looking for functionality similar to [sidekiq-throttled](https://github.com/sensortower/sidekiq-throttled#usage) here.
If throttling isn't possible just now - do you expect it to be on the roadmap at some point down the line?

Thanks!

rafibaum:
This PR adds support for an `interval` configuration option for queues and groups when using the `StrictPriority` polling strategy. The way `interval` works is that, when defined, Shoryuken will put a queue to sleep for `n * interval` seconds after fetching `n` messages. We wanted to implement this since we have a use case where we need to throttle how quickly Shoryuken pulls from specific queues and adding a configuration option like this would make that possible.

This is just my first shot at the implementation so I'll happily take feedback, or we can close this PR if this functionality it better served in a custom polling strategy rather than a patch to `StrictPriority`. The biggest implementation design question I have right now is what the best way would be to pass the `interval` value into `StrictPriority`? Right now I have it being passed through similarly to `delay` but I'm not sure if it makes sense to bloat classes with arbitrary configuration options. Let me know if you have any thoughts on this.

Closes #535

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -88,7 +88,8 @@ module Shoryuken
     :on,
     :cache_visibility_timeout?,
     :cache_visibility_timeout=,
-    :delay
+    :delay,
+    :interval
   )
 end
 
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -124,7 +124,7 @@ module Shoryuken
       end
 
       Shoryuken.options[:groups].to_a.each do |group, options|
-        Shoryuken.add_group(group, options[:concurrency], delay: options[:delay])
+        Shoryuken.add_group(group, options[:concurrency], delay: options[:delay], interval: options[:interval])
 
         options[:queues].to_a.each do |queue, weight|
           parse_queue(queue, weight, group)
--- a/lib/shoryuken/launcher.rb
+++ b/lib/shoryuken/launcher.rb
@@ -88,10 +88,12 @@ module Shoryuken
 
     def create_managers
       Shoryuken.groups.map do |group, options|
+        polling_options = { interval: Shoryuken.interval(group) }
+
         Shoryuken::Manager.new(
           group,
           Shoryuken::Fetcher.new(group),
-          Shoryuken.polling_strategy(group).new(options[:queues], Shoryuken.delay(group)),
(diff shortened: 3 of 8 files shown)


cjlarose:
I'm open to this change and I appreciate you tackling this problem. In general, I prefer to implement resource management by limiting concurrency and not by introducing latency or adding restrictions on throughput, but in Shoryuken's current design it's not really practical to implement a concurrency limit like 0.5 erlangs, which would mean something like on average, the number of shoyuken threads occupied by a message from a particular queue is 0.5. At best, with separate processing groups, you can limit a queue to 1 erlang.

Adding a sleep time between polling for messages on the queue puts some burden on the application developer because if their goal was to limit concurrency, but the only tool they have is to increase latency with this new setting, they'd have to do some math and take into consideration how long the job normally takes, and this approach wouldn't respond well to situations where the job starts to take longer (e.g., a downstream resource is responding more slowly than usual).

All of that said though, this change is a definite improvement because it allows folks to do something they couldn't before and we can always improve things later.

> The biggest implementation design question I have right now is what the best way would be to pass the `interval` value into `StrictPriority`? Right now I have it being passed through similarly to `delay` but I'm not sure if it makes sense to bloat classes with arbitrary configuration options. Let me know if you have any thoughts on this.

`delay`, in my mind, isn't the best thing to use as a model for the new `interval` setting. `delay` is a setting that affects the entire processing group, but it's my expectation that the new `interval` setting is more useful to configure on an individual queue level. `weight` is the closest existing example of a per-queue setting, so I'd use that as a model to follow. One thing that's weird about that, though, is that to configure weights in the YAML configuration format, you have to specify queues as arrays:

```yaml
queues:
- [queue_1, 4]
```

But if we add `interval` as a per-queue setting in the Array format, it's not very readable:

```yaml
queues:
- [queue_1, 4, 5]
```

So I'm open to a change that would make it possible to specify queues as Hashes.

```yaml
queues:
- name: queue_1
  weight: 5
  interval: 3
```

Lastly, I'm not crazy about the name `interval`. I don't have a perfect name for this, so I'm open to suggestions, but one that might work is`fetching_throttle_seconds`.

PR #707, opened 2022-06-29
Fix simply chained nil-safe call

taakuuyaa:


phstc:
@taakuuyaa thank you!

taakuuyaa:
Sorry, spec check was omitted.

Issue #719, opened 2022-11-08
ActiveSupport::CurrentAttributes parameters resetting after being set in server middleware


saksham-jain:
- I am using Shoryuken for ActiveJob.
- Using ActiveSupport::CurrentAttributes to store **firm_name** in a multi-tenant application. This firm_name is coming from the SQS message attribute and is stored in CurrentAttributes to keep it available through the job processing. 
This flow works fine with sidekiq. But with the Shoryuken, after **firm_name** is set in middleware, the ActiveSupport reloader runs which resets all CurrentAttributes. 
- This issue was strange until I look into Sidekiq's process method(in processor.rb). Sidekiq runs the reloader before the server_middlewares, whereas with Shoryuken reloader runs after server_middlewares. 

**Sidekiq process method-**
```
dispatch(job_hash, queue, jobstr) do |inst|    # ----> this dispatch method runs reloader
   config.server_middleware.invoke(inst, job_hash, queue) do 
      execute_job(inst, job_hash["args"]) 
   end
end

#reloader running before server_middleware
```

**Shoryuken process method-**
```
Shoryuken::Logging.with_context("#{worker_name(worker.class, sqs_msg, body)}/#{queue}/#{sqs_msg.message_id}") do
   worker.class.server_middleware.invoke(worker, queue, sqs_msg, body) do
      worker.perform(sqs_msg, body) # ----> this ultimately calls active job which has reloader callback
   end
end

#reloader running after server_middleware
```

- The proposed fix is to introduce a reloader around server_middleware in Shoryuken similar to Sidekiq. Like this-
```
::Rails.application.reloader.wrap do  # ----> run reloader before server_middleware
  Shoryuken::Logging.with_context("#{worker_name(worker.class, sqs_msg, body)}/#{queue}/#{sqs_msg.message_id}") do
    worker.class.server_middleware.invoke(worker, queue, sqs_msg, body) do
        worker.perform(sqs_msg, body) 
    end
  end
end
```

- This fixes the issue and ActiveSupport::CurrentAttributes doesn't reset after being set in server_middleware

matt-taylor:
Hi @saksham-jain 

Thanks for the detail here. One thing to note is that Shoryuken (And Sidekiq) can be used without rails so we would need to be cognizant of that for any change that gets made.

---

[...]

PR #721, opened 2022-11-15
Fix for ActiveSupport::CurrentAttributes resetting issue

saksham-jain:
- Wrap processor's process code around rails reloader to prevent resetting of current attributes after they are assigned in server middleware
- Linked to issue #719


matt-taylor:
Hey @saksham-jain Is this still needed?

saksham-jain:
> Hey @saksham-jain Is this still needed?

Hey @matt-taylor, I have this change in my codebase. But still, it would be better to have some generalized solution in Shoryuken to fix the issue.

PR #727, opened 2023-02-22
Fix spec for ruby 3.0

yyamanoi1222:
Fixed some tests were failing when running rspec on ruby 3.0

### DIFF
--- a/spec/shoryuken/fetcher_spec.rb
+++ b/spec/shoryuken/fetcher_spec.rb
@@ -28,12 +28,12 @@ RSpec.describe Shoryuken::Fetcher do
 
       Shoryuken.sqs_client_receive_message_opts[group] = { wait_time_seconds: 10 }
 
-      expect(queue).to receive(:receive_messages).with(
+      expect(queue).to receive(:receive_messages).with({
         wait_time_seconds: 10,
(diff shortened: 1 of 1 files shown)


spec/shoryuken/fetcher_spec.rb
```diff
@@ -28,12 +28,12 @@
 
       Shoryuken.sqs_client_receive_message_opts[group] = { wait_time_seconds: 10 }
 
-      expect(queue).to receive(:receive_messages).with(
+      expect(queue).to receive(:receive_messages).with({
```
andreleoni, line 31: why does this is a problem in ruby 3? 🤔

yyamanoi1222: I'm probably having this same problem, and the test fails in my environment (ruby 3.0.0)
https://github.com/lsegal/yard/pull/1433

ZeroInputCtrl: @andreleoni Keyword arguments are fully separated from hashes in ruby 3.x. https://rubyreferences.github.io/rubychanges/3.0.html#keyword-arguments-are-now-fully-separated-from-positional-arguments

phstc (approved):

PR #733, opened 2023-04-26
Add default body value to worker perform methods

AlexandreL0pes:
### Context
Recently, I’ve been working on a new shoryuken worker and this particulary one does not require an argument to be scheduled and executed. 

For this scenario, the solution was give a empty string as the argument, since the body argument it’s required, something like this:

```ruby
class Job
  include Shoryuken::Worker
	
  shoryuken_options queue: "queue-name", auto_delete: true

  def perform(_message, _body)
  end
end

Job.perform_async(" ")
```

### Suggestion
The idea with this PR is turn the body argument into optional, passing a string with a blank space as the default value, because it’s not possible to send a message without body, according to the aws sqs client ([link](https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/SQS/Client.html#send_message-instance_method)).

With the changes, workers could be called like:
```ruby
class Job
  include Shoryuken::Worker
	
  shoryuken_options queue: "queue-name", auto_delete: true

  def perform(_message, _body)
  end
end

Job.perform_async
```
Let me know if this makes sense, happy to close if not!

### DIFF
--- a/lib/shoryuken/worker.rb
+++ b/lib/shoryuken/worker.rb
@@ -6,11 +6,11 @@ module Shoryuken
     end
 
     module ClassMethods
-      def perform_async(body, options = {})
+      def perform_async(body = ' ', options = {})
         Shoryuken.worker_executor.perform_async(self, body, options)
       end
 
-      def perform_in(interval, body, options = {})
+      def perform_in(interval, body = ' ', options = {})
         Shoryuken.worker_executor.perform_in(self, interval, body, options)
       end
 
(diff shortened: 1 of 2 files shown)


phstc:
@AlexandreL0pes @andreleoni Olá! Shoryuken tries to be more transparent with SQS, and SQS does not accept empty bodies. A change like this may require a major version bump to avoid breaking changes. 

Are you migrating an existing code base with Sidekiq (empty body) to Shoryuken, or is it more a dev experience suggestion?

AlexandreL0pes:
Olá, @phstc!! It's a dev experience suggestion .

This is a new worker, created since some latency issues were identified, so this slow method has been converted into a shoryuken worker.

AlexandreL0pes:
Okay, I'm closing this PR and later I'll add this topic to the migration guide.

Thanks for the clarifications and congratulations for this awesome project!

PR #738, opened 2023-06-09
Batches greater than 10

sinecode:
Hello!

First of all, thank you for providing this excellent gem!

This pull request introduces a feature that allows batches with a size greater than 10. To achieve this, I have added an additional worker parameter called `batch_options`. With `batch_options`, users can specify the desired maximum batch size and batch timeout. The batch timeout ensures that we don't have to wait indefinitely to reach the desired batch size.

To handle the scenario where multiple calls are made to SQS in a single fetch, I have introduced a check on the queue visibility timeout. If the queue visibility timeout is lower than the batch timeout, a warning message is logged. This prevents the situation where fetched but unprocessed messages are returned to the queue.

Here's an example of how to use the new feature:

```ruby
shoryuken_options(
   queue: 'queue_name',
   batch: true,
   batch_options: {
     max_size: 15,
     timeout: 100  # in seconds
   }
)
```

Currently, I am already running several workers that process batches greater than 10 with this algorithm using a custom framework outside of Shoryuken. However, I believe that including this feature within the gem would be beneficial, as I use Shoryuken for all other workers that don't require batches greater than 10.

I would greatly appreciate your feedback on this proposal. Please let me know if there are any considerations or conflicts with existing Shoryuken features that I may have overlooked. If everything seems acceptable, I am more than willing to extensively test this feature in real-world code.


lib/shoryuken/core_ext.rb
```diff
@@ -47,6 +59,33 @@ def constantize
...
+
+      if block_given?
+        collection.each_slice(number, &block)
+      else
+        collection.each_slice(number).to_a
+      end
+    end
+  end
+end
+
```
sinecode, line 88: Brutally copied from ActiveRecord: https://apidock.com/rails/v5.2.3/Array/in_groups_of

lib/shoryuken/fetcher.rb
```diff
@@ -44,6 +48,36 @@ def fetch_with_auto_retry(max_attempts)
...
+        sqs_msgs = Array(receive_messages(queue, [limit, batch.max_size, batch.max_size - batch.size].min))
+        logger.debug { "Found #{sqs_msgs.size} messages for #{queue.name}" } unless sqs_msgs.empty?
+        sqs_msgs.each { |sqs_msg| batch.add_message!(sqs_msg) }
+        break if batch.full? || batch.timeout_expired?
+      end
+      batch.messages
+    end
+
+    def check_visibility_timeout(queue, batch_timeout)
+      return if @visibility_timeout_ok
```
phstc, line 70: nit: If `@visibility_timeout_ok` is being used for memorization, we usually suffix with `_`. 

```suggestion
      return if @_visibility_timeout_ok
```

lib/shoryuken/fetcher.rb
```diff
@@ -44,6 +48,36 @@ def fetch_with_auto_retry(max_attempts)
...
+    def regular_fetch(queue, limit)
+      sqs_msgs = Array(receive_messages(queue, limit))
+      logger.debug { "Found #{sqs_msgs.size} messages for #{queue.name}" } unless sqs_msgs.empty?
+      sqs_msgs
+    end
+
+    def fetch_with_batch_options(queue, limit, batch_max_size, batch_timeout)
+      batch = Shoryuken::MessageBatch.new(max_size: batch_max_size, timeout: batch_timeout)
+      loop do
+        check_visibility_timeout(queue, batch.timeout)
```
phstc, line 60: Does this need to be inside of the loop? It seems the first time it gets true. It gets memorized and always return.

phstc:
Hi @sinecode

Thanks for your contribution. The code looks solid.

Shoryuken tries to be as transparent as possible with SQS. I'm not sure if increasing the batch size is a fit for Shoryuken.

The visibility timeout worries me. I understand you add some control around that, but still, it's not something we have in SQS. 

Could you give a more specific example of your use case? I understand that we may need to process more than 10 at a time. But I'm wondering if there are other solutions instead of increasing the batch limit.

lib/shoryuken/middleware/server/auto_delete.rb
```diff
@@ -2,14 +2,18 @@ module Shoryuken
...
 
           return unless worker.class.auto_delete?
 
-          entries = [sqs_msg].flatten.map.with_index { |message, i| { id: i.to_s, receipt_handle: message.receipt_handle } }
+          [sqs_msg].flatten.in_groups_of(MAX_BATCH_SIZE, false).each do |sqs_msgs|
+            entries = sqs_msgs.map.with_index { |message, i| { id: i.to_s, receipt_handle: message.receipt_handle } }
 
-          Shoryuken::Client.queues(queue).delete_messages(entries: entries)
+            Shoryuken::Client.queues(queue).delete_messages(entries: entries)
+          end
```
phstc, line 16: This can also cause duplicate processing depending on the volume. If it takes too long to delete a large batch, messages can go back and be consumed again.

lib/shoryuken/fetcher.rb
sinecode: Yes, the check can be performed only once outside of the loop. I decided to put it inside the loop to be more aggressive with the warning log, but eventually, the loop will end so we'll come back on the check multiple times anyway. Thanks for the suggestion.

lib/shoryuken/middleware/server/auto_delete.rb
sinecode: That's true, but isn't this true also for any message processing? Even if you're using a regular worker, if it takes too long to process the messages the visibility timeout can expire. I think that the visibility timeout must be carefully chosen in any case.

sinecode:
> Hi @sinecode
> 
> Thanks for your contribution. The code looks solid.
> 
> Shoryuken tries to be as transparent as possible with SQS. I'm not sure if increasing the batch size is a fit for Shoryuken.
> 
> The visibility timeout worries me. I understand you add some control around that, but still, it's not something we have in SQS.
> 
> Could you give a more specific example of your use case? I understand that we may need to process more than 10 at a time. But I'm wondering if there are other solutions instead of increasing the batch limit.

Hi @phstc, thanks a lot for your feedback.

In order to optimize performance and minimize the number of API calls, I usually require batches consisting of more than 10 messages, ideally around 50-150 messages. Working with smaller batches often results in excessive API calls or underutilization of batch processing capabilities (think of GPUs).

Alternatively, one might consider storing the messages in a secondary data structure such as Redis. However, this approach introduces challenges in managing the lifecycle of the messages. It would require handling both the SQS queue and the additional data structure simultaneously. Therefore, I think that we can directly use only the SQS queue as support.

I totally get your point. If you believe that this approach deviates too much from SQS, please feel free to close this PR. I really appreciated your feedback anyway.

PR #739, opened 2023-08-15
Upgrade test matrix

noelrappin:
Upgrade test matrix to include Ruby 3.1, Ruby 3.2 and Rails 7


spec/spec_helper.rb
```diff
@@ -2,7 +2,7 @@
 Bundler.setup
 
 begin
-  require 'pry-byebug'
+  require 'pry-byebug' if RUBY_VERSION <= '3.2'
```
phstc, line 5: @noelrappin is pry-byebug incompatible with 3.2? 

I was checking their [gemspec](https://github.com/deivid-rodriguez/pry-byebug/blob/master/pry-byebug.gemspec), and it seems to be ok.

noelrappin: I got this error

```
NameError:
  undefined method `=~' for class `Pry::Code'
```

noelrappin: Might be a version thing, actually

noelrappin: let me check

noelrappin: The tricky part seems to be getting a version of pry and byebug that works for all the ruby versions

Gemfile
```diff
@@ -15,6 +15,6 @@ end
 
 group :development do
   gem 'appraisal', git: 'https://github.com/thoughtbot/appraisal.git'
-  gem 'pry-byebug', '3.9.0'
+  gem 'pry-byebug'
```
noelrappin, line 18: Needed to unpin pry-byebug to let it find the right version for each Ruby version

phstc (approved):

spec/spec_helper.rb
phstc: Got it, the problem was with the version of pry-byebug as you suggested. Great catch!

gemfiles/rails_7_0.gemfile
```diff
@@ -0,0 +1,22 @@
...
+  gem "httparty"
+  gem "multi_xml"
+  gem "simplecov"
+end
+
+group :development do
+  gem "appraisal", git: "https://github.com/thoughtbot/appraisal.git"
+  gem "rubocop"
+  gem "pry", ">= 0.14.2"
+  gem "pry-byebug", ">= 3.10.1"
```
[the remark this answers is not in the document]
noelrappin: You need a minimum, but it'd probably find it on it's own, really. It's much less of an issue in the Rails specs since each one is different, then the Ruby specs, that share a gemfile.

PR #741, opened 2023-09-04
fire stopped event after executor is stopped

rajyan:
adds a new event "stopped" after stopping the executor.

closes https://github.com/ruby-shoryuken/shoryuken/issues/740


phstc:
Looks solid! Thanks for the contribution.

rajyan:
Thank you for your quick responses!

Issue #308, opened 2017-01-26
How to get more detailed log for Shoryuken?


sony-mathew:
In my shoryuken workers log, I'm getting the error trace like this :

```
2017-01-26T22:41:50Z 30706 TID-oupijs4bg DEBUG: Ready: 7, Busy: 3, Active Queues: []
2017-01-26T22:41:50Z 30706 TID-oupijs4bg DEBUG: Pausing fetcher, because all queues are paused
2017-01-26T22:41:50Z 30706 TID-oupjp7ovk Pushes::PollingViewsWorker/development_push_views/batch-with-0-messages INFO: failed in: 7000.449 ms
2017-01-26T22:41:50Z 30706 TID-oupjp7ovk ERROR: Actor crashed!
NoMethodError: undefined method attributes' for []:Array
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/server/exponential_backoff_retry.rb:23:in handle_failure'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/server/exponential_backoff_retry.rb:13:in rescue in call'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/server/exponential_backoff_retry.rb:8:in call'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/chain.rb:96:in block in invoke'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/server/timing.rb:14:in block in call'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/logging.rb:21:in with_context'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/server/timing.rb:8:in call'
 /Users/SJ/.rvm/gems/ruby-2.2.4@projectx/gems/shoryuken-2.1.2/lib/shoryuken/middleware/chain.rb:96:in block in invoke'
[...]

PR #742, opened 2023-09-04
Allow setup custom exception handlers for failing jobs

nbulaj:
Allow to setup custom exception handlers in case anybody needs more actions and/or logging data when processor fails.

Idea with multiple handlers [taken from the Sidekiq ](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/config.rb#L268)

Fixes also to https://github.com/ruby-shoryuken/shoryuken/issues/544 & #308

### DIFF
--- a/lib/shoryuken/processor.rb
+++ b/lib/shoryuken/processor.rb
@@ -22,8 +22,7 @@ module Shoryuken
         end
       end
     rescue Exception => ex
-      logger.error { "Processor failed: #{ex.message}" }
-      logger.error { ex.backtrace.join("\n") } unless ex.backtrace.nil?
+      Array(Shoryuken.exception_handlers).each { |handler| handler.call(ex, queue, sqs_msg) }
 
       raise
     end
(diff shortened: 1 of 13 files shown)


phstc:
Solid! Thanks for the contribution.

lib/shoryuken/processor.rb
```diff
@@ -22,8 +22,7 @@ def process
         end
       end
     rescue Exception => ex
-      logger.error { "Processor failed: #{ex.message}" }
-      logger.error { ex.backtrace.join("\n") } unless ex.backtrace.nil?
+      Array(Shoryuken.exception_handlers).each { |handler| handler.call(ex, queue, sqs_msg) }
```
phstc, line 25: Is there a way you can test this too? Maybe here https://github.com/ruby-shoryuken/shoryuken/blob/main/spec/shoryuken/processor_spec.rb

nbulaj: This actually is tested in `spec/shoryuken/default_exception_handler_spec.rb` with InlineExecutor, but I can add more tests

phstc: nvm I misread that. It looks great! Thank you again.

PR #750, opened 2023-10-20
Stop the dispatching of new messages when a SIGTERM signal has been received

### LINKED ISSUE #749: SIGTERM does not stop dispatching of new messages from the queue
After sending `SIGTERM` to Shoryuken, messages are still fetched from SQS.

I have found [`Launcher#stop!`](https://github.com/ruby-shoryuken/shoryuken/blob/main/lib/shoryuken/launcher.rb#L16), which does not stop the dispatching other than `Launcher#stop`.
I am not sure why. Doesn't it make sense to also stop the dispatching in this case?

When running the Shoryuken Worker on AWS ECS there is always a `SIGTERM` when the ECS task is stopping and waits up to 30 seconds by default until it is force killed: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-lifecycle.html#lifecycle-states.
So i cannot use the `SIGUSR1` signal.

hoffi:
Fixes #749

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,6 @@ gemspec
 
 group :test do
   gem 'activejob'
-  gem 'activesupport'
   gem 'aws-sdk-core', '~> 3'
   gem 'aws-sdk-sqs'
   gem 'codeclimate-test-reporter', require: nil
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ gemspec
 
 group :test do
   gem 'activejob'
+  gem 'activesupport'
   gem 'aws-sdk-core', '~> 3'
   gem 'aws-sdk-sqs'
   gem 'codeclimate-test-reporter', require: nil
--- a/lib/shoryuken/extensions/active_job_adapter.rb
+++ b/lib/shoryuken/extensions/active_job_adapter.rb
@@ -66,8 +66,11 @@ module ActiveJob
         }
 
         if queue.fifo?
-          # See https://github.com/phstc/shoryuken/issues/457
-          msg[:message_deduplication_id] = Digest::SHA256.hexdigest(JSON.dump(body.except('job_id')))
+          # See https://github.com/ruby-shoryuken/shoryuken/issues/457 and
+          # https://github.com/ruby-shoryuken/shoryuken/pull/750#issuecomment-1781317929
+          msg[:message_deduplication_id] = Digest::SHA256.hexdigest(
+            JSON.dump(body.except('job_id', 'enqueued_at'))
+          )
         end
 
         msg.merge(job_params.except(:message_attributes))
--- a/lib/shoryuken/launcher.rb
+++ b/lib/shoryuken/launcher.rb
@@ -16,6 +16,9 @@ module Shoryuken
     def stop!
       initiate_stop
 
+      # Don't await here so the timeout below is not delayed
+      stop_new_dispatching
+
       executor.shutdown
       executor.kill unless executor.wait_for_termination(Shoryuken.options[:timeout])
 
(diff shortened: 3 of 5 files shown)


phstc:
@hoffi thanks for the submission. That's greatly appreciated. 

Some specs are failing but don't seem related to your change. I don't recall those being flaky before. But it seems they are also failing here https://github.com/ruby-shoryuken/shoryuken/pull/747

Can you have a look?

hoffi:
The `enqueued_at` timestamp in `job.serialize` differs slightly from the value in `hash` here: https://github.com/ruby-shoryuken/shoryuken/blob/main/spec/shared_examples_for_active_job.rb#L47

I have looked into the `#serialize` method from ActiveJob and it always returns a new timestamp: https://github.com/rails/rails/blob/7-1-stable/activejob/lib/active_job/core.rb#L122

But the correct `enqueued_at` timestamp is not stored in the job object, so i'm not sure where to get it. Should we use something like `ActiveSupport::Testing::TimeHelpers#freeze_time` to ensure the same timestamp is used?

phstc:
@hoffi `freeze_time` seems like a good idea, WDYT? It seems the `hash` requires more changes. 

I don't know why it's only happening now 🤷

hoffi:
@phstc i have added the `freeze_time` helper which fixed the test on my machine

phstc:
> @phstc i have added the `freeze_time` helper which fixed the test on my machine

That's awesome! Thank you

https://github.com/ruby-shoryuken/shoryuken/actions/runs/6644142158/job/18052725694?pr=750#step:4:34

```
undefined method `freeze_time' for #<RSpec::ExampleGroups::ActiveJobQueueAdaptersShoryukenAdapter::Enqueue::WhenFifo:0x00555b2f3f[34](https://github.com/ruby-shoryuken/shoryuken/actions/runs/6644142158/job/18052725694?pr=750#step:4:35)48>
     Shared Example Group: "active_job_adapters" called from ./spec/shoryuken/extensions/active_job_adapter_spec.rb:6
     # ./spec/shared_examples_for_active_job.rb:49:in `block (4 levels) in <top (required)>'
```

Unfortunately, it's failing for Rails 4. Is there any alternative we could use? Maybe https://github.com/travisjeffery/timecop?

hoffi:
Yes, i have seen now it is only available since rails 5.

However i just realized that it might be problematic to include the enqueued_at timestamp in the deduplication id?

isn‘t it the same as with #457?

phstc:
@hoffi hm for standard works, we use only the body https://github.com/ruby-shoryuken/shoryuken/blob/f02d28b4bf79258233e977736d1fd878e203a687/lib/shoryuken/queue.rb#L133

It's probably OK to remove `enqueued_at`. The uniqueness should be based on what the user sent, not on what was generated.

It makes me a bit uncomfortable to change something like that after so long (especially given that it was working before), but rationally thinking makes it feel safe. 

I'm up for either change: Timecop or removing `enqueued_at`. 

If you prefer the `enqueued_at`, can you test it locally to check if the deduplication works?

hoffi:
> It makes me a bit uncomfortable to change something like that after so long (especially given that it was working before), but rationally thinking makes it feel safe.

Yes however it was changed in rails 6 that the `enqueued_at` timestamp is returned in the message body:
* https://github.com/rails/rails/blob/5-2-stable/activejob/lib/active_job/core.rb#L81
* vs. https://github.com/rails/rails/blob/6-0-stable/activejob/lib/active_job/core.rb#L94

> If you prefer the enqueued_at removal, can you test it locally to check if the deduplication works?

I will try to see if i can test it

phstc:
> Yes however it was changed in rails 6 that the enqueued_at timestamp is returned in the message body

@hoffi if that's the case, let's remove it then. It feels way safer given that ^ the deduplication was added prior to Rails 6.

Thanks for thoroughly reviewing this.

hoffi:
@phstc i have removed the timestamp. Can sou take a look at it?

phstc:
@hoffi looks great! Awesome work. Thank you

hoffi:
Nice! Thanks :)

PR #753, opened 2023-11-21
Adds nullish coalescing to output of Aws::SQS::Client#receive_message to an empty array

### LINKED ISSUE #752: Exception in Shoryuken::Queue#receive_messages
Last night, we saw our Shoryuken processes begin to fail with the below error. It appears that the `aws-ruby-sdk` SQS client's `receive_message` function is sometimes returning an empty array and sometimes returning `nil`. It's unclear if this is a change in the `aws-sdk` gem behavior or the AWS APIs. My guess would be the latter since this happened overnight in between releases to our application.

```
Shoryuken -- Manager failed: undefined method `map' for nil:NilClass
.../shoryuken-6.1.0/lib/shoryuken/queue.rb:46:in `receive_messages'
```

We've monkey patched this in our application for the time being:

```rb
module Shoryuken
  class Queue
    def receive_messages(options)
      messages = client.receive_message(options.merge(queue_url: url)).messages || []
      messages.map { |m| Message.new(client, self, m) }
    end
  end
end
```

mkeemon:
Fixes #752 due to change in behavior of `aws-sdk-sqs` (https://github.com/aws/aws-sdk-ruby/issues/2947)

- This adds a default value of `[]` to all invocations of `Aws::SQS::Client#receive_message`
- The underlying behavior in the AWS Ruby SDK appears to have been patched, but this will prevent regressions in the future should something similar happen again

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,10 @@ gemspec
 group :test do
   gem 'activejob'
   gem 'aws-sdk-core', '~> 3'
-  gem 'aws-sdk-sqs'
+  # Pin to 1.65.0 because of below issues:
+  # - https://github.com/ruby-shoryuken/shoryuken/pull/753#issuecomment-1822720647
+  # - https://github.com/getmoto/moto/issues/7054
+  gem 'aws-sdk-sqs', '1.65.0'
   gem 'codeclimate-test-reporter', require: nil
   gem 'httparty'
   gem 'multi_xml'
--- a/bin/cli/sqs.rb
+++ b/bin/cli/sqs.rb
@@ -117,7 +117,7 @@ module Shoryuken
               max_number_of_messages: batch_size,
               attribute_names: ['All'],
               message_attribute_names: ['All']
-            ).messages
+            ).messages || []
 
             messages.each { |m| yield m }
 
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -43,7 +43,8 @@ module Shoryuken
     end
 
     def receive_messages(options)
-      client.receive_message(options.merge(queue_url: url)).messages.map { |m| Message.new(client, self, m) }
+      messages = client.receive_message(options.merge(queue_url: url)).messages || []
+      messages.map { |m| Message.new(client, self, m) }
     end
 
     def fifo?


mkeemon:
@phstc any ideas on why `moto` and `Aws::SQS::Client` are now at odds? Moto is returning XML while the client expects JSON. Did you run into any similar issues when creating the newest release a few weeks ago?

**Moto**
- Python version: 3.12.0
- Moto Version: 4.2.9

**AWS Client**
- aws-sdk-core: 3.187.1
- aws-sdk-sqs: 1.67.0

```
Failure/Error: Shoryuken::Client.sqs.delete_queue(queue_url: queue_url)
     Aws::Json::ParseError:
       unexpected token at '<DeleteQueueResponse><ResponseMetadata><RequestId>qYSGBzaVm7ABwMZ9tT15MD8XDLycSkezkGhzDSFKhCmgCc68O3re</RequestId></ResponseMetadata></DeleteQueueResponse>'
```

**Local Docker Image**
```
$ pip show moto
Name: moto
Version: 4.2.9
Summary: 
Home-page: https://github.com/getmoto/moto
Author: Steve Pulec
Author-email: "spulec@gmail.com"
License: Apache License 2.0
Location: /usr/local/lib/python3.12/site-packages
Requires: boto3, botocore, cryptography, Jinja2, python-dateutil, requests, responses, werkzeug, xmltodict
```

phstc:
@mkeemon thanks for checking it. The last release passed ✅ fine https://github.com/ruby-shoryuken/shoryuken/pull/751

image


I'm wondering if there's something with the AWS change, and now we need to specify the type to be JSON instead of XML somewhere [here](https://github.com/ruby-shoryuken/shoryuken/blob/d65636ebacd6a33b68cdc7f539aa9ce6a50eef9f/spec/integration/launcher_spec.rb#L8-L13). 

I couldn't find an option on the [initializer](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html#initialize-instance_method) to configure that though.

mkeemon:
> **What if I am already on the latest AWS SDK version, but my open sourced solution does not support JSON?**
> You must change your SDK version to the version previous to what you are using. See [How do I get started with AWS JSON protocols for Amazon SQS?](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-json-faqs.html#json-protocol-getting-started) for more infomation. AWS SDK versions listed in [How do I get started with AWS JSON protocols for Amazon SQS?](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-json-faqs.html#json-protocol-getting-started) uses JSON wire protocol for Amazon SQS APIs. If you change your AWS SDK to the previous version, your Amazon SQS APIs will use the AWS query.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-json-faqs.html

It looks like 1.67.0 of `aws-sdk-sqs`, which came out a couple weeks ago, moves to JSON. I'll try pinning the version to 1.66.0 for now and open up an issue in Moto repo to migrate the responses to JSON

phstc:
@mkeemon great find! Re-running tests. Thanks for helping with this. I really appreciate it.

Gemfile
```diff
@@ -6,7 +6,7 @@ gemspec
 group :test do
   gem 'activejob'
   gem 'aws-sdk-core', '~> 3'
-  gem 'aws-sdk-sqs'
+  gem 'aws-sdk-sqs', '1.65.0'
```
phstc, line 9: @mkeemon if you create an issue on the Moto side, could also link it here in the comment?

```suggestion
  # 1.65.0 works with XML + Moto
  # See https://github.com/ruby-shoryuken/shoryuken/pull/753#issuecomment-1822720647
  gem 'aws-sdk-sqs', '1.65.0'
```

mkeemon:
@phstc done and done!

mscrivo:
I'm a bit curious, why the need to pin the version of aws sdk for tests if this PR fixes the issue with the newer versions? Won't that just prevent finding issues with newer versions of the sdk in the future?

mkeemon:
@mscrivo This PR was merged _slightly_ before the one in Moto. I didn't realize just how quickly the Moto team would take action on this. We should be able to remove the strict versioning in the Gemfile now

phstc:
@mkeemon version 6.1.1 is out with this PR! Thank you 🎉

PR #754, opened 2023-11-27
delete_message_batch failed is returning nil rather than empty array …

JohnMaguir:
…as before, default to empty array if nil returned.

I am seeing errors for `Shoryuken::Queue#delete_messages`
```
Processor failed: undefined method `any?' for nil:NilClass
ERROR: lib/shoryuken/queue.rb:26:in `delete_messages'
```

This simply defaults to an empty array if nil is returned for failed_messages.

Similar fix to #753

### DIFF
--- a/lib/shoryuken/queue.rb
+++ b/lib/shoryuken/queue.rb
@@ -21,9 +21,10 @@ module Shoryuken
     end
 
     def delete_messages(options)
-      client.delete_message_batch(
+      failed_messages = client.delete_message_batch(
         options.merge(queue_url: url)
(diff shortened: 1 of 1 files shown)


marcelohenriq99 (approved):
:rocket:

silasrnc:
@JohnMaguir 
I had the same problem, updated to the latest version, and it worked.

phstc:
@JohnMaguir thank you 🎉

phstc:
Version 6.1.1 is out with your changes! Thank you again!

PR #755, opened 2023-12-05
Fix activerecord 7.1 deprecation warnings

tomgi:
Fixes the following:

<code>DEPRECATION WARNING: Calling `ActiveRecord::Base.clear_active_connections! is deprecated. Please call the method directly on the connection handler; for example: `ActiveRecord::Base.connection_handler.clear_active_connections!`. (called from call at /app/.bundle/ruby/3.2.0/bundler/gems/shoryuken-869fb0c77f8e/lib/shoryuken/middleware/server/active_record.rb:8)</code>

<code>DEPRECATION WARNING: `clear_active_connections!` currently only applies to connection pools in the current role (`writing`). In Rails 7.2, this method will apply to all known pools, regardless of role. To affect only those connections belonging to a specific role, pass the role name as an argument. To switch to the new behavior, pass `:all` as the role name. (called from call at /app/.bundle/ruby/3.2.0/bundler/gems/shoryuken-869fb0c77f8e/lib/shoryuken/middleware/server/active_record.rb:8)</code>

### DIFF
--- a/lib/shoryuken/middleware/server/active_record.rb
+++ b/lib/shoryuken/middleware/server/active_record.rb
@@ -5,7 +5,7 @@ module Shoryuken
         def call(*_args)
           yield
         ensure
-          ::ActiveRecord::Base.clear_active_connections!
+          ::ActiveRecord::Base.connection_handler.clear_active_connections!(:all)
         end
       end
     end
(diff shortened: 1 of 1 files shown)


phstc:
@tomgi thank you! PR merged.

vladhadzhiyski:
Thanks for merging / fixing this. Looks like it hasn't been released yet. Any rough estimate when a new release with this fix can be cut?

phstc:
hi @vladhadzhiyski 

Sorry for the delay on this! I just released a new version.

Issue #720, opened 2022-11-14
Rails Reloading Support for Development Environment


littleali:
Hi,
I am using `Shoryuken` in a Rails app. 
Because it doesn't support autoreload, it means for any change that I make on shoryuken-related code, I should restart the Shoryuken manually and also because everything else is having autoreload in the development environment, I should be aware of which changes need to restart the Shoryuken, so having the autoreload feature in development improves productivity and decreases errors during development. 

I've read both #627 and #658, and It seems there were some memory problems with #627 in the production environment, I'm confused about it as @cjlarose himself suggest to disable autoreload on production using `config.cache_classes = true`.

I tried the solution on #627 and it works properly in the development environment. Is it possible to have this feature just for the development environment? 

Thank you in Advance.

matt-taylor:
Hi @littleali Apologies for the delay here.

Reloading was talked about in a recent issue https://github.com/ruby-shoryuken/shoryuken/issues/719. Seems like Reloading can be useful for the Shoryuken community.

----

I do not fully understand what sort of Memory issues occurred in #627 for it to get abandoned, but it seems like a good starting point. Looks like you are already starting on the work with https://github.com/certsy/shoryuken/pull/1. 
Please reach out if you need any help

Some things to be careful of:
- Spring can cause issues with App relaoder and `cache_classes` Rails option
- Rails is not a guarantee when using Shoryuken. This means the reloader will not be available all the time

littleali:
@matt-taylor, thank you for your response, 
I don't know what memory problem happened in #627 either. 
So we are testing on our staging/production and see if there are any memory-related problems, and then we submit a new PR.

littleali:
@matt-taylor 
After testing the Rails reloader-enabled version with no problem for more than 2 months, I created a new PR (#725) to add this feature to the Shoryuken.

PR #756, opened 2023-12-06
Enable hot-reload via adding ActiveSupport::Reloader delegate

littleali:
This PR is related to #720 and it is same as #627 and #725

# Pull Request Detail (Copied from #627) 
This change adds explicit integration for Rails autoloading. Rails autoloading already partially worked in Shoryuken in that constants that weren't eager loaded were automatically autoloaded. Shoryuken, however, didn't respect config.cache_classes. This meant that once a constant was loaded by the Shoryuken process, it would never be unloaded and reloaded. If you changed the code in a worker class after Shoryuken had loaded it, the change would have no effect until you restarted the Shoryuken process.

The proposed change uses Rails.application.reloader, which takes care of safely reloading constants. This is the same technique employed by [active_job](https://github.com/rails/rails/blob/ab4ab090667e96c0e8a2623e33573ae06be14b54/activejob/lib/active_job/railtie.rb#L42-L44) and [sidekiq](https://github.com/mperham/sidekiq/blob/460ce532714444e186a60d49c9251b93909d4a58/lib/sidekiq/rails.rb#L13-L15) and even explicitly documented in the [Rails guides](https://github.com/rails/rails/blob/ab4ab090667e96c0e8a2623e33573ae06be14b54/guides/source/threading_and_code_execution.md#reloader). The reloader takes the appropriate locks to ensure that no two threads ever simultaneously unload or reload constants. If the application developer modifies code in an autoloaded directory (including the implementation of workers themselves in app/workers), the changes will take effect the next time a job is processed.

# Breaking changes
This change probably warrants a major version bump. The reason for that is that if Shoryuken library code has any reference to objects that belong to autoloaded modules, it's quite easy to run into autoloading errors. If other people's apps are like my own, this is most likely to happen with custom middleware. This is an example from a Rails initializer:

```ruby
    chain.add MyCustomMiddleware
  end
end
```

Assume MyCustomMiddleware is defined in an autoloaded directory (basically anything under app, in addition to directories explicitly added to config.autoload_paths), and that the execution of MyCustomMiddleware#call has unqualified references to autoloaded constants. In this case, if Shoryuken tries to invoke that middleware after the application developer has triggered reloading (by modifying files in autoloaded directories), then autoloading will fail with A copy of MyCustomMiddleware has been removed from the module tree but is still active!.

The fix is to ensure that Shoryuken library code never retains a reference to any object defined in autoloaded modules and whose code has references to autoloaded modules. In the example of custom middleware, the easiest solution is to just attach the middleware to workers instead of the global server_middleware. You can even do this in your base worker class.

```ruby
class ApplicationWorker
  include Shoryuken::Worker

  def self.inherited(klass)
    klass.server_middleware do |chain|
      chain.add MyCustomMiddleware
    end
  end
end
```

This works because the worker-specific middleware chain has a lifetime associated with the worker class. Once unloading happens, the old references to MyCustomMiddleware are unreachable by the Shoryuken process. When a new job comes in, Shoryuken autoloads the worker class, and the inherited block will run, constructing a new instance of the MyCustomMiddleware class through autoloading.

Users upgrading to a version of Shoryuken that includes this change can, in theory, alternatively just turn on config.cache_classes in their Shoryuken server process (even in development). The experience will be the same as before in that classes won't be reloaded, but it might be an easier upgrade path for some.

```ruby
Shoryuken.configure_server do
  Rails.application.configure do
    config.cache_classes = true
  end
end
```

### DIFF
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -93,6 +93,8 @@ module Shoryuken
     :cache_visibility_timeout=,
     :reloader,
     :reloader=,
+    :enable_reloading,
+    :enable_reloading=,
     :delay
   )
 end
--- a/lib/shoryuken.rb
+++ b/lib/shoryuken.rb
@@ -91,6 +91,8 @@ module Shoryuken
     :on,
     :cache_visibility_timeout?,
     :cache_visibility_timeout=,
+    :reloader,
+    :reloader=,
     :delay
   )
 end
--- a/lib/shoryuken/environment_loader.rb
+++ b/lib/shoryuken/environment_loader.rb
@@ -70,6 +70,13 @@ module Shoryuken
             ::Rails.application.config.eager_load = true
           end
         end
+        ::Rails::Application.initializer 'shoryuken.set_reloader_hook' do |app|
+          Shoryuken.reloader = proc do |&block|
+            app.reloader.wrap do
+              block.call
+            end
+          end
+        end
         if Shoryuken.active_job?
           require 'shoryuken/extensions/active_job_extensions'
           require 'shoryuken/extensions/active_job_adapter'
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -17,7 +17,7 @@ module Shoryuken
     }.freeze
 
     attr_accessor :active_job_queue_name_prefixing, :cache_visibility_timeout, :groups,
-                  :launcher_executor,
+                  :launcher_executor, :reloader,
                   :start_callback, :stop_callback, :worker_executor, :worker_registry, :exception_handlers
     attr_writer :default_worker_options, :sqs_client
     attr_reader :sqs_client_receive_message_opts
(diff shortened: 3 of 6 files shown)


phstc:
hi @littleali 

Thank you for the PR explanation! This is very helpful. 

I was wondering if instead of a major bump, we should enable `cache_classes = true` or any other way of making reloading optional. 

```ruby
Shoryuken.configure_server do
  Rails.application.configure do
    config.cache_classes = true
  end
end
```

I believe most people don't define middleware like this

```ruby
  def self.inherited(klass)
    klass.server_middleware do |chain|
      chain.add MyCustomMiddleware
    end
  end
```
I'm worried that the change may take some solid debugging time to figure out the issue with the middleware definition.

Any thoughts?

littleali:
Hi @phstc 

Regarding your comment I've added a flag called `enable_reloading` to control reloading. The default value is false which means this non-breaking change provides more control over reloading behaviour. 
In addition to the code changes, I've also updated the project's wiki to [include documentation on this new flag](https://github.com/ruby-shoryuken/shoryuken/wiki/Enable-hot-reload-for-rails-dev-environment). This should provide guidance to developers on how and when to use this feature.
Finally, I've updated the existing RSpec tests to ensure this new functionality is covered and works as expected.

phstc:
@littleali thank you for the explanation and pull request. The new version has been released with your changes.

PR #759, opened 2024-01-24
Bugfix: Not able to use extended polling strategy

agrajgarg1526:
Linked to issue https://github.com/ruby-shoryuken/shoryuken/issues/760

### DIFF
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -69,7 +69,7 @@ module Shoryuken
         Polling::StrictPriority
       when String
         begin
-        Object.const_get(strategy)
+          Object.const_get(strategy)
         rescue NameError
           raise ArgumentError, "#{strategy} is not a valid polling_strategy"
         end
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -67,6 +67,8 @@ module Shoryuken
         Polling::StrictPriority
       when String
         Object.const_get(strategy)
+      when Class
+        strategy
       else
         raise ArgumentError, "#{strategy} is not a valid polling_strategy"
       end
--- a/lib/shoryuken/options.rb
+++ b/lib/shoryuken/options.rb
@@ -65,8 +65,8 @@ module Shoryuken
         Polling::WeightedRoundRobin
       when 'StrictPriority'
         Polling::StrictPriority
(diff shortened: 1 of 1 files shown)


lib/shoryuken/options.rb
```diff
@@ -65,8 +65,8 @@ def polling_strategy(group)
         Polling::WeightedRoundRobin
       when 'StrictPriority'
         Polling::StrictPriority
-      when Class
-        strategy
+      when String
```
phstc, line 68: @agrajgarg1526 would this change cause issues for users defining a `polling_strategy` with Ruby just passing a class?

agrajgarg1526: @phstc I am not aware of how can I pass a whole class to `polling_strategy` in yaml file configuration. But even for that case, we can add ` when String
      Object.const_get(strategy)` as a separate case above `when Class
        strategy`. This will not break anything. Thoughts?

phstc: @agrajgarg1526 that's a good point. Can users can access `groups` and mutate it (adding `polling_strategy`)? If so, although unlikely, changing this can be a breaking change.

```ruby
group[...][:polling_strategy] = MyClass
``

phstc: Sorry for being extremely careful. I just want to avoid breaking changes, if any.

agrajgarg1526: @phstc I tried doing `group[...][:polling_strategy] = MyClass` in`polling_strategy` function using monkey patching and it got matched with `Class`. But I don't think users can access `groups` and mutate it without overriding functions. But still to avoid this, keeping both the cases would be the best approach here imo.

phstc: Thank you for double-checking this. I agree with keeping both.

phstc: I'm running the test suite 🙏

phstc: @agrajgarg1526 can you check the test errors?

https://github.com/ruby-shoryuken/shoryuken/actions/runs/7835447741/job/21380862789

agrajgarg1526: @phstc moved the argument error exception inside the `when String` case.

PR #765, opened 2024-03-06
Scope the shoryuken context to the current fiber.

ioquatix:
I'd like to start a discussion about how to more efficiently manage the context during logging.

This exposes a proper interface for getting the current context.

In addition, I'd like to understand, do you expect if someone creates child fibers or threads, that the context should be shared? e.g. map-reduce / fan-out style workloads. I'm assuming the answer should be yes, but I'd like to know your opinion. In that case, it's better to use `Fiber#storage`.

### DIFF
--- a/lib/shoryuken/logging.rb
+++ b/lib/shoryuken/logging.rb
@@ -1,6 +1,10 @@
 require 'time'
 require 'logger'
 
+class Fiber
+  attr_accessor :shoryuken_context
+end
+
 module Shoryuken
   module Logging
     class Pretty < Logger::Formatter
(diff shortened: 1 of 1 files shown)


ioquatix:
@phstc do you have time to review? Thanks!

phstc:
Hi @ioquatix 

Sorry, I'm a bit out of the loop here. Are there any breaking changes here?

> I'd like to start a discussion about how to more efficiently manage the context during logging.

Asking in case this is good to go as a patch version or if there is any concern with a specific version of Ruby or Rails.

ioquatix:
In general, I don't believe this is a breaking change.

ioquatix:
@phstc any update?

phstc:
@ioquatix, Unfortunately, I'm not currently active with Ruby or Shoryuken. I'm mainly merging maintenance pull requests.

It would be nice to have input from another current active user. 

Can you try the Slack channel? https://join.slack.com/t/shoryuken/shared_invite/zt-19xjq3iqc-KmoJ6eU6~qvZNqcLzIrjww

PR #770, opened 2024-06-04
Bugfix: Extract only the supported message attributes for AWS SQS when `normalize_dump_message`

rsosag:
Linked to issue https://github.com/ruby-shoryuken/shoryuken/issues/769


phstc:
@rsosag Thank you the commit 🎉 

I'm not sure if related, but main is failing after merging this change.

https://github.com/ruby-shoryuken/shoryuken/actions/runs/10044740616/job/27760858106

PR #781, opened 2025-01-21
Provide an xml parser engine for aws sdk core version 2

jpawlyn:
otherwise we see the error:

```spec
Failure/Error: self.url  = client.get_queue_url(queue_name: name).queue_url

      NoMethodError:
        undefined method `new' for false
```

when running tests


phstc:
@jpawlyn thank you! tests are in progress. 

> Rails 4.2 test run

Can it be fixed? I wonder if we remove it and there are indeed people using 4.2.

jpawlyn:
> Can it be fixed? I wonder if we remove it and there are indeed people using 4.2.

4.2 was first released in [2014](https://rubygems.org/gems/rails/versions?page=3) and stopping support for it perhaps makes sense for ease of maintenance @phstc ?

phstc:
@jpawlyn good call, that makes sense. Can you remove it then?

We will probably need to release a major version given the breaking change, WDYT?

jpawlyn:
> @jpawlyn good call, that makes sense. Can you remove it then?
> 
> We will probably need to release a major version given the breaking change, WDYT?

I've removed Rails 4.2 from the matrix. I guess the change is not actually a breaking one since all we are doing is removing Rails 4.2 from the test matrix and there is no minimum Rails version specified in the gemspec.

phstc:
> I guess the change is not actually a breaking one

@jpawlyn I thought it was a breaking change because the change was breaking something on Rails 4.2.

jpawlyn:
> @jpawlyn I thought it was a breaking change because the change was breaking something on Rails 4.2.

Can you clarify @phstc what the change is ? Is there perhaps a merged PR that may have caused this test failure ?

phstc:
@mensfeld I think it's a solid idea. Sidekiq for instance is Rails 7.0+

mensfeld:
closing this one as after my todays changes there's nothing here to merge that would make sense.

PR #782, opened 2025-01-25
Fix integration tests by updating aws-sdk-sqs and replacing moto with LocalStack

myumura:
This PR replaces moto with LocalStack for integration testing. Changes include:

1. Updates aws-sdk-sqs to latest version
2. Adds docker-compose configuration for LocalStack
3. Updates integration tests to use LocalStack

Background:
Currently, aws-sdk-sqs is pinned to v1.65.0 due to moto compatibility issues. The mock service returns XML responses while aws-sdk-sqs expects JSON, causing parse errors. 

After experimenting with LocalStack, it proved to handle the AWS SDK responses correctly, allowing us to:
- Use the latest aws-sdk-sqs version
- Remove version constraints
- Have a more AWS-like testing environment

What do you think about this approach?

Related to #781

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -13,10 +13,17 @@ jobs:
         gemfile: ['Gemfile', 'gemfiles/aws_sdk_core_2.gemfile']
     runs-on: ubuntu-20.04
     services:
-      moto_sqs:
-        image: quay.io/cjlarose/moto-sqs-server:1.1.0
-        ports:
-          - 5000:5000
+      localstack:
+      image: localstack/localstack:latest
+      environment:
+        - SERVICES=sqs
+      ports:
+        - 4566:4566
+      healthcheck:
+        test: ["CMD", "curl", "-f", "http://localhost:4566/_localstack/health"]
+        interval: 5s
+        timeout: 10s
+        retries: 5
     env:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
     steps:
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -15,15 +15,15 @@ jobs:
     services:
       localstack:
         image: localstack/localstack:latest
-        environment:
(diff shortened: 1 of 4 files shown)


myumura:
📝
The cleanup code was causing format mismatch errors with moto (XML vs JSON responses). 
While we could switch to localstack which properly handles JSON responses, cleanup isn't necessary since both moto and localstack automatically clean up resources on process exit.

mensfeld:
@phstc I was able to get this running:

```bash
(base) [shoryuken (fix-integration-tests)]$ bundle exec rake spec:integration
/home/mencio/.rbenv/versions/3.2.2/bin/ruby -I/home/mencio/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/rspec-core-3.13.2/lib:/home/mencio/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/rspec-support-3.13.2/lib /home/mencio/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/rspec-core-3.13.2/exe/rspec --pattern spec/integration/\*\*/\*_spec.rb
...
Finished in 3.64 seconds (files took 0.45868 seconds to load)
3 examples, 0 failures

```

it does require newer aws-sdk-sqs as mentioned above but the connection with localstack works locally when using this:

```bash
docker run \
  -e SERVICES=sqs \
  -p 4566:4566 \
  --health-cmd "curl -f http://localhost:4566/_localstack/health" \
  --health-interval=5s \
  --health-timeout=10s \
  --health-retries=5 \
  localstack/localstack:latest
```

the issue though is with the gemspec requirement on newer `aws-sdk-sqs`. @myumura do you know by any chance the minimal one that supports new API format so we could pin it in the gemspec?

mensfeld:
Ok, I will retarget this to a custom branch where I can continue reworking this until ready to be merged to master. That way @myumura contribution will be recorded :pray:

PR #803, opened 2025-02-07
Cleanup redundant gems in the Gemfiles

mensfeld:
1. I'm opening rubocop to keep up with its changes
2. Removing redundant core references and other duplicate references
3. Removing codeclimate gem as it has been stale for 4 years now. We connect codecov reporting directly


mensfeld:
@phstc please remove the core_2 spec requirement because we already dropped it previously. This is just a cleanup

phstc (approved):

PR #869, opened 2025-08-13
Replace Concurrent::AtomicBoolean with pure Ruby implementation

myumura:
Replaces `Concurrent::AtomicBoolean` with a pure Ruby implementation `Shoryuken::Helpers::AtomicBoolean` to reduce external dependencies on the concurrent-ruby gem.


mensfeld (commented):
A small remark, aside from that great work!

mensfeld (approved):

PR #873, opened 2025-08-17
Replace Concurrent::TimerTask with pure Ruby implementation

myumura:
Replaces Concurrent::TimerTask with a pure Ruby implementation Shoryuken::Helpers::TimerTask to reduce external dependencies on the concurrent-ruby gem.


mensfeld:
@myumura i think we can do it that way for now but I also think that in the future we could use a model where we have a single thread that performs all the time and one-shot operations based on a queue + a periodic scheduler. Great job!

mensfeld (approved):

